Issue #4, opened 2018-01-27
Show All Clusters Across All Projects (can be filtered by e.g. having issues or not)


vlerenc:
# Story
- As ~~operator~~ admin (the same means we use to give the operator of the week access to all projects), I want to see all clusters across all projects and filter them by e.g. having issues or not, having a certain label set or not (see #33), so that I can do my ops tasks more easily.

# Acceptance Criteria
- [ ] Dashboard offers means to show all clusters across all projects ~~the logged in user is member of~~ if an admin is logged in (e.g. from the project selector)
- [ ] Dashboard allows to filter the above list, e.g. by having issues or not or by having a certain label set or not (see #33)

# Definition of Done
- [ ] Knowledge is distributed: Have you spread your knowledge in pair programming/code review?
- [ ] Unit Tests are provided: Have you written automated unit tests or added manual NGPTT tickets?
- [ ] Integration Tests are provided: Have you written automated integration tests?
- [ ] Minimum API exposure: If you have added public API, was it really necessary/is it minimal?
- [ ] Operations guide: Have you updated the [operations guide](https://github.wdf.sap.corp/kubernetes/kube-docs/wiki/Operations%20Guide)?

PR #40, opened 2018-03-19
[Feature] #4 Show All Clusters & Add ability to filter by Shoots with Issues

grolu:


backend/lib/services/shoots.js
```diff
@@ -30,7 +30,11 @@ function Core ({auth}) {
 }
 
 exports.list = async function ({user, namespace}) {
-  return Garden(user).namespaces(namespace).shoots.get({})
+  if (!namespace) {
```
petersutter: do we still make use of this? (calling this function without a namespace)

PR #124, opened 2018-07-04
Use github octokit client library

holgerkoser:
!!!Do not merge this PR until we have discussed the changes!!!
### Intended improvements
* use github octokit client library
* added acceptance tests for webhook and journal cache
* tried to follow nodejs best practices and conventions

### Changes in detail
* [lib/github/octokit.js](/gardener/dashboard/blob/octokit/backend/lib/github/octokit.js)
Wrapper around the octokit client library. Adds `PageStream` class which creates a `Readable` stream based on pageable github resources. It adds `reduce` and `thru` methods to the streams. They allow interception / transform the chunk or to concat all data. Both return a `Promise` if the stream ended or failed. It also read authentiction from config.
* [lib/github/index.js](/gardener/dashboard/blob/octokit/backend/lib/github/index.js)
This module exposes those functions needed for issues and comments in our case. They make use of the `PageStream` for pageable resources.
* [lib/github/webhook.js](/gardener/dashboard/blob/octokit/backend/lib/github/webhook.js)
Implements the github webhook route and does JSON body parsing with signature validation using the body-parser verify hook. It handles the github webhook events and updates the cache for issues and comments. In case of an reopended issues it reloads all comments and updates the cache in the background. All webhooks return synchronous. 
* [lib/services/journals.js](/gardener/dashboard/blob/octokit/backend/lib/services/journals.js)
This module exposes higher level functions for handling issues and comments. All function starting with `load` return an empty Promise and do a cache update. The functions starting with `get` return a stream with a `reduce` method. This allows to e.g write to the cache.
* [lib/cache/journals.js](/gardener/dashboard/blob/octokit/backend/lib/cache/journals.js)
This module now returns a function to create the cache instead of the cache object. This is needed for testing. The event registration is now done by the `EventEmitter` class.
* [lib/io.js](/gardener/dashboard/blob/octokit/backend/lib/io.js)
Extracted some functions to make it more modular. In case of journalNsp all comments for each issue are fetch sequentially in order to handle errors individually.


backend/lib/github/index.js
```diff
@@ -16,125 +16,59 @@
...
-const logger = require('../logger')
+const octokit = require('./octokit')()
+
 const config = require('../config')
-const parseLink = require('parse-link-header')
-const { URL } = require('url')
-const urljoin = require('url-join')
+const {
+  org: owner = 'kubernetes-dev',
+  repository: repo = 'journal-dev'
```
petersutter: do we have to default the org and repo? As I see we also have default values in config/gardener.js

petersutter:
we should also add a *ca* field under *github* property in [values.yaml](https://github.com/gardener/dashboard/blob/master/charts/gardener-dashboard/values.yaml#L53) file
and use it accordingly in [configmap.yaml](https://github.com/gardener/dashboard/blob/master/charts/gardener-dashboard/templates/configmap.yaml#L63)

holgerkoser:
Replaced by PR #146

PR #136, opened 2018-07-19
use american english for pipeline definition

RaphaelVogel:
Users expect to write american english in pipeline definitions.
Therefore rename "finalise" to "finalize"


petersutter:
@RaphaelVogel: I assume this change is already compatible with the pipeline?

Issue #105, opened 2018-06-12
Show indicator when websocket connection is lost


petersutter:
The UI is showing potentially old data when the websocket connection is lost. Currently, the user does not notice this as we do not show an indicator / error on the UI.

PR #140, opened 2018-07-25
Snotify Notification Framework

### LINKED ISSUE #129: Introduce Notification Framework
As Vuetify does not provide notifications, we should evaluate notification frameworks available for Vue.js. We should then use the new framework for all global notifications / errors that we currently show as toasts / using snackbars. In addition, we can show persistent notifications for lost websocket connection and maybe get rid of some v-alerts (show them as persistent notifications instead)

grolu:
This PR introduces snotify notification framework which replaces vue toasts that we previously used to show errors to the user.

Fixes #129 #105 

```improvement user
Introduced new notification framework
```

```noteworthy user
A warning notification is shown when the server connection is lost. Related issue: gardener/dashboard#105
```

```improvement user
A notification is shown in case a cluster is about to self terminate
```

```improvement user
Improved project member add user / service account dialog usability
```

```improvement user
No default project name is proposed when creating a new project
```

```improvement user
Improved overall action feedback: Show errors inside dialogs and give feedback about successful operations via notifications
```


backend/lib/services/members.js
```diff
@@ -160,7 +160,9 @@ async function setRoleBindingSubject (rbac, namespace, name) {
   }
   const subjects = body.subjects = body.subjects || []
   if (_.find(subjects, ['name', name])) {
-    return body
+    const err = new Error(`User '${name}' is already member of this project`)
```
petersutter: use Conflict error from error.js

PR #146, opened 2018-07-27
Use octokit github client

### LINKED ISSUE #145: Use github octokit client library
Use the official [@octokit/rest](https://octokit.github.io/rest.js/) GitHub REST API client for Node.js instead of our own http client based implementation

holgerkoser:
#### Intended improvements
- use github octokit client library
- added acceptance tests for webhook and journal cache
- tried to follow nodejs best practices and conventions
- allow configuration of github ca in helm chart

#### Changes in detail
- backend/lib/github/octokit.js
Wrapper around the octokit client library. Adds PageStream class which creates a Readable stream based on pageable github resources. It adds reduce and thru methods to the streams. They allow interception / transform the chunk or to concat all data. Both return a Promise if the stream ended or failed. It also read authentiction from config.
- backend/lib/github/index.js
This module exposes those functions needed for issues and comments in our case. They make use of the PageStream for pageable resources.
- backend/lib/github/webhook.js
Implements the github webhook route and does JSON body parsing with signature validation using the body-parser verify hook. It handles the github webhook events and updates the cache for issues and comments. In case of an reopended issues it reloads all comments and updates the cache in the background. All webhooks return synchronous.
- backend/lib/services/journals.js
This module exposes higher level functions for handling issues and comments. All function starting with load return an empty Promise and do a cache update. The functions starting with get return a stream with a reduce method. This allows to e.g write to the cache.
- backend/lib/cache/journals.js
This module now returns a function to create the cache instead of the cache object. This is needed for testing. The event registration is now done by the EventEmitter class.
- backend/lib/io.js
Extracted some functions to make it more modular. In case of journalNsp all comments for each issue are fetch sequentially in order to handle errors individually.
- charts/gardener-dashboard/templates/configmap.yaml
Render ca file content in JSON format for oidc and github ca.

Fixes #145 

**Release note**:
```improvement operator
[Journal] New optional configuration value in helm chart of gardener-dashboard for github certificate authority
```

*configmap.yaml* example:
```yaml
...
gitHub:
  apiUrl: https://api.foo-github.com
  ca: |-
    -----BEGIN CERTIFICATE-----
    Li4u
    -----END CERTIFICATE-----
```


backend/lib/config/gardener.js
```diff
@@ -50,6 +50,16 @@ module.exports = {
...
       logLevel: isProd ? 'warn' : 'debug',
       port,
+      apiServerUrl: 'https://minikube:8443',
+      gitHub: {
+        apiUrl: 'https://api.github.com',
+        org: 'gardener',
+        repository: 'journal-dev',
+        webhookSecret: '776562686f6f6b536563726574',
+        authentication: {
+          token: '746f6b656e'
```
petersutter: 776562686f6f6b536563726574 === hexEncode('webhookSecret')
746f6b656e === hexEncode('token')
-> these are just dummy values used for unit tests, correct?

backend/lib/config/testing.yaml
```diff
@@ -1,6 +1,7 @@
 port: 3030
 logLevel: info
 logFormat: text
+apiServerUrl: https://minikube:8443
```
petersutter: is this still used? as we have the same value in *backend/lib/config/gardener.js*

PR #153, opened 2018-08-07
Healthz endpoint

### LINKED ISSUE #108: add /healthz endpoint
The dashboard should provide a simple health endpoint that returns 200 if the dashboard is up and running

grolu:
Add a simple healthz endpoint to the dashboard which returns 200/ok in case the backend is up and has connection to the Kubernetes API server

Fixes #108

**Release note**:
```noteworthy operator
Added a simple healthz endpoint to the dashboard which returns 200 / ok in case the backend is up and has connection to the Kubernetes API server
```


petersutter:
FYI @mandelsoft

holgerkoser:
* Does this endpoint need an authenticated user?
* A small acceptance test would be great. This would clarify if an authenticated user is intended or not.
* Why do we create the client instance on each request? What do you think could change between two requests?

grolu:
- It does require an authenticated user (does not require admin user - which is actually a good point, I corrected this, see last commit), I will write a test for it.
- @mandelsoft Do you require an unauthenticated endpoint? If so, we cannot ping the kubernetes api server as we would return protected information with an unprotected endpoint. We could only give a 200 ok for the dashboard instance itself
- We can cache the instance, in case we do not expect the kubeconfig to change (if we want to support this, we need a central client cache where we can invalidate the clients if the config changes)
- We should also do this in other places where we do not cache the instance, I can create a ticket for this.

holgerkoser:
In a default kubernetes cluster the cluserrole `system:discovery` is allowed to read the healthz endpoint. This clusterrole is bound to group `system:unauthenticated`. I don't understand why we need an authenticated user. Maybe I am missing something. 
```yaml
---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  annotations:
    rbac.authorization.kubernetes.io/autoupdate: "true"
  labels:
    kubernetes.io/bootstrapping: rbac-defaults
  name: system:discovery
rules:
- nonResourceURLs:
  - /api
  - /api/*
  - /apis
  - /apis/*
  - /healthz
  - /openapi
  - /openapi/*
  - /swagger-2.0.0.pb-v1
  - /swagger.json
  - /swaggerapi
  - /swaggerapi/*
  - /version
  - /version/
  verbs:
  - get
---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  annotations:
    rbac.authorization.kubernetes.io/autoupdate: "true"
  labels:
    kubernetes.io/bootstrapping: rbac-defaults
  name: system:discovery
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: system:discovery
subjects:
- apiGroup: rbac.authorization.k8s.io
  kind: Group
  name: system:authenticated
- apiGroup: rbac.authorization.k8s.io
  kind: Group
  name: system:unauthenticated
```

rfranzke:
We start the kube-apiservers with `--anonymous-auth=false`, hence, any unauthenticated request is blocked.

holgerkoser:
@rfranzke thanks for clarification.

holgerkoser:
I understood we need an authenticated user because the apiserver also requires an authenticated user. But unfortunately we only allow users authenticated by the oidc issuer (see https://github.com/gardener/dashboard/blob/master/backend/lib/api.js#L39). No service accounts, no basic auth, no client certs. I don't think that @mandelsoft can pass a jwt token of an oidc user.

grolu:
Correct @msohn

PR #154, opened 2018-08-07
use node:8.11-alpine base image

petersutter:
node 8.11.3 fixes CVE-2018-7167 https://github.com/nodejs/node/releases/tag/v8.11.3

Fixes 

**Release note**:

```noteworthy operator
Bumped node version to 8.11.3 that fixes CVE-2018-7167
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,7 +14,7 @@
 # limitations under the License.
 #
 
-FROM node:8.9-alpine
+FROM node:8.11-alpine
 
 # Create app directory
 RUN mkdir -p /usr/src/app/public


holgerkoser:
Why not directly to node:10.8-alpine?

petersutter:
I figured 8.11 would be the least intrusive change to include a fix for this vulnerability, that was the only reason..

holgerkoser:
Of course. The question is: "Do we proactively upgrade from time to time whenever there is a new version with some enhancements. Or do we only upgrade when is is a vulnerability in the old old version." I think in 10.8 are some good features we could use e.g "util.promisify"

petersutter:
yes, I also think it makes sense to proactively upgrade from time to time if there are some nice features that we can make use of
I got your point, however, just as a side remark: is util.promisify not already part of 8.x?

holgerkoser:
Yes `util.promisify` is part of 8. I have mixed it up with `stream.pipeline` which comes with 10.

petersutter:
protecode still flags the 8.11.3 to contain the CVE-2018-7167 vulnerability (even though in the release notes of node.js it says that it is fixed), hence we agreed to use 10.8 which does not have this vulnerability

PR #157, opened 2018-08-08
Tag dashboard image as 'latest'

### LINKED ISSUE #141: Tag dashboard image as 'latest'
tag dashboard image as latest, so that we can reference it directly [here](https://github.com/gardener/dashboard/blob/master/charts/gardener-dashboard/values.yaml#L8) so that we do not have to update this file after every build
 

> You may now tag images published by our Concourse as 'latest' *in addition* to the effective version by setting the `tag_as_latest` attribute to `true`. Example:

```yaml
...
  traits:
    publish:
      dockerimages:
        some-image-name:
          registry: gcr-readwrite
          image: eu.gcr.io/foo/bar
          dockerfile: Dockerfile
          tag_as_latest: true
...
```

petersutter:
tag dashboard image as latest, so that we can reference it directly [here](https://github.com/gardener/dashboard/blob/master/charts/gardener-dashboard/values.yaml#L8) so that we do not have to update this file after every build

Fixes #141

**Release note**:

```improvement operator
NONE
```

### DIFF
--- a/.ci/pipeline_definitions
+++ b/.ci/pipeline_definitions
@@ -35,6 +35,10 @@ dashboard:
         release:
           nextversion: 'bump_minor'
         component_descriptor: ~
+        publish:
+          dockerimages:
+            dashboard:
+              tag_as_latest: true
       steps:
         prepare_release:
           image: *node_image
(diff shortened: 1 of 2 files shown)


petersutter:
currently it would be the latest build (coming from head update / release), can you confirm @AndreasBurger ?

AndreasBurger:
Yep, I rendered the pipeline locally _just_ to be sure and it does what is expected: Only images published in the release job are also tagged with latest.

PR #176, opened 2018-08-15
Fix too many project issue

grolu:
We face issues when a user (operator) has access to many projects and tries to fetch the all shoots list. This PR adds a special handling for users that have the operator role. These users can fetch all shoots at once (without namespace selector). This way, we do not need to make a requests for each namespace to the Kubernetes API.

Fixes #

```noteworthy operator
Fixes an issue that leads to errors when a user (most likely with the operator role) has access to many projects and tries to fetch the all clusters list
```

### DIFF
--- a/backend/lib/io.js
+++ b/backend/lib/io.js
@@ -111,7 +111,7 @@ function leaveCommentsRooms (socket) {
 }
 
 function setupShootsNamespace (shootsNsp) {
-  const subscribeShoots = async function ({socket, namespaces, validateNamespaces}) {
+  const subscribeShoots = async function ({socket, namespaces, projectList}) {
     leaveShootsAndShootRoom(socket)
 
     /* join current rooms */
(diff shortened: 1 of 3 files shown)


backend/lib/io.js
```diff
@@ -126,30 +126,62 @@ function setupShootsNamespace (shootsNsp) {
...
       const user = getUserFromSocket(socket)
       const batchEmitter = new NamespacedBatchEmitter({kind, socket, objectKeyPath: 'metadata.uid'})
-
       const projectList = await projects.list({user})
-      await _
-        .chain(namespaces)
-        .filter(({namespace}) => !!_.find(projectList, ['metadata.namespace', namespace]))
-        .map(async ({namespace, filter}) => {
+
+      if (_.get(_.head(namespaces), 'namespace') === '_all') {
```
petersutter: extract to variable

backend/lib/io.js
```diff
@@ -126,30 +126,62 @@ function setupShootsNamespace (shootsNsp) {
...
+        try {
+          const filter = _.get(_.head(namespaces), 'filter')
```
petersutter: use variable

grolu:
@holgerkoser I implemented and tested the changes proposed by you. In addition, I removed the filter statement(s) as I believe this is a leftover (before the apiserver set the unhealthy label).

PR #180, opened 2018-08-31
Small fixes

### LINKED ISSUE #179: Randomise default AZ selection in creation wizard
It would be a good idea to randomise AZ selection in the creation wizard. That way we could prevent that especially first time users will end up creating a lot of clusters in one AZ (which ever is selected by default). This is especially important for our OpenStack support where the capacity per AZ is limited.

grolu:

Fixes #179 

**Release note**:
```improvement operator
Randomize Availability Zone in Cluster Creation Dialog to Help Spreading Usage Evenly Across Different Zones
```
```improvement user
Secrets Page: UI Failed Rendering Shared Secret Bindings for OpenStack
```

### DIFF
--- a/frontend/src/components/Secret.vue
+++ b/frontend/src/components/Secret.vue
@@ -46,7 +46,7 @@ limitations under the License.
         @delete="onDelete"
       >
         <!-- forward slot -->
-        <slot slot="rowSubTitle" name="rowSubTitle" :data="row.data"></slot>
+        <slot slot="rowSubTitle" name="rowSubTitle" :secret="row"></slot>
       </secret-row>
     </v-list>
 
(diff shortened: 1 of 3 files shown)


frontend/src/components/Secret.vue
```diff
@@ -46,7 +46,7 @@ limitations under the License.
         @delete="onDelete"
       >
         <!-- forward slot -->
-        <slot slot="rowSubTitle" name="rowSubTitle" :data="row.data"></slot>
+        <slot slot="rowSubTitle" name="rowSubTitle" :secret="row"></slot>
```
petersutter: slot="rowSubTitle"

PR #184, opened 2018-09-07
support .status.lastOperation.state=Aborted

### LINKED ISSUE #181: New Shoot status should be reflected
The `.status.lastOperation.state` can have one of the `Processing|Succeeded|Error|Failed` values. Since the last version Gardener sets `.status.lastOperation.state=Aborted` for all Shoots that were `Processing` during its start.
The dashboard does not know about this new state and shows the green checkbox as if the state would be `Succeeded`.

petersutter:
see #181

Fixes #181

**Release note**:

```noteworthy user
If the Gardener instance working on a Shoot has been interrupted the last operation status will be set to `Aborted`, which is now properly displayed with an own icon.
```


petersutter:
instead of showing an own "aborted/cancel" icon, I think would be better to show the "previous" icon (dependant on the type) but greyed out (including a tooltip), hence for now I will mark it as do not merge until we have discussed it / made the changes

petersutter:
screen shot 2018-09-12 at 12 46 18
screen shot 2018-09-12 at 12 46 39

grolu:
I like the idea to gray-out icons, however the create icon without progress circle looks a bit like an add button to me.

petersutter:
My first try was to keep the progress circle but I quickly discarded it as there is no progress in this state. Also the icon is within the status column so I think it's not that "critical". 

I suggest we keep it that way for now and check if we get user feedback

grolu:
Yes, if there are multiple clusters it will be even more self-explaining. I agree, let's keep it that way and wait for user feedback. Besides, I only wanted to share my first impression, I'm not saying that I've got a better idea.

PR #186, opened 2018-09-10
Remove Heapster

### LINKED ISSUE #185: Remove Heapster from Shoot spec
Gardener has removed support for deploying Heapster and replaced it with the standard deployment of the Kubernetes metrics-server.
The `.spec.addons.heapster` field is still available in the Shoot spec, but deprecated and without any effect anymore. It will be removed in a future version.
The dashboard should not set this field anymore or show any reference to Heapster itself.

petersutter:
see #185

Fixes #185

**Release note**:

```noteworthy user
Removed support for deploying Heapster as addon. It is replaced by the standard deployment of the Kubernetes metrics-server.
```


frontend/src/pages/ShootItem.vue
```diff
@@ -326,11 +326,6 @@ limitations under the License.
             title: 'Cluster Autoscaler',
             description: 'Cluster Autoscaler is a tool that automatically adjusts the size of the Kubernetes cluster.'
           },
-          {
-            name: 'heapster',
-            title: 'Heapster',
-            description: 'Heapster enables Container Cluster Monitoring and Performance Analysis.'
-          },
```
holgerkoser: I would keep this because old cluster can still have the heapster addon. Only new clusters should not use it anymore.

petersutter: good point

petersutter:
Done. I have reverted the changes in ShootItem.vue

PR #205, opened 2018-10-10
patch package.json in release commit

ccwienk:

Configures dashboard's release job in such a way that it also contains updates to package.json files. This differs from the previous approach of creating and publishing a separate commit with those changes upfront the release commit.

As a side effect, this change should repair the currently broken release build job, which failed because of the "missing" rebase prior to publishing the release commit (which was not based on the afforementioned upgrade commit).

### DIFF
--- a/.ci/prepare_release
+++ b/.ci/prepare_release
@@ -2,40 +2,38 @@
 
 set -e
 
-if [ -z "$SOURCE_PATH" ]; then
-  export SOURCE_PATH="$(readlink -f $(dirname ${0})/..)"
+if [ -z "$REPO_DIR" ]; then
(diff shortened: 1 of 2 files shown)


.ci/prepare_release
```diff
@@ -2,40 +2,38 @@
...
+      exit 1
+    fi
+else
+    version="${EFFECTIVE_VERSION}"
 fi
 
-# revert changes VERSION
-git checkout "${SOURCE_PATH}/VERSION"
+# revert changed VERSION
+git checkout "${REPO_DIR}/VERSION"
```
ccwienk: btw: the patching of VERSION file within work tree can be disabled via `pipeline_definition`; then there would not be the need to revert it

PR #224, opened 2018-10-18
Refactor kubeconfig ui element

### LINKED ISSUE #148: Refactor Shoot Access Card
We should show the KubeConfig as part of the Cluster Access Popup on the Cluster List (same as on shoot details). This way we can get rid of the KubeConfig Button on the Cluster List. Moreover, we should offer a possibility to download the KubeConfig (as we provide for Service Accounts).

petersutter:
With this PR it is possible to download the kubeconfig for a shoot. Also the ui element has been changed a bit so that you don't need to view the kubeconfig in order to copy it.
With these changes we could also get rid of the show kubeconfig action on the cluster list as this is now part of the cluster access dialog (see screenshot below)

**Cluster Details**
screenshot 2018-10-18 at 16 29 53

**Cluster List**
screenshot 2018-10-18 at 16 34 37

Fixes #148 

**Release note**:

```improvement user
Kubeconfig improvements:
- You can download the kubeconfig for your cluster
- You can copy the kubeconfig for your cluster without viewing it
```

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -1803,9 +1803,9 @@
       }
     },
     "@mdi/font": {
-      "version": "2.8.94",
-      "resolved": "https://registry.npmjs.org/@mdi/font/-/font-2.8.94.tgz",
-      "integrity": "sha512-QSmuK0BedRMDSgh7RCOc0yO+cAHAaAKfj11BPv0XDxvGKNNKGRqDJcJKdrWUpVmO1/xiFLWfPEmqMi/PKwvjcg=="
+      "version": "3.0.39",
+      "resolved": "https://registry.npmjs.org/@mdi/font/-/font-3.0.39.tgz",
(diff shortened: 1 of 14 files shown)


grolu:
What about the service account section on the members page? We also have a show kubeconfig button there and it is currently impossible to copy the kubeconfig without viewing it...

petersutter:
If we change it also there we would have 4 buttons:
Download, Copy, View, Delete

Viewing the kubeconfig could either open a popup (as it is currently) or we could expand the section. What do you think?

petersutter:
I just checked the coding and we can't copy the kubeconfig of a robot to the clipboard as we download it asynchronously and thus can't put it in the clipboard. If we want to support this we have to take a closer look at the new `Clipboard API` which is not yet supported by safari and the latest (stable) Firefox which is 62
https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API#Browser_compatibility

in the future we could do something like this
```
    async copyKubeconfigToClipboard (name) {
      const kubeconfig = await this.downloadKubeconfig(name)
      navigator.clipboard.writeText(kubeconfig)
        .then(() => {
          // show success message
        })
        //.catch and show error message
    }
```

for now, I have adapted the show kubeconfig icon so that it looks the same as on the cluster details e.g. but still the popup is shown

screenshot 2018-10-19 at 17 20 44

PR #231, opened 2018-10-22
Adapt to new project resource behavior

holgerkoser:
The new project resource has the `gardener` as initializer. The corresponding namespace and the RBAC role bindings are created later on. The dashboard has to wait until the initializer is ready before the project can be used. Only the project owner is allowed to modify the project. Project members are only allowed to read the project. Every authenticated user can create project resources.

Fixes #

**Release note**:
```improvement user
Project members are not allowed to modify the project anymore. This can only be done by the owner. Project creation takes some time (a few seconds).
```

### DIFF
--- a/backend/lib/services/projects.js
+++ b/backend/lib/services/projects.js
@@ -21,7 +21,7 @@ const kubernetes = require('../kubernetes')
 const Resources = kubernetes.Resources
 const garden = kubernetes.garden()
 const core = kubernetes.core()
-const { PreconditionFailed, NotFound } = require('../errors')
+const { PreconditionFailed, NotFound, GatewayTimeout, InternalServerError } = require('../errors')
 const logger = require('../logger')
 const shoots = require('./shoots')
 const administrators = require('./administrators')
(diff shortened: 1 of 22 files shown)


holgerkoser:
We may should add a spinner with text or a snackbar to inform the user that he should be patient after project creation.

petersutter:
depends on https://github.com/gardener/gardener/pull/480

backend/lib/services/projects.js
```diff
@@ -17,66 +17,72 @@
...
+  return {
+    metadata: {
+      name,
+      namespace,
+      resourceVersion,
+      role,
+      creationTimestamp
+    },
+    data: {
+      createdBy,
```
petersutter: @rfranzke said we should not set createdBy. This is done by gardener
-> remove

backend/lib/services/projects.js
```diff
@@ -17,66 +17,72 @@
...
+  return {
+    apiVersion,
+    kind,
+    metadata: {
+      name,
+      resourceVersion
+    },
+    spec: {
+      namespace,
+      createdBy: createdByRef,
```
petersutter: @rfranzke said we should not set createdBy. This is done by gardener
-> remove

PR #246, opened 2018-11-09
Fix Browser search on shoot editor

petersutter:
With this PR we render the complete shoot resource so that the browser's text search works on it

Fixes #

As alternative we could use codemirror's search addon https://codemirror.net/demo/search.html but for now this should be sufficient

**Release note**:

```improvement user
Fixed issue that the Browser's text search on the cluster editor did not work reliably in some cases
```


petersutter:
cc @ThormaehlenFred

PR #247, opened 2018-11-09
Disable create button if user has not enough priviliges

### LINKED ISSUE #234: Disable Create Project Button
If a user has not enough privileges to create a project, we should disable the button. It should be possible to define a custom error message in the dashboard config map for this case.
Idea: Provide privileges via /user (userinfo) endpoint.
We should also redesign the "Empty Project List" Page. This is the first page a user sees when he starts working with the dashboard. We should make this page look a bit more appealing.

holgerkoser:
If a user has not enough privileges to create a project, the create project button is disabled.

Fixes #234 

**Release note**:

```improvement user
The create project button is only enabled, if a user has not enough privileges to create a project (this fixes issue #234).
```

### DIFF
--- a/backend/lib/services/shoots.js
+++ b/backend/lib/services/shoots.js
@@ -19,7 +19,7 @@
 const kubernetes = require('../kubernetes')
 const { decodeBase64 } = require('../utils')
 const { getSeeds } = require('../cache')
-const administrators = require('./administrators')
+const authorization = require('./authorization')
 const _ = require('lodash')
 
 function Garden ({auth}) {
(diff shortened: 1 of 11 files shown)


backend/lib/services/shoots.js
```diff
@@ -62,7 +62,7 @@ exports.list = async function ({user, namespace, shootsWithIssuesOnly = false})
   if (namespace) {
     return Garden(user).namespaces(namespace).shoots.get({qs})
   } else {
-    // only works if user is member of garden-administrators (admin)
+    // only works if user is member of garden-authorization (admin)
```
petersutter: revert

PR #252, opened 2018-11-21
set env var NODE_END for release job

ccwienk:

Passes a flag to npm to indicate the build is "productive" during release builds


holgerkoser:
Sorry for the inconvenience but the NODE_ENV already is production in the build step by default (see https://cli.vuejs.org/guide/cli-service.html#vue-cli-service-build)

PR #286, opened 2019-01-10
Add kibana ingress to cluster details

ialidzhikov:
Add kibana ingress to cluster details.

Fixes #

**Release note**:

```improvement operator
Kibana ingress has been added to `Cluster Details` page.
```

### DIFF
--- a/backend/lib/services/shoots.js
+++ b/backend/lib/services/shoots.js
@@ -31,6 +31,35 @@ function Core ({ auth }) {
   return kubernetes.core({ auth })
 }
 
+function getSecret (core, namespace, name) {
+  return core.ns(namespace).secrets.get({ name: name })
+    .catch(err => {
+      if (err.code === 404) {
+        return
+      }
+      logger.error('failed to fetch %s secret: %s', name, err)
+      throw err
+    })
+}
+
+async function assignComponentSecret (core, namespace, component, data) {
+  const secret = await getSecret(core, namespace, `${component}-ingress-credentials`)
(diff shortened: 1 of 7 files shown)


backend/lib/services/shoots.js
```diff
@@ -227,27 +227,31 @@ exports.info = async function ({ user, namespace, name }) {
...
-            .pick('username', 'password')
-            .forEach((value, key) => {
-              if (key === 'password') {
-                data['monitoring_password'] = decodeBase64(value)
-              } else if (key === 'username') {
-                data['monitoring_username'] = decodeBase64(value)
+        const components = ['monitoring', 'logging']
+        for (let i = 0; i < components.length; i++) {
+          const component = components[i]
+          const componentSecret = await kubernetes.core(kubernetes.fromKubeconfig(seedKubeconfig)).ns(seedShootNS).secrets.get({ name: `${component}-ingress-credentials` })
```
petersutter: extract to variable and reuse `kubernetes.fromKubeconfig(seedKubeconfig)`
also or alternatively, reuse the client `kubernetes.core(kubernetes.fromKubeconfig(seedKubeconfig)).ns(seedShootNS).secrets`

petersutter: in addition, instead of fetching the secrets sequentially, use Promise.all
you could also consider to move parts of this method to an own function for better re-usability and (just as an example) pass the kubernetes client and components for which secrets should be fetched and the data object

PR #288, opened 2019-01-16
Moved ingress annotations to values file to allow custom annotations

eaterm:
Changed the ingress annotations to allow for custom annotations.
The is useful if specific annotations need to be set, for instance for lets encrypt. 

Fixes #

**Release note**:

```improvement operator
Added the ability to add custom ingress annotations in helm chart
```

### DIFF
--- a/charts/gardener-dashboard/values.yaml
+++ b/charts/gardener-dashboard/values.yaml
@@ -22,7 +22,10 @@ resources:
 hosts:
   - dashboard.ingress.example.org
   - dashboard.example.org
-
+annotations:
+  nginx.ingress.kubernetes.io/ssl-redirect: "true"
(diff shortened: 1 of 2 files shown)


charts/gardener-dashboard/values.yaml
```diff
@@ -22,7 +22,10 @@ resources:
 hosts:
   - dashboard.ingress.example.org
   - dashboard.example.org
-
+annotations:
```
petersutter: please put `annotations` under an `ingress` property, in case there is a need to customize the annotations of a different resource

eaterm: The `annotations` have been moved to the `ingress` property

PR #303, opened 2019-02-12
Fix IP Range Issue for Alicloud

### LINKED ISSUE #298: Pod CIDR in shoot for Alicloud is not valid
In Alicloud, 100.100.X.X is used by infrastructure like metadata endpoint, DNS server. This ip range should be avoid when creating a shoot cluster

jia-jerry:
Fixes #298


jia-jerry:
@petersutter , let's talk in this PR thread.

> 1. We would like to keep the shoot specs as homogenous as possible across different landscapes and infrastructures. Therefore, please only change the CIDR range for Alicloud if this is required, e.g. because Alibaba uses some of the IPs for their services. If you got a conflict with seed IPs please change the seed config. Could you please elaborate / point me to some documentation about possible address conflicts on Alicloud?
>2. The new ranges you proposed are not in the private range (see https://en.wikipedia.org/wiki/Reserved_IP_addresses). This will lead to conflicts. Why did you chose those ranges?

1. Only change the shoot of Alicloud. This PR is not for conflict with Seed IPs. Alicloud's meta-data ip is 100.100.100.200 (DNS is 100.100.x.x). So we need change in shoot definition.  Document: https://www.alibabacloud.com/help/doc-detail/49122.htm
2. Thanks for you suggestion. I correct it in this PR>

grolu:
Hi Jerry, the IP ranges look good now. However, we discussed this with @rfranzke and he told us that the dashboard should not set defaults for the pods and services network. This is done by the Gardener, so you need to change defaults for Alicloud in the Gardener itself. We will remove the defaults from the dashboard and make a release soon, so that we do not overwrite defaults set by the Gardener anymore.

jia-jerry:
@petersutter , would you please merge it? We are a little in a hurry to integrate Alicloud into our Dev/Staging/Canary landscape. It is better to show to our stakeholders that we enable Alicloud this week. We are looking forward to the new release of the Dashboard. As about the Gardener, Franzke is not sure about the next release date. 

> Hi, not sure yet, https://github.com/gardener/gardener/pull/721 is definitely needed, so definitely not before this PR was merged


About remove pods and services network in dashboard, shall we do it together with dashboard. Sorry for the inconvenience brought to you.

jia-jerry:
I will do the fix in Gardener first.

grolu:
Hi Jerry, I already removed the defaults in the dashboard: #304

jia-jerry:
@grolu Thanks a lot.

PR #328, opened 2019-03-13
Remove OIDC dependency

### LINKED ISSUE #272: Username != Email
There are many places in the code (both backend and frontend) where we expect that the username equals the email address of the user. This is not always the case (e.g. Azure), therefore we need to walk through the code and fix this.
In addition, the frontend expects that the username is always an email address (e.g. email links in the member list), we should check if this is the case (e.g. regex check).

### LINKED ISSUE #285: Allow login with Token / Kubeconfig
Currently, an OIDC provider is required to login to the dashboard. A first step to allow different login methods is to provide an input field to access the dashboard with an access token. Also, we need to remove checks in the frontend that evaluate if the token is a valid Kubernetes token which it is in our current setup, however this is not an requirement and not always the case. The backend acts as proxy and uses the provided token to access the APIServer where the actual validation is done. However, we need to pay attention when we use the system account (admin) to make calls.

@holgerkoser please correct and complement the text above

### LINKED ISSUE #329: Retain Navigation State after Login
We keep most of the navigation state in the URL (including pages and tabs, but currently excluding dialogs and form data). This way, it is possible to send e.g. links, that point to a specific cluster of the gardener Dashboard.
However, if the user that opens the URL is currently not logged in to the dashboard, he is redirected to the login page and looses navigation information: After successful authentication, the user is redirected to the start page (which is - for most cases - the cluster overview).
We need to improve here, especially as this issue will increase after #328 gets merged. This PR will significantly reduce the token lifetime and lead to more authentication and redirects to the start page.
Moreover, we should discuss if we want to include more information (about open dialogs, unsent form input, etc.) into the URL, or store it in local browser store.

holgerkoser:

Fixes #285 Fixes #329 Fixes #272

**Release note**:

```improvement user

```


backend/lib/auth.js
```diff
@@ -0,0 +1,68 @@
...
+  .get(async (req, res, next) => {
+    try {
+      res.redirect(await authorizationUrl(req, res))
+    } catch (err) {
+      next(err)
+    }
+  })
+  .post(async (req, res, next) => {
+    try {
+      res.send(await authorizeToken(req, res))
```
holgerkoser: handle error with redirect

backend/lib/services/authorization.js
```diff
@@ -29,12 +29,19 @@ async function hasAuthorization (user, resourceAttributes) {
...
+      resourceAttributes
+    }
+  }
   const response = await Authorization(user).selfsubjectaccessreviews.post({ body })
   return _.get(response, 'status.allowed', false)
 }
+exports.hasAuthorization = hasAuthorization
 
-exports.isAdmin = async function (user) {
+exports.isAdmin = function (user) {
```
holgerkoser: async if promise is already returned?

backend/lib/services/cloudprofiles.js
```diff
@@ -17,60 +17,70 @@
...
+function getSeedsForCloudProfileName ({ seeds, cloudProfileName }) {
+  return _
+    .chain(seeds)
+    .filter(['spec.cloud.profile', cloudProfileName])
+    .map(fromSeedResource)
+    .value()
 }
 
-exports.list = async function () {
+exports.list = function () {
```
holgerkoser: service methods should always be considered async

backend/lib/security.js
```diff
@@ -0,0 +1,339 @@
...
+    encryptedText
+  ])
+  return base64url.encode(buffer)
+}
+
+async function decrypt (data) {
+  const buffer = base64url.toBuffer(data)
+  let start
+  let end = 0
+  // salt
```
petersutter: rm comments except // decrypt text

petersutter:
Topic presented by @holgerkoser in Gardener Community Meeting. Recording: https://youtu.be/MHQF93nBVfA?t=1925s

PR #331, opened 2019-03-15
Fix regex named groups error in Firefox

grolu:
Use old style index based regex access as firefox does not support ES2018 named groups (yet)

Fixes #

**Release note**:

```improvement user
NONE
```

### DIFF
--- a/frontend/src/components/HibernationSchedule.vue
+++ b/frontend/src/components/HibernationSchedule.vue
@@ -128,23 +128,19 @@ export default {
     reset () {
       this.parseSchedules(this.scheduleCrontab)
     },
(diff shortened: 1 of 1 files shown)


frontend/src/components/HibernationSchedule.vue
```diff
@@ -128,11 +128,19 @@ export default {
     reset () {
       this.parseSchedules(this.scheduleCrontab)
     },
+    scheduleEventObjFromRegex (regex) {
```
holgerkoser: The name of the parameter should be something like `crontabLine`. I think the `console.warn` should be done in the `scheduleEventObjFromRegex` function if the entry exists but could not be parsed.

PR #339, opened 2019-03-25
Retain Navigation State after Login

### LINKED ISSUE #329: Retain Navigation State after Login

petersutter:
See #329. Probably we will not merge this as we have to redo most parts in the new `auth` branch but we can use this PR as starting point

Fixes #329

@holgerkoser 

**Release note**:

```improvement user

```

### DIFF
--- a/frontend/src/layouts/Login.vue
+++ b/frontend/src/layouts/Login.vue
@@ -68,7 +68,8 @@ export default {
   },
   methods: {
     handleLogin () {
-      signin(this.$userManager)
+      const { redirectTo } = this.$route.query
+      signin(this.$userManager, redirectTo)
         .catch(error => {
           this.showSnotifyLoginError(error.message)
           throw error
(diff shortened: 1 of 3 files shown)


petersutter:
closed as already part in #328

PR #345, opened 2019-04-03
Expose logging credentials for users

KristianZH:
This PR depends on gardener/gardener/pull/851

Expose Logging credentials for Kibana to the users

Fixes #

@holgerkoser 
**Release note**:
-->
```improvement user
NONE
```

### DIFF
--- a/backend/lib/services/shoots.js
+++ b/backend/lib/services/shoots.js
@@ -42,8 +42,8 @@ function getSecret (core, namespace, name) {
     })
 }
 
-async function assignComponentSecret (core, namespace, component, data) {
-  const secret = await getSecret(core, namespace, `${component}-ingress-credentials`)
+async function assignComponentSecret (core, namespace, component, secretName, data) {
+  const secret = await getSecret(core, namespace, secretName)
   if (secret) {
     _
       .chain(secret)
@@ -216,6 +216,13 @@ exports.info = async function ({ user, namespace, name }) {
   const namespaces = core.namespaces
   const project = await getProjectByNamespace(projects, namespaces, namespace)
   const projectName = project.metadata.name
+
+  const monitoringComponent = 'monitoring'
+  const loggingComponent = 'logging'
+  const monitoringIngressSecretName = 'monitoring-ingress-credentials'
+  const loggingIngressUserSecretName = name + '.logging'
+  const loggingIngressAdminSecretName = 'logging-ingress-credentials'
+
   const data = {
     seedShootIngressDomain: `${name}.${projectName}.${ingressDomain}`
   }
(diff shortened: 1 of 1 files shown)


holgerkoser:
Is the secret `logging-ingress-user-credentials` located in the garden cluster in the project namespace? According to the coding it is! If yes, why is there only one secret for all shoots? I don't understand that.

vpnachev:
It should be one secret per shoot, i.e. `<shoot-name>.logging-ingress-user-credentials` just like the secret containing the kubeconfig file(`<shoot-name>.kubeconfig`). So if I am not wrong, the PR should be adapted to this naming convention.

holgerkoser:
Who copies this secret `<shoot-name>.logging-ingress-user-credentials` into the project namespace of the garden cluster? Why don't you use a secret in the seed cluster as we do it for the administrator?

vpnachev:
The GCM syncs the secret from the seed to the Garden cluster.
Now we have two users(`admin` and `user`) for the kibana dashboard. For the `admin` user nothing is changed, while the `user` user will be visible to the shoot owners and they will be able to see the logs of their `kube-apiserver` and `kube-controller-manager` pods. 

The problem is that the dashboard cannot sync secrets from the seed clusters for a user without the cluster role `garden.sapcloud.io:system:administrators`. This is why we decided to copy the secret to the Garden cluster.

Of course, we have no experience with the dashboard development, so what we have done may not be the best solution. Thus we are open for suggestion how to get a secret from a seed cluster for a non `garden.sapcloud.io:system:administrators` user.

KristianZH:
Thank to @vpnachev for replying the comments. The secret will be added in the garden cluster after this PR is merged: gardener/gardener/pull/851

backend/lib/services/shoots.js
```diff
@@ -277,11 +284,15 @@ exports.info = async function ({ user, namespace, name }) {
...
         await Promise.all([
-          assignComponentSecret(core, seedShootNS, 'monitoring', data),
-          assignComponentSecret(core, seedShootNS, 'logging', data)
+          assignComponentSecret(core, seedShootNS, monitoringComponent, monitoringIngressSecretName, data),
+          assignComponentSecret(core, seedShootNS, loggingComponent, loggingIngressAdminSecretName, data),
         ])
       }
     }
+  } else {
+    await Promise.all([
```
holgerkoser: please do not use promise all for only one promise:
```
await assignComponentSecret(core, namespace, loggingComponent, loggingIngressUserSecretName, data)
```

KristianZH:
@holgerkoser I think that this PR is ready to be merged.

PR #371, opened 2019-04-30
Update node.js to v10.15.3 and alpine to v3.9

holgerkoser:
The official node 10-alpine image has been updated to version  v10.15.3 and with this update the alpine base image has also been updated to version 3.9. But our production base image was still alpine 3.8. This situation needs to be avoided in the future. With this PR the node image version is fully specified and has to be updated explicitly. It is important that we always use the same base image as the node alpine image.

Fixes #

**Release note**:

```improvement operator
Update node.js to v10.15.3 and alpine OS to v3.9
```

### DIFF
--- a/backend/test/dockerfile.spec.js
+++ b/backend/test/dockerfile.spec.js
@@ -16,7 +16,6 @@
 
 'use strict'
 
-const assert = require('assert').strict
 const fs = require('fs')
 const path = require('path')
 const _ = require('lodash')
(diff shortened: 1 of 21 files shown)


backend/test/dockerfile.spec.js
```diff
@@ -0,0 +1,65 @@
...
+        const dashboardDockerfile = await getDashboardDockerfile()
+        
+        expect(dashboardDockerfile.getFROMs()).to.have.length(4)
+        const buildStages = _
+            .chain(dashboardDockerfile.getFROMs())
+            .map(from => [from.getBuildStage(), from])
+            .fromPairs()
+            .value()
+        const imageTag = buildStages.base.getImageTag()
+        const [, nodeVersion] = /^(1[0-9])\-alpine$/.exec(imageTag) || []
```
petersutter: only allow 10 and 12 and add comment for future release upgrades

PR #382, opened 2019-05-14
Respect the value of the ignore annotation

petersutter:
Until now the shoot ignore (is reconcile deactivated) check was to check if shoot.garden.sapcloud.io/ignore annotation is `'true'`. This PR modifies the check to respect the value of the annotation. Truthy values are `'1'`, `'t'`, `'T'`, `'true'`, `'TRUE'`, `'True'`.

Fixes #

See related PR in gardener https://github.com/gardener/gardener/pull/878

**Release note**:

```improvement user
Improved check if cluster reconciliation is deactivated
```

### DIFF
--- a/frontend/src/utils/index.js
+++ b/frontend/src/utils/index.js
@@ -368,7 +368,8 @@ export function shootHasIssue (shoot) {
 }
 
 export function isReconciliationDeactivated (metadata) {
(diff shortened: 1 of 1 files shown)


petersutter:
yes those are the exact values, the `ParseBool` function is used in gardener, see https://github.com/golang/go/blob/master/src/strconv/atob.go#L12

PR #399, opened 2019-05-28
Make identity tls secret configurable

Diaphteiros:
If you create a secret containing a valid certificate for the dashboard and that secret is maintained/updated by another instance - as for example the case with the cert-manager - there is currently no nice way to use this secret as tls secret for identity. The identity deployment will either overwrite it, or you copy the data from that secret to the `identity-tls` secret and have to keep it in sync, which requires an extra controller.

This PR does two things:
First, it makes the deployment of the `identity-tls` secret optional. By setting `tls` in the values file to something that helm evaluates to `false`, the secret won't be deployed anymore. Take care that you'll have to provide the secret yourself in that case, otherwise you'll be left with an incorrect configuration.
Second, a new value `tlsSecretName` in the values file allows to configure the name of the secret. This was done so the component that creates the secret (if not deployed by identity) does not have to rely on a name hardcoded in the identity helm chart, but can guarantee that identity uses the secret it creates.

None

None

**Release note**:

```improvement operator
The identity helm chart can now be configured to not deploy the tls secret. Furthermore, the name of that secret can be configured by overwriting `tlsSecretName` in the values file.
```

### DIFF
--- a/charts/identity/templates/ingress.yaml
+++ b/charts/identity/templates/ingress.yaml
@@ -16,7 +16,7 @@ metadata:
 {{- $pathname := regexReplaceAll "https://[^/]*(?:/(.+))?" $.Values.issuerUrl "/${1}" }}
 spec:
   tls:
-    - secretName: identity-tls
+    - secretName: {{ .Values.tlsSecretName | default "identity-tls" }}
       hosts:
         - {{ $hostname }}
   rules:
--- a/charts/identity/templates/secret.yaml
+++ b/charts/identity/templates/secret.yaml
@@ -2,7 +2,7 @@
 apiVersion: v1
 kind: Secret
 metadata:
-  name: identity-tls
+  name: {{ .Values.tlsSecretName | default "identity-tls" }}
   namespace: {{ .Release.Namespace }}
   labels:
     app: identity
--- a/charts/identity/templates/secret.yaml
+++ b/charts/identity/templates/secret.yaml
@@ -1,3 +1,4 @@
+{{- if .Values.tls }}
 apiVersion: v1
 kind: Secret
 metadata:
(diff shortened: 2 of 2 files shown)


Diaphteiros:
> I don't understand why changing the name of the secret make a difference.

@holgerkoser We can also drop that commit if you don't like it. The reasoning behind it is as follows:
If I have a component that is supposed to create the tls secret for identity, I have no nice way to let the component find out which name that secret is supposed to have. I can make it create a secret called `identity-tls`, but that value is hardcoded in the identity helm chart and if it ever changes for whatever reason, my deployment will fail. If the secret name is configurable, I can choose a name and provide that name to my component as well as identity and by this make sure that both will always refer to the same secret.

holgerkoser:
> > I don't understand why changing the name of the secret make a difference.
> 
> @holgerkoser We can also drop that commit if you don't like it. The reasoning behind it is as follows:
> If I have a component that is supposed to create the tls secret for identity, I have no nice way to let the component find out which name that secret is supposed to have. I can make it create a secret called `identity-tls`, but that value is hardcoded in the identity helm chart and if it ever changes for whatever reason, my deployment will fail. If the secret name is configurable, I can choose a name and provide that name to my component as well as identity and by this make sure that both will always refer to the same secret.

holgerkoser:
I accidentally closed the PR.

holgerkoser:
@Diaphteiros I think it would make sense to configure the secret name used in the ingress and in the tls secret independently. You may don't want to create the secret but specify certmanager secret in the ingress.

Diaphteiros:
@holgerkoser I changed the charts to use helms `default` operator. Now no changes in the values.yaml file are required.
I don't think that it is necessary to modify name of the secret and the reference independently - you either want to create the secret and reference it (using the same name for both), or you want to reference another secret and can just disable the secret creation.

PR #414, opened 2019-06-19
Login session related issues

### LINKED ISSUE #335: Timeout Notification and Reconnect
When the new auth logic gets merged (#328) we will have tokens with significantly shorter lifetime. This will lead to more situations where the user gets logged out, e.g. because he is inactive. However, there are also situations where the user interacts with the dashboard and the session times out nonetheless. An example where this could be happen, are terminals. The reason for this is, that terminals use a different communication channel than the dashboard http calls.
We should at least inform the user about timeouts, or even better proactively show a notification with a renew button (or "stay logged in") button, _before_ the session times out. We could leverage our existing notification framework ("Snotify") for that.
Moreover, we could try to keep sessions alive, when we detect user activity (e.g. on terminal windows) even when no http calls are triggered.

### LINKED ISSUE #407: Fix deprecation warnings with node v12
> [DEP0123] DeprecationWarning: Setting the TLS ServerName to an IP address is not permitted by RFC 6066. This will be ignored in a future version.

Caused by WebSocket watches if the apiserver hostname is an IP address. I have opened a [PR 1587](https://github.com/websockets/ws/pull/1587) for websocket library.

> [DEP0066] DeprecationWarning: OutgoingMessage.prototype._headers is deprecated. 

Caused by `openid-client`. Major version upgrade required. Some changes in our code have to be done.

> [before-after-hook]: "Hook()" repurposing warning, use "Hook.Collection()". Read more: https://git.io/upgrade-before-after-hook-to-1.4

Caused by `@octokit/rest@15.18.1` which depends on `before-after-hook@1.1.0`. Upgrade to version 1.4.0 requires upgrade of major version `@octokit/rest@16.28.1`. This requires many changes in our code :-(

### LINKED ISSUE #408: no redirect to login page on 401 when fetching user priviliges
when `/api/user/privileges` returns 401 the dashboard does not redirect to the logon screen. Instead, just an empty page is shown

holgerkoser:
This PR fixes some issues related to the login process and the login session lifetime.
* If the frontend session cookie is valid but the Users JWT token is not valid or the user is not authorized to access the user privileges the user is not redirected to the login screen and an error message is displayed. This fixes #408.
* The frontend session cookie does not expire due to inactivity anymore. It is bound to the browser session not persistent. This fixes #335.
* If the user token is expired the dashboard does an automatic logout and redirect the user back to the login page. 

Node version and package dependency upgrades.
* Some dependencies have been updated because of deprecation warnings. This fixes #407.
* Node version has been updated to version 12 because the [openid client](https://github.com/panva/node-openid-client) we use requires at least that version.

Fixes #335, #407 and #408,

**Release note**:

```improvement user
User session does not expire after 30 minutes
```

```improvement user
If the users token is expired he will be automatically redirected to the login page
```

```noteworthy operator
The configuration value `cookieMaxAge` in the dashboard helm chart is not used anymore and has been removed
```


backend/lib/kubernetes/watch.js
```diff
@@ -150,10 +151,12 @@ function createWebSocket (resource, options = {}) {
   url.search = querystring.stringify(assign(qs, resource.qs, options.qs, { watch: true }))
   url.pathname = resource.path
   const { key, cert, ca, strictSSL, auth = {} } = get(api, 'http.requestOptions', {})
-  const origin = get(options, 'origin', api.url)
   const rejectUnauthorized = get(options, 'rejectUnauthorized', !strictSSL)
   const headers = {}
   const websocketOptions = { origin, headers, rejectUnauthorized, key, cert, ca }
+  if (net.isIP(url.hostname) !== 0) {
+    websocketOptions.servername = ''
```
holgerkoser: comment with link to docs

backend/lib/security.js
```diff
@@ -19,17 +19,17 @@
 const { split, join, noop, trim, isPlainObject } = require('lodash')
 const { promisify } = require('util')
 const jwt = require('jsonwebtoken')
-const { Issuer } = require('openid-client')
+const { Issuer, custom } = require('openid-client')
 const cookieParser = require('cookie-parser')
-const { JWK, JWE } = require('node-jose')
+const { JWK, JWE } = require('@panva/jose')
```
holgerkoser: check ack

backend/test/acceptance/security.spec.js
```diff
@@ -18,17 +18,19 @@
...
   it('should return the oidc issuer client', async function () {
     oidc.stub.getIssuerClient()
     const client = await security.getIssuerClient()
-    expect(client.issuer.httpOptions()).to.include({
+    expect(client[custom.http_options]()).to.include({
       followRedirect: false,
       rejectUnauthorized,
       ca
     })
+    expect(client[custom.clock_tolerance]).to.equal(15)
```
holgerkoser: do not use default in test

frontend/src/router/index.js
```diff
@@ -277,6 +278,20 @@ export default function createRouter ({ store, userManager }) {
...
   const routerOptions = { mode, scrollBehavior, routes }
 
+  /* automatic signout when token expires */
+  let timeoutID
+  store.watch((state, getters) => getters.userExpiresAt, expirationTime => {
+    if (timeoutID) {
+      clearTimeout(timeoutID)
+    }
+    const currentTime = Date.now()
+    if (expirationTime && expirationTime > currentTime) {
```
holgerkoser: console.error in else

frontend/src/router/index.js
```diff
@@ -311,6 +326,15 @@ export default function createRouter ({ store, userManager }) {
         query
       })
     } catch (err) {
+      const { response: { status, data = {} } = {} } = err
+      if (status === 401) {
+        userManager.removeUser()
+        const hash = data.message ? `#error=${encodeURIComponent(data.message)}` : undefined
```
holgerkoser: redirect to logout and forward error

backend/test/acceptance/security.spec.js
holgerkoser: ✔️

backend/lib/security.js
holgerkoser: ✔️

backend/lib/kubernetes/watch.js
holgerkoser: ✔️

frontend/src/router/index.js
holgerkoser: ✔️

frontend/src/router/index.js
holgerkoser: ✔️

PR #416, opened 2019-06-25
Refactor create shoot

grolu:
**What this PR does / why we need it**:
This PR replaces the Create Cluster Popup with a dedicated create cluster page. The user can now switch between UI and yaml mode when creating clusters (using the same UI paradigm as on the cluster details page). This solves several issues by allowing the user to configure the shoot resource before creating the cluster (which is required for some things, e.g. dns settings)

**Which issue(s) this PR fixes**:
Fixes #208 Fixes #340 Fixes #302 Fixes #23 Fixes #432 Fixes #390 Fixes #415 Fixes #435

**Special notes for your reviewer**:
Screen Shot 2019-06-28 at 09 10 58
Screen Shot 2019-06-28 at 09 11 25

**Release note**:

```noteworthy user
- Create Cluster: Replaced create cluster popup with a dedicated create cluster page. You can now switch between UI and yaml mode when creating clusters. This allows to fine-tune the shoot resource by making custom configurations before the cluster is created #23
- Worker Groups: You can now configure `Machine Image` and `Max. Surge` via the Dashboard UI
```
```improvement user
- Maintenance: You can now disable automatic operating system updates of your workers #390
- Availability Zones: You can now configure multiple availability zones for your cluster #432
- Add-ons: Manage cluster add-ons on details page #302
- Purpose: Easy change the purpose of a cluster after creation
```
```improvement operator
- Actions for clusters with purpose `infrastructure` are now disabled to avoid accidental operations on infrastructure clusters
```
```action operator
:warning: This Dashboard version is not compatible with Gardener versions prior 0.27.x
```

PR #428, opened 2019-07-04
Machine/Volume type drop list now changes with the zone selected for alicloud

EmoinLanyu:
Machine/Volume type drop list now changes in accordance to the zone selected in shoot creation dialog for alicloud.
Remove invalid machine/volume types when zone is changed in shoot creation dialog for alicloud profile

_why we need it_

In Alicloud (not sure if other cloud providers have the same problem), different zones may have different available machine types and volume types. However, current create shoot cluster dialog code does not filter available types in worker tab's drop list.
This will cause the problem that when a user selects a zone and configures its workers, some machine types and volume types that this specific zone does not support are listed in the drop lists. The user can select such types, however, when clicking create button, error messages are shown, indicating that these types are not supported.

Alicloud cloud profile example: https://github.com/gardener/gardener/blob/master/example/30-cloudprofile-alicloud.yaml
Note that each machine type and volume type can have its individual zone setting.

_what this pr does_
- The changes are currently only for alicloud.
- The machine type drop list and volume type drop list now changes according to the zone selected.
- When user changes the zone in create shoot cluster dialog, corresponding changes in worker tab occur. The changes are listed below:
1. The options in machine type drop list and volume type drop list change in accordance to the selected zone.
2. If previously selected machine type or volume type is unavailable in the new zone, the selection will be revoked, and the worker group setting becomes invalid, and thus the shoot creation button becomes disabled. User has to re-select the types to be able to create the cluster.

As can be seen in the code, alicloud is specified for this change. I'm not sure if it is also needed by other cloud providers, and if all the cloud profiles can be dealt with the same way.

**Release note**:

```improvement user

```

### DIFF
--- a/frontend/src/components/MachineType.vue
+++ b/frontend/src/components/MachineType.vue
@@ -70,6 +70,7 @@ export default {
       this.validateInput()
     },
     validateInput () {
+      this.valid = this.valid && this.worker.valid && !this.$v.$invalid
       if (this.valid !== !this.$v.$invalid) {
         this.valid = !this.$v.$invalid
         this.$emit('valid', { id: this.worker.id, valid: this.valid })
--- a/frontend/src/components/ManageWorkers.vue
+++ b/frontend/src/components/ManageWorkers.vue
@@ -21,7 +21,9 @@ limitations under the License.
         ref="workerInput"
         :worker="worker"
         :workers="internalWorkers"
+        :infrastructureKind="infrastructureKind"
         :cloudProfileName="cloudProfileName"
+        :zone="zone"
         @updateName="onUpdateWorkerName"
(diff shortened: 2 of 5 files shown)


grolu:
Thank you for your contribution. I will have a look at it. However, as we refactored the Create Cluster Dialog (#416) I will need to integrate your changes into the new logic.

EmoinLanyu:
> Thank you for your contribution. I will have a look at it. However, as we refactored the Create Cluster Dialog (#416) I will need to integrate your changes into the new logic.

Absolutely no problem. Thanks!

grolu:
Closed in favor of #428 (part of PR #416)

PR #442, opened 2019-08-09
Disclaimer for Secrets Creation Dialog

### LINKED ISSUE #430: Disclaimer for Secrets Creation Dialog
**What?**
Can we add a disclaimer field to the secrets creation dialog stating something like: The user understood that he is not going to use personalised login credentials and if so that he knows exactly what he is doing? 

**Why?**
In case of an OpenStack setup, where the user base comes from lets say a corporate LDAP, you want to inform the user that he should not use his personal login credentials as an infrastructure secret.

grolu:
Show Disclaimer for Secrets Creation Dialog

Fixes #430

Screen Shot 2019-08-09 at 15 39 37

**Release note**:

```improvement user

```

### DIFF
--- a/frontend/src/dialogs/SecretDialog.vue
+++ b/frontend/src/dialogs/SecretDialog.vue
@@ -28,7 +28,7 @@ limitations under the License.
               <infra-icon v-model="infraIcon" :width="42"></infra-icon>
             </v-flex>
             <v-flex>
-              <div class="credential_title">{{title}}</div>
+              <div class="credential_title ml-3">{{title}}</div>
             </v-flex>
           </v-layout>
         </v-container>
@@ -69,9 +69,10 @@ limitations under the License.
 
           <slot name="data-slot"></slot>
           <alert color="error" :message.sync="errorMessage" :detailedMessage.sync="detailedErrorMessage"></alert>
+          <p class="subheading mt-3">Do not use personalised login credentials as they should not be used as infrastructure secret.</p>
         </v-container>
       </v-card-text>
-      <v-alert :value="!isCreateMode && relatedShootCount > 1" type="warning">
+      <v-alert :value="!isCreateMode && relatedShootCount > 0" type="warning">
         This secret is used by {{relatedShootCount}} clusters. The new secret should be part of the same account as the one that gets replaced.
       </v-alert>
       <v-card-actions>


grolu:
Should we show this for all infra secret dialogs?

kayrus:
@grolu it will be also great to add a configurable disclaimer to the **domain name**. E.g. in CCloud gardener will have its own domain, and if you put a different domain, the cluster will be created, but the connectivity between the seed and the shoot won't work.
/cc @afritzler

afritzler:
This is related to the issue in the extension repo: https://github.com/gardener/gardener-extensions/issues/154

Currently we are in the progress of adding fields
- `DomainID` (or `DomainName`, at least one has to be provided)
- `TenantID` (or `TenantName`, at least one has to be provided)
- `UserDomainName` (optional)

afritzler:
Just saw that this is a disclaimer PR. Will create a dedicated issue for that to track the progress.

frontend/src/dialogs/SecretDialog.vue
```diff
@@ -69,9 +69,10 @@ limitations under the License.
 
           <slot name="data-slot"></slot>
           <alert color="error" :message.sync="errorMessage" :detailedMessage.sync="detailedErrorMessage"></alert>
+          <p class="subheading mt-3">Do not use personalised login credentials as they should not be used as infrastructure secret.</p>
```
grolu:
```suggestion
          <p class="subheading mt-3">Do not use personalized login credentials as they should not be used as secret.</p>
```

frontend/src/dialogs/SecretDialog.vue
```diff
@@ -69,9 +69,10 @@ limitations under the License.
 
           <slot name="data-slot"></slot>
           <alert color="error" :message.sync="errorMessage" :detailedMessage.sync="detailedErrorMessage"></alert>
+          <p class="subheading mt-3">Do not use personalized login credentials as they should not be used as secret.</p>
```
afritzler: Shouldn't we rather point out here to use a technical user? So something like "It is not recommended to use personalized login credentials here. Please use the credentials of a technical user."

grolu: Yes what do the others think? @petersutter @holgerkoser

petersutter: I would phrase it as follows:
"Do not use personalized login credentials. Instead, use the credentials of a technical user.

grolu:
```suggestion
          <p class="subheading mt-3">Do not use personalized login credentials. Instead, use the credentials of a technical user.</p>
```

PR #444, opened 2019-08-12
Only display Alertmanager link when shoot has an alertmanager

### LINKED ISSUE #423: Suppress Alertmanager Link (show msg instead) When Disabled/Not Deployed
Monitoring doesn't deploy the alertmanager always since a couple of weeks. It does so only if an `alertTo` annotation is set. If it's not set, maybe we should either suppress the link or show an info msg instead, that there is no alertmanager, because the user has not configured to receive alerts.

wyb1:
Only shows a link to the Alertmanager if the shoot is annotated with the `garden.sapcloud.io/operatedBy` annotation.

Fixes #423 

`isAdmin` is removed because users should also have access to an Alertmanager if they annotated their shoot.
**Release note**:

```improvement user
NONE
```

### DIFF
--- a/frontend/src/components/ClusterMetrics.vue
+++ b/frontend/src/components/ClusterMetrics.vue
@@ -60,7 +60,7 @@ limitations under the License.
         </v-list-tile-title>
       </v-list-tile-content>
     </v-list-tile>
-    <v-list-tile v-if="isAdmin">
+    <v-list-tile v-if="hasAlertmanager">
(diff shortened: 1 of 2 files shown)


grolu:
@wyb1 PR is ok
However, alertmanager can currently only be accessed with admin credentials. We will merge this PR as soon as alertmanager can be accessed with user credentials

wyb1:
Rebased and resolved conflicts

PR #496, opened 2019-11-12
Remove deprecated addon monocular

petersutter:
Gardener has deprecated all addons except kubernetes-dashboard and nginx-ingress, hence we should remove the deprecated addons from the create cluster list.

Fixes #

see gardener/gardener#471

**Release note**:

```improvement user
Removed deprecated addon `Monocular ` from cluster details page. (See also gardener/gardener#471)

```


petersutter:
cc @rfranzke

PR #500, opened 2019-11-14
removed vm.nexttick workarounds where possible

### LINKED ISSUE #491: Deal with vm.nexttick workarounds
Although we already got rid of most `vm.nexttick`workarounds, there still plenty of them in the code. We need to check all of them if they are (still) required and if yes add a comment so that we area aware of this and more confident when changing code in those places.

grolu:
Checked all vm.nextticks we have in the frontend, removed where possible, added comments otherwise

Fixes #491

**Release note**:

```improvement user

```

### DIFF
--- a/frontend/src/components/ShootWorkers/ManageWorkers.vue
+++ b/frontend/src/components/ShootWorkers/ManageWorkers.vue
@@ -133,15 +133,13 @@ export default {
     },
     onRemoveWorker (index) {
       this.internalWorkers.splice(index, 1)
-      this.$nextTick(() => {
-        // Need to evaluate the other components as well, as valid state may depend on each other (e.g. duplicate name)
-        // Lack of doing so can lead to worker valid state != true even if conflict has been resolved, if input happens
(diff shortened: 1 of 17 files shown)


frontend/src/components/ShootWorkers/ManageWorkers.vue
```diff
@@ -133,33 +133,34 @@ export default {
...
-      const wasValid = worker.valid
-      worker.valid = valid
-      if (valid !== wasValid) {
-        // Need to evaluate the other components as well, as valid state may depend on each other (e.g. duplicate name)
-        // Lack of doing so can lead to worker valid state != true even if conflict has been resolved, if input happens
-        // on component which did not report valid = false in the first place
-        forEach(this.$refs.workerInput, workerInput => {
-          workerInput.validateInput()
-        })
+      if (worker) {
```
grolu: return

grolu: done

PR #542, opened 2020-01-10
Fix sorting edge case

### LINKED ISSUE #494: Fix sorting for edge cases
For sorting the cluster list, we sometimes use different values that we concatenate and then sort it. This works for most cases but if we have similar values sorting might not be accurate. We should rather sort by the values independently.

grolu:

Fixes #494

**Release note**:

```improvement user
NONE
```


frontend/tests/unit/store.shoots.spec.js
```diff
@@ -0,0 +1,206 @@
...
+//
+
+import Vuex from 'vuex'
+import { expect } from 'chai'
+import { state, actions, getters, mutations, modules } from '@/store'
+
+let store
+
+describe('Store.Shoots', function () {
+  beforeEach(function (done) {
```
petersutter: rm done

PR #602, opened 2020-02-18
Introduce seed-info endpoint

### LINKED ISSUE #581: Seperate shoot info endpoint
Currently, with the shoot info endpoint we fetch the kubeconfig from the garden cluster and monitoring / logging credentials from the seed.
For seeds behind a firewall that can't be reached by the dashboard backend the request to the info endpoint takes very long to return because we run in a timeout when trying to talk to the seed kube-apiserver. The consequence is that also the access details (kubconfig) for this cluster is only shown after the timeout.
We should split the info endpoint, so that we have one endpoint to fetch data from the garden cluster and one (for admins) to fetch data from the seed.

petersutter:
This PR introduces a `shoots/:name/seed-info` endpoint which is used to fetch data from the seed cluster, which was previously done in the `shoots/:name/info` endpoint. See #581 for more details.

Fixes #581 

**Release note**:

```improvement operator
The `kubeconfig` on the cluster details page is shown much faster for clusters of which the seed is isolated / air-gapped
```

### DIFF
--- a/backend/lib/services/shoots.js
+++ b/backend/lib/services/shoots.js
@@ -297,18 +297,34 @@ exports.info = async function ({ user, namespace, name }) {
   }
 
   const isAdmin = await authorization.isAdmin(user)
-  if (isAdmin) {
-    if (seed) {
-      try {
-        const seedClient = await client.createKubeconfigClient(seed.spec.secretRef)
-        const seedShootNamespace = shoot.status.technicalID
-        await assignComponentSecrets(seedClient, data, seedShootNamespace)
-      } catch (err) {
-        logger.error('Failed to retrieve information using seed core client', err)
-      }
-    }
-  } else {
+  if (!isAdmin) { // TODO return (user) component secrets also for admin?
     await assignComponentSecrets(client, data, namespace, name)
+  }
+
+  return data
+}
+
+exports.seedInfo = async function ({ user, namespace, name }) {
+  const client = user.client
+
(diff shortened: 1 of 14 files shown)


frontend/src/utils/Emitter.js
```diff
@@ -201,10 +201,16 @@ class ShootSubscription extends AbstractSubscription {
     this.socket.on('shootSubscriptionDone', async ({ kind, target }) => {
       const { name, namespace } = target
       throttledNsEventEmitter.flush()
+      const promises = [
+        store.dispatch('getShootInfo', { name, namespace }),
```
grolu: unexpected trailing comma

petersutter: ✅

backend/lib/services/shoots.js
```diff
@@ -297,23 +297,39 @@ exports.info = async function ({ user, namespace, name }) {
...
-      try {
-        const seedClient = await client.createKubeconfigClient(seed.spec.secretRef)
-        const seedShootNamespace = shoot.status.technicalID
-        await assignComponentSecrets(seedClient, data, seedShootNamespace)
-      } catch (err) {
-        logger.error('Failed to retrieve information using seed core client', err)
-      }
-    }
-  } else {
+  if (!isAdmin) { // TODO return (user) component secrets also for admin?
```
grolu:
```suggestion
  if (!isAdmin) {
```

PR #605, opened 2020-02-20
Webterminals: Support kube-apiserver port other than 443

petersutter:
If the kube-apiserver is running on a port other than 443 the browser can't connect to the kube-apiserver.
With this PR, we use the port that is defined in the client config / kubeconfig of the respective cluster.

Fixes #

**Release note**:

```improvement user

```


petersutter:
cc @EmoinLanyu

PR #607, opened 2020-02-25
Add journal polling feature

### LINKED ISSUE #599: Regularly Pull Journals in the Gardener Dashboard
Gardener landscapes which are behind a firewall have the issue that after a journal was created via the Dashboard, the journal is not automatically updated via the GitHub webhook functionality. Only after restarting the Dashboard pod this happens. A regular pull is required in the Dashboard to regularly  update journals even when between the Dashborad and the GitHub with the journals is a firewall.

petersutter:
When the dashboard starts up it initially fetches all the issues from github and receives updates of the issues and comments using the github webhook functionality.
However, if the dashboard's github webhook can't be reached by github due to network isolation, we need to regularly poll the issues from github.

This PR introduces a new config to set the optional `pollIntervalSeconds` field.
Example `values.yaml` file:
```yaml
gitHub:
  ...
  pollIntervalSeconds: 30
```

Fixes #599

**Release note**:

```improvement operator
The dashboard now support polling GitHub issues in case the dashboard's webhook is not reachable from GitHub due to network isolation
```

### DIFF
--- a/backend/lib/watches/journals.js
+++ b/backend/lib/watches/journals.js
@@ -22,10 +22,8 @@ const journals = require('../services/journals')
 const cache = require('../cache')
 const config = require('../config')
 const delay = require('delay')
-const pForever = require('p-forever')
-const _ = require('lodash')
 
-module.exports = (io, retryOptions = {}) => {
+module.exports = async (io, retryOptions = {}) => {
   if (!config.gitHub) {
     logger.warn('Missing gitHub property in config for journals feature')
     return
@@ -73,20 +71,21 @@ module.exports = (io, retryOptions = {}) => {
   if (isNaN(pollIntervalSeconds)) {
     pollIntervalSeconds = undefined
   }
-  return pForever(async () => {
+
+  do {
     await loadAllOpenIssues()
     if (!pollIntervalSeconds) {
-      return pForever.end
+      break
     }
 
     const issueNumbers = journalCache.getIssueNumbers()
-    await _.forEach(issueNumbers, async number => {
+    for (const number of issueNumbers) {
       try {
         await journals.loadIssueComments({ number })
       } catch (err) {
         logger.error('failed to fetch comments for reopened issue %s: %s', number, err)
       }
-    })
+    }
     await delay(pollIntervalSeconds * 1000)
-  })
+  } while (true)
 }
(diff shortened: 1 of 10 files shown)


petersutter:
cc @ThormaehlenFred

backend/lib/watches/journals.js
```diff
@@ -66,5 +69,24 @@ module.exports = (io, retryOptions = {}) => {
...
+    pollIntervalSeconds = undefined
+  }
+  return pForever(async () => {
+    await loadAllOpenIssues()
+    if (!pollIntervalSeconds) {
+      return pForever.end
+    }
+
+    const issueNumbers = journalCache.getIssueNumbers()
+    await _.forEach(issueNumbers, async number => {
```
holgerkoser: does not work as expected. Use normal for loop
```js
for (const number of issueNumbers) {
  try {
    await journals.loadIssueComments({ number })
  } catch (err) {
    logger.error('failed to fetch comments for reopened issue %s: %s', number, err)
  }
}
```

backend/lib/watches/journals.js
```diff
@@ -66,5 +69,24 @@ module.exports = (io, retryOptions = {}) => {
     }
   }
 
-  return loadAllOpenIssues()
+  let pollIntervalSeconds = parseInt(config.gitHub.pollIntervalSeconds)
+  if (isNaN(pollIntervalSeconds)) {
+    pollIntervalSeconds = undefined
+  }
+  return pForever(async () => {
```
holgerkoser: A simple do while loop is better:
```js
do {
  await loadAllOpenIssues()
  ...
} while(pollIntervalSeconds)
```

backend/lib/watches/journals.js
petersutter: ✅

backend/lib/watches/journals.js
```diff
@@ -22,7 +22,7 @@ const journals = require('../services/journals')
 const cache = require('../cache')
 const config = require('../config')
 
-module.exports = (io, retryOptions = {}) => {
+module.exports = async (io, retryOptions = {}) => {
```
holgerkoser:
```suggestion
module.exports = (io, retryOptions = {}) => {
```

petersutter:
cc @AndreasBurger @danielfoehrKn @ThormaehlenFred this was released with version 1.39.0

PR #611, opened 2020-02-27
Adapt kyma description

petersutter:
Min supported kubernetes version for kyma is now 1.16.x

Fixes #

Only merge once https://github.com/gardener/gardener/pull/1982 is merged and included in a gardener release

**Release note**:

```improvement user

```


petersutter:
cc @a-thaler

a-thaler:
Change looks good, but is dependent on https://github.com/gardener/gardener/pull/1982 as the related validation rule for the k8s version needs to be adjusted still

grolu:
@petersutter Didn't we agree to move the detailed description (with specific kyma version) to the config and only have a generic message in the code?

petersutter:
@grolu It's already configurable. 
Yes this is what we agreed but on second thought currently also the validation of the kubernetes version is hardcoded in gardener so I think it still makes sense to have this also in the dashboard.

PR #614, opened 2020-02-27
Show OpenStack personalized credential warning as hint

### LINKED ISSUE #595: Make openstack warning for personalized credentials more visible
In secret dialog for OpenStack the warning `Do not use personalized login credentials. Instead, use credentials of a technical user.` is not visible enough, see the screenshot. 

Please put it on the top of the dialog and use some more-special formatting so that it will be the first thing that a user will read.

image

grolu:
Show OpenStack personalized credential warning as hint, currently not persistent
Improved Colorize Component to not overwrite error color classes
Renamed `SelectColorizer` to `Colorizer` Component as it is a universal component which can be used to colorize different kind of input elements

Fixes #595

Screen Shot 2020-02-27 at 21 43 57

**Release note**:

```improvement user
NONE
```


grolu:
cc @vpnachev

frontend/src/components/HintColorizer.vue
```diff
@@ -21,13 +21,21 @@ limitations under the License.
...
-  name: 'select-hint-colorizer',
+  name: 'hint-colorizer',
   props: {
     hintColor: {
       type: String
     }
   },
+  computed: {
+    isSelectErrorColor () {
+      const color = get(this, '$children[0].$children[0].color')
```
petersutter: do we have a test for this?

petersutter: just saw that a test exists 👍

frontend/tests/unit/HintColorizer.spec.js
```diff
@@ -39,4 +39,19 @@ describe('SelectHintColorizer.vue', function () {
...
     expect(colorizerComponent.$el.className).to.not.contain('hintColor-cyan')
   })
+
+  it('should not overwrite error color class', async function () {
+    const data = () => {
+      return {
+        errorMessage: undefined
+      }
+    }
+    const template = '<hint-colorizer hintColor="orange" ref="hintColorizer"><v-text-field :error-messages="errorMessage"></v-text-field></hint-colorizer>'
```
petersutter: add test for v-select

vpnachev:
Does the password have similar warning ?

grolu:
> Does the password have similar warning ?
@vpnachev No, which case do you want to catch with that? If the user uses a technical account, then the password is (hopefully) different from the one he uses for his personal account anyway

vpnachev:
You are right, but even if the user types his personal ID, maybe he will re-consider doing it once he sees one more warning for the password field.

Also, could you increase the font and make it red ?

grolu:
I added a hint for the password field, too. We don't want to show the hint with red color, as this usually indicates an error for the input field. The user would probably not understand this. We already show the hint in warning color which should be sufficient. Font size and weight is the same form all hints on the UI and it would be a b ad idea from ux perspective to change it only for this use case.
If we continue to have issues with this after we released the changes, we can have another discussion.

PR #619, opened 2020-03-05
Require costObject on project

petersutter:
If `costObject` is configured in `values.yaml`, the project (in which the infrastructure secret resides) must have a cost object set in order to be able to create a cluster.

Screenshot 2020-03-05 at 17 53 41

Fixes #

**Release note**:

```noteworthy user
If the `costObject` feature is enabled by the gardener administrator, projects must have a cost object set in order to be able to create a cluster
```

### DIFF
--- a/backend/lib/services/infrastructureSecrets.js
+++ b/backend/lib/services/infrastructureSecrets.js
@@ -16,7 +16,7 @@
 
 const _ = require('lodash')
 const logger = require('../logger')
-const { Resources } = require('../kubernetes-client')
+const { dashboardClient, Resources, isHttpError } = require('../kubernetes-client')
 const { UnprocessableEntity, PreconditionFailed, MethodNotAllowed } = require('../errors')
 const { format: fmt } = require('util')
 const { decodeBase64, encodeBase64 } = require('../utils')
@@ -25,8 +25,9 @@ const cloudprofiles = require('./cloudprofiles')
 const shoots = require('./shoots')
 const { getQuotas } = require('../cache')
 
-function fromResource ({ secretBinding, cloudProviderKind, secret, quotas = [] }) {
+function fromResource ({ secretBinding, cloudProviderKind, secret, project, quotas = [] }) {
   const cloudProfileName = secretBinding.metadata.labels['cloudprofile.garden.sapcloud.io/name']
+  const costObject = _.get(project, 'metadata.annotations["billing.gardener.cloud/costObject"]')
 
   const infrastructureSecret = {}
   infrastructureSecret.metadata = _
@@ -36,7 +37,9 @@ function fromResource ({ secretBinding, cloudProviderKind, secret, quotas = [] }
       cloudProviderKind,
       cloudProfileName,
       bindingNamespace: _.get(secretBinding, 'metadata.namespace'),
-      bindingName: _.get(secretBinding, 'metadata.name')
+      bindingName: _.get(secretBinding, 'metadata.name'),
+      hasCostObject: !_.isEmpty(costObject),
+      projectName: _.get(project, 'metadata.name')
     })
     .value()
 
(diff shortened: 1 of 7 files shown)


frontend/src/components/NewShoot/NewShootInfrastructureDetails.vue
```diff
@@ -134,7 +134,8 @@ import CloudProfile from '@/components/CloudProfile'
 import SecretDialogWrapper from '@/dialogs/SecretDialogWrapper'
 import { required, requiredIf } from 'vuelidate/lib/validators'
 import { getValidationErrors, isOwnSecretBinding, selfTerminationDaysForSecret } from '@/utils'
-import { includesIfAvailable } from '@/utils/validators'
+import { getCostObjectSettings, projectFromProjectList } from '@/utils/projects'
```
holgerkoser: utils should not depend on store (not relevant for this PR, just a remark)

frontend/src/components/NewShoot/NewShootInfrastructureDetails.vue
```diff
@@ -261,6 +244,43 @@ export default {
...
+    validationErrors () {
+      const validationErrors = {
+        secret: {
+          required: 'Secret is required',
+          requiresCostObjectIfEnabled: () => {
+            const projectName = get(this.secret, 'metadata.projectName')
+            const project = projectFromProjectList()
+            const isSecretInProject = project.metadata.name === projectName
+
+            return isSecretInProject ? `${this.costObjectTitle} is required. Go to the ADMINISTRATION page to edit the project and set the ${this.costObjectTitle}.` : `${this.costObjectTitle} is required and has to be set on the Project ${toUpper(projectName)}`
```
grolu: In case of own project, we could also make the `ADMINISTARTION page`a link to the project admin page and open the project settings dialog. But this is only optional.

frontend/src/components/NewShoot/NewShootInfrastructureDetails.vue
petersutter: yes, I will fix this in an upcoming pr (evaluate user permissions on UI)

frontend/src/components/NewShoot/NewShootInfrastructureDetails.vue
petersutter: we can't have html in the error message, right? or do you mean we should display a link somewhere else?

grolu: I don't know. maybe? Let's leave it as is for now

petersutter: 👍

backend/lib/services/infrastructureSecrets.js
```diff
@@ -177,10 +186,35 @@ exports.list = async function ({ user, namespace }) {
       client.core.secrets.list(namespace),
       client['core.gardener.cloud'].secretbindings.list(namespace)
     ])
+
```
holgerkoser:
```   
const defaultNamespace = 'default'
    const namespaces = _
      .chain(secretBindings.items)
      .map(({
        secretRef: {
          namespace = defaultNamespace
        }
      }) => namespace)
      .uniq()
      .value()


async function getProjectByNamespace (namespace) {
  let object
  try {
    object = await dashboardClient.getProjectByNamespace(namespace)
  } catch (err) {
    if (!isHttpError(err, 404)) {
      throw err
    }
  }
  return object
}

async function getProjectMap (namespaces) {
  const list = await _
    .chain(namespaces)
    .uniq()
    .compact()
    .map(getProjectByNamespace)
    .thru(list => Promise.all(list))
    .value()
  return _
    .chain(list)
    .compact()
    .keyBy('spec.namespace')
    .value()
}
```

PR #621, opened 2020-03-06
Improve Project Selection

### LINKED ISSUE #519: Optimize Project List
The Project List in the left navigation bar shows rendering performance issues when a user (e.g. an operator) has access to a huge amount of projects. To increase rendering performance we should use a virtual scroller, so that only visible cells are rendered.
In addition, it makes sense for operators to show only projects that the operator is actually member of (+ the "garden" project) and relabel the filter input field to "Search projects", to search all projects.

### LINKED ISSUE #547: Improve projects searching
Steps to reproduce:

1. Navigate to the dashboard (for example the initial page for operators - `$HOST/namespace/garden/shoots/`)
1. Search for a project named `it` for example.
1. Ensure that multiple project (and project owner) names containing the substring `it` are shown before the project named exactly `it` is shown.

I guess this could be improved as in this case you need to scroll around 50-60 projects to find the project which name you already provided in the search bar. I expect when there is exact match, this exact match to be displayed as the very first suggestion.

grolu:
This PR solves a couple of issues with the MainNavigation Project List selection, especially performance issues and ranking of exact matches.

Screen Shot 2020-03-06 at 17 49 21

Fixes #547 Fixes #519 

**Release note**:

```improvement user
Fixed: Project list navigation via arrow keys in main navigation
```
```improvement user
When searching for projects in main navigation project list, exact matches are now highlighted, ranked at the top and selected by default
```
```improvement user
Project list rendering in main navigation is now lazy. This significantly improves performance when searching or scrolling through large project lists
```


frontend/src/components/MainNavigation.vue
```diff
@@ -255,22 +287,28 @@ export default {
       return (route) => {
         return namespacedRoute(route, this.namespace)
       }
+    },
+    projectFilterHasExactMatch () {
+      return this.isProjectNameMatchingFilter(get(head(this.sortedAndFilteredProjectList), 'metadata.name'))
```
petersutter: extract variable

frontend/src/components/MainNavigation.vue
```diff
@@ -292,6 +330,57 @@ export default {
...
+      const el = get(this.$refs.projectList.$children[projectIndexInVisibleList], '$el')
+
+      if (el) {
+        el.scrollIntoView(false)
+      }
+    },
+    handleProjectListScroll (event) {
+      const parentEl = this.$refs.projectList.$el
+      const parentYPos = parentEl.getBoundingClientRect().top + parentEl.getBoundingClientRect().height
+      const lastProjectEl = get(last(this.$refs.projectList.$children), '$el')
```
petersutter: lastProjectEl could be undefined, right? -> it would brake in the next line

grolu: yes- could be. If the project list is empty. Let's review this method in detail

frontend/src/components/MainNavigation.vue
```diff
@@ -292,6 +331,59 @@ export default {
...
+      if (lastProjectEl) {
```
petersutter: early exit

frontend/src/components/MainNavigation.vue
```diff
@@ -292,6 +331,59 @@ export default {
...
+        const lastProjectElYPos = parentYPos - lastProjectEl.getBoundingClientRect().top
+        if (lastProjectElYPos > 0) {
```
petersutter: const scrolledToLastElement = lastProjectElYPos > 0

frontend/src/components/MainNavigation.vue
```diff
@@ -289,26 +289,30 @@ export default {
...
     ]),
     onEnterProjectFilter () {
-      this.projectMenu = false
-      this.project = this.selectedProject
+      this.navigateToProject(this.highlightedProject)
     },
     onProjectClick (event, project) {
-      this.projectMenu = false
       if (event.isTrusted) {
+        // skip untrusted events - e.g. events triggered via enter key
```
petersutter: ... as they are handled via `onEnterProjectFilter`, right? 
if yes, the comment should mention that

grolu: yes, we listen for the enter key event and navigate to the `highlighted` project. The events we skip here are caused by default browser behavior - the project list items are rendered as links and can be selected using the arrow keys, pressing enter will trigger a mouse-like event (untrusted) for the `focused` link. This is what we skip here as we handle enter events as you mentioned. Moreover, the default selection behavior is behaving incorrectly for our use case

petersutter:
close project menu when selecting the project that is already selected
project enter selection does not work if filter is not there

grolu:
> close project menu when selecting the project that is already selected
> project enter selection does not work if filter is not there

done

frontend/src/components/MainNavigation.vue
```diff
@@ -36,84 +36,99 @@ limitations under the License.
...
-      </v-menu>
+                <v-list-tile-avatar>
+                  <v-icon v-if="project.metadata.name === projectName" color="teal">check</v-icon>
+                </v-list-tile-avatar>
+                <v-list-tile-content>
+                  <v-list-tile-title>{{project.metadata.name}}</v-list-tile-title>
+                  <v-list-tile-sub-title class="project-owner">{{getProjectOwner(project)}}</v-list-tile-sub-title>
+                </v-list-tile-content>
+                <v-list-tile-action v-if="isProjectNameMatchingFilter(project.metadata.name)">
+                  <v-chip small color="cyan darken-2">Exact match</v-chip>
```
holgerkoser: Would not add the exact match chip.

frontend/src/components/MainNavigation.vue
```diff
@@ -255,22 +288,33 @@ export default {
       return (route) => {
         return namespacedRoute(route, this.namespace)
       }
+    },
+    projectFilterHasExactMatch () {
+      const projectName = get(head(this.sortedAndFilteredProjectList), 'metadata.name')
```
holgerkoser: do not inline lodash functions if possible since we don't have chain.
```js
const project = head(this.sortedAndFilteredProjectList)
const projectName = get(project, 'metadata.name')
```

frontend/src/components/MainNavigation.vue
```diff
@@ -429,6 +496,40 @@ export default {
...
+          max-width: 300px;
+
+          div > a {
+            height: 54px;
+          }
+          .project-owner {
+            font-size: 11px;
+          }
+
+          .v-list__tile--highlighted {
```
holgerkoser: fragile

frontend/src/components/MainNavigation.vue
```diff
@@ -429,6 +496,40 @@ export default {
         }
       }
     }
+
+    .project-menu {
+      border-radius: 0;
+
+      >>>.v-card {
```
holgerkoser: fragile

frontend/src/components/MainNavigation.vue
grolu: okay, we can remove it. I have no strong opinion about it. @petersutter WDYT?

frontend/src/components/MainNavigation.vue
grolu: Yes, this is why I added a test that checks for this class.

holgerkoser:
You have to explain the changes a bit more.

frontend/src/components/MainNavigation.vue
grolu: True, but actually old code that I moved from non-scoped css to scoped. I added tests that analyze the class hierarchy and (hopefully) detect incompatible API changes. Of course, this will probably break with vuetify 2 anyway but then we will have to deal with it and adapt accordingly.

frontend/src/components/MainNavigation.vue
grolu: done

frontend/src/components/MainNavigation.vue
petersutter: I thought it was a nice feature but I'm not attached to it so we can remove it

frontend/src/components/MainNavigation.vue
petersutter: - add checks for undefined / not set
- el -> element

frontend/src/components/MainNavigation.vue
```diff
@@ -36,84 +36,99 @@ limitations under the License.
...
+              </v-card-title>
+              <v-divider></v-divider>
+            </template>
+            <v-list light class="project-list" ref="projectList" @scroll.native="handleProjectListScroll">
+              <v-list-tile
+                class="project-list-tile"
+                v-for="(project, index) in visibleProjectList"
+                @click="onProjectClick($event, project)"
+                :class="{'grey lighten-4' : index === highlightedProjectIndex}"
+                :key="index"
```
petersutter: get rid of index if possible

frontend/src/components/MainNavigation.vue
```diff
@@ -240,12 +260,25 @@ export default {
...
       }
-      const sortedList = sortBy(filter(this.projectList, predicate))
-      if (sortedList.length > 1) {
-        sortedList.unshift(this.allProjectsItem)
+      const filteredList = filter(this.projectList, predicate)
+
+      const exactMatch = item => {
+        return this.isProjectNameMatchingFilter(item.metadata.name) ? 0 : 1
       }
+      const sortedList = sortBy(filteredList, exactMatch, 'metadata.name')
```
petersutter: array for sortBy

frontend/src/components/MainNavigation.vue
```diff
@@ -292,6 +337,61 @@ export default {
...
+
+      if (this.highlightedProjectIndex >= this.numberOfVisibleProjects - 1) {
+        this.numberOfVisibleProjects++
+      }
+
+      this.scrollSelectedProjectIntoView()
+    },
+    scrollSelectedProjectIntoView () {
+      const projectIndexInVisibleList = findIndex(this.visibleProjectList, this.highlightedProject)
+      const el = get(this.$refs.projectList.$children[projectIndexInVisibleList], '$el')
```
petersutter: find child by name / id

frontend/tests/unit/MainNavigation.spec.js
```diff
@@ -0,0 +1,196 @@
...
+import noop from 'lodash/noop'
+
+Vue.use(Vuetify)
+Vue.use(Vuex)
+
+window.HTMLElement.prototype.scrollIntoView = noop
+
+let storeProjectList = []
+
+let $store = new Vuex.Store({
```
petersutter: const?

frontend/tests/unit/MainNavigation.spec.js
```diff
@@ -0,0 +1,196 @@
...
+import Vuex from 'vuex'
+import Vuetify from 'vuetify'
+import noop from 'lodash/noop'
+
+Vue.use(Vuetify)
+Vue.use(Vuex)
+
+window.HTMLElement.prototype.scrollIntoView = noop
+
+let storeProjectList = []
```
petersutter: const

frontend/tests/unit/MainNavigation.spec.js
```diff
@@ -0,0 +1,196 @@
...
+    },
+    data: {
+      owner: 'owner'
+    }
+  }
+}
+
+describe('MainNavigation.vue', function () {
+  beforeEach(function () {
+    storeProjectList.length = 0
```
petersutter: comment that this clears the array

frontend/tests/unit/MainNavigation.spec.js
```diff
@@ -0,0 +1,196 @@
...
+    expect(projectMenuCard.element).not.to.be.undefined
+
+    const projectMenuHighlightedTile = wrapper.find('aside .project-menu .v-card .project-list .v-list__tile--highlighted')
+    expect(projectMenuHighlightedTile.element).to.be.undefined
+
+    // highlight (default v-list-tile highlighting)
+    const projectList = wrapper.find('aside .project-menu .v-card .project-list')
+
+    projectList.trigger('keydown.down')
+    expect(projectList.element).not.to.be.undefined
```
petersutter: check

frontend/tests/unit/MainNavigation.spec.js
```diff
@@ -0,0 +1,196 @@
...
+    projectMenuButton.trigger('keydown.up')
+    expect(wrapper.vm.highlightedProjectIndex).to.equal(1)
+
+    const projectFilterInput = wrapper.find('input')
+    projectFilterInput.trigger('keydown.down')
+    projectFilterInput.trigger('keydown.down')
+    expect(wrapper.vm.highlightedProjectIndex).to.equal(3)
+    projectFilterInput.trigger('keydown.up')
+    expect(wrapper.vm.highlightedProjectIndex).to.equal(2)
+  })
```
petersutter: is there a test for lazy rendering?

frontend/tests/unit/MainNavigation.spec.js
```diff
@@ -0,0 +1,196 @@
...
+    expect(exactMatchEl.className).to.contain('grey lighten-4')
+    const exactMatchChip = projectListWrapper.find('.v-chip__content')
+    expect(exactMatchChip.element).not.to.be.undefined
+    expect(exactMatchChip.text()).to.equal('Exact match')
+  })
+
+  it('Projects can be highlighted via arrow keys', function () {
+    storeProjectList.push(createProjectListItem('fooz'))
+    storeProjectList.push(createProjectListItem('foobar'))
+    let wrapper = createMainNavigationComponent()
```
petersutter: const

frontend/tests/unit/MainNavigation.spec.js
grolu: done

frontend/src/components/MainNavigation.vue
grolu: done

frontend/src/components/MainNavigation.vue
grolu: removed

frontend/src/components/MainNavigation.vue
grolu: done

frontend/src/components/MainNavigation.vue
grolu: done

frontend/src/components/MainNavigation.vue
```diff
@@ -292,51 +349,93 @@ export default {
...
-        justify-content: left;
+      this.scrollSelectedProjectIntoView()
+    },
+    scrollSelectedProjectIntoView () {
+      const projectListChildren = get(this, '$refs.projectList.$children')
+      if (!projectListChildren) {
+        return
+      }
+      const projectListItem = find(projectListChildren, child => {
+        return get(child, '$vnode.data.key') === this.highlightedProject.metadata.name
```
grolu: is this save? we access internal properties, but I've added a test for this method. We could also introduce an `id` attribute with identical value. WDYT?

frontend/src/components/MainNavigation.vue
```diff
@@ -92,10 +92,10 @@ limitations under the License.
             <v-list light class="project-list" ref="projectList" @scroll.native="handleProjectListScroll">
               <v-list-tile
                 class="project-list-tile"
-                v-for="(project, index) in visibleProjectList"
+                v-for="project in visibleProjectList"
                 @click="onProjectClick($event, project)"
-                :class="{'grey lighten-4' : index === highlightedProjectIndex}"
-                :key="index"
+                :class="{'grey lighten-4' : project === highlightedProject}"
```
holgerkoser:
```suggestion
                :class="{'grey lighten-4' : isHighlightedProject(project.metadata.name)}"
```

holgerkoser:
```js
computed: {
  ...
  isHighlightedProject(project) {
    return project.metadata.name === this.highlightedProjectName
  },
  ...
}
```

frontend/src/components/MainNavigation.vue
```diff
@@ -339,50 +351,69 @@ export default {
...
-      } else if (this.sortedAndFilteredProjectList.length === 1) {
-        const project = head(this.sortedAndFilteredProjectList)
-        this.highlightedProjectIndex = findIndex(this.sortedAndFilteredProjectListWithAllProjects, project)
+        this.highlightedProjectName = this.projectFilter
       }
-      this.scrollSelectedProjectIntoView()
+      this.$nextTick(() => this.scrollSelectedProjectIntoView())
     },
     selectProjectWithKeys (keyDirection) {
+      let currentHighlightedIndex = this.findProjectIndexCaseInsensitive(this.highlightedProject.metadata.name)
```
holgerkoser:
```suggestion
      let currentHighlightedIndex = this.findProjectIndexCaseInsensitive(this.highlightedProjectName)
```

frontend/src/components/MainNavigation.vue
```diff
@@ -339,50 +351,69 @@ export default {
...
     },
     scrollSelectedProjectIntoView () {
-      const projectIndexInVisibleList = findIndex(this.visibleProjectList, this.highlightedProject)
-      const el = get(this.$refs.projectList.$children[projectIndexInVisibleList], '$el')
+      const projectListChildren = get(this, '$refs.projectList.$children')
+      if (!projectListChildren) {
+        return
+      }
+      const projectListItem = find(projectListChildren, child => {
+        return get(child, '$vnode.data.key') === this.highlightedProject.metadata.name
```
holgerkoser:
```suggestion
        return get(child, '$vnode.data.key') === this.highlightedProjectName
```

frontend/src/components/MainNavigation.vue
```diff
@@ -295,13 +294,26 @@ export default {
       return this.isProjectNameMatchingFilter(projectName)
     },
     highlightedProject () {
-      return this.sortedAndFilteredProjectListWithAllProjects[this.highlightedProjectIndex]
+      if (!this.highlightedProjectName) {
```
holgerkoser: if we only need it for navigateToProject it should be a method not a computed prop

PR #626, opened 2020-03-11
Evaluate HibernationPossible constraint

### LINKED ISSUE #612: Evaluate HibernationPossible constraint
Disable the hibernation button in case the constraint `HibernationPossible` has status `False`.

In addition it makes also sense to disable the reason / message why it can't be hibernated (.status.constraints[].type.HibernationPossible.message)

petersutter:
This PR disables the hibernation button when the hibernation is not possible.
It also gives a warning on the hibernation schedule dialog informing the user in case the schedule can't be carried out as planned.

Screenshot 2020-03-11 at 12 46 08

The tooltip on the disabled button shows the message of the `HibernationPossible` constraint:
Screenshot 2020-03-11 at 12 46 20

Fixes #612 

**Release note**:

```improvement user
The hibernation button will be disabled if `gardener` detects that the cluster cannot be hibernated because of problematic admission webhook configurations on the shoot cluster
```

### DIFF
--- a/frontend/src/components/ShootHibernation/ManageHibernationSchedule.vue
+++ b/frontend/src/components/ShootHibernation/ManageHibernationSchedule.vue
@@ -68,10 +68,12 @@ limitations under the License.
       </v-alert>
     </v-layout>
     <v-layout row v-if="!isHibernationPossible" class="pt-2">
-      <v-alert :value="true" type="warning" outline>
-        <div class="font-weight-bold">Your hibernation schedule may not have any effect:</div>
-        <div>{{hibernationPossibleMessage}}</div>
-      </v-alert>
+      <v-flex>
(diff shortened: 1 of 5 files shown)


frontend/src/components/ShootHibernation/ManageHibernationSchedule.vue
```diff
@@ -67,6 +67,12 @@ limitations under the License.
         You probably configured crontab lines for your hibernation schedule manually. Please edit your schedules directly in the cluster specification. You can also delete it there and come back to this screen to configure your schedule via the Dashboard UI.
       </v-alert>
     </v-layout>
+    <v-layout row v-if="!isHibernationPossible" class="pt-2">
+      <v-alert :value="true" type="warning" outline>
```
holgerkoser: remove `:value="true"`

PR #627, opened 2020-03-11
Fix: fetch kyma info

petersutter:
Currently the kyma info is not properly loaded. With this PR the kyma info is fetched when the details page is shown / on reload.

Fixes #

**Release note**:

```improvement user

```

### DIFF
--- a/frontend/src/pages/ShootDetails.vue
+++ b/frontend/src/pages/ShootDetails.vue
@@ -148,9 +148,10 @@ export default {
       'getShootAddonKyma'
     ]),
     fetchKymaInfo () {
-      if (!this.isKymaFeatureEnabled && this.isKymaAddonEnabled) {
+      if (!this.isKymaFeatureEnabled || !this.isKymaAddonEnabled) {
         return
       }
+
       this.getShootAddonKyma({ name: this.shootName, namespace: this.shootNamespace })
     }
   },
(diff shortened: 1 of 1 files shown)


frontend/src/pages/ShootDetails.vue
```diff
@@ -146,20 +146,24 @@ export default {
   methods: {
     ...mapActions([
       'getShootAddonKyma'
-    ])
+    ]),
+    fetchKymaInfo () {
+      if (!this.isKymaFeatureEnabled && this.isKymaAddonEnabled) {
```
holgerkoser: KymaAddon is disabled but we fetch the KymaInfo?

frontend/src/pages/ShootDetails.vue
```diff
@@ -146,20 +146,24 @@ export default {
...
       this.$refs.shootLifecycle.showHibernationConfigurationDialog()
     }
+    this.fetchKymaInfo()
   },
   watch: {
     isKymaAddonEnabled (currentlyEnabled, previouslyEnabled) {
-      if (!this.isKymaFeatureEnabled) {
-        return
-      }
       if (!previouslyEnabled && currentlyEnabled) {
```
holgerkoser: why check the previous value? If current value is true and it has changed we should fetch the infi.

frontend/src/pages/ShootDetails.vue
petersutter: right, will fix it

frontend/src/pages/ShootDetails.vue
petersutter: ✅

PR #628, opened 2020-03-11
Check whether gardenlet respects the ignore annotation

### LINKED ISSUE #386: Check whether GCM respect the ignore annotation
The gardener controller manager could be configured[1] to not respect the annotation `shoot.garden.sapcloud.io/ignore`, which means that it will reconcile the shoot cluster even if the annotation is set properly. The dashboard itself is just checking the annotation to determine whether the reconciliation is enabled or not. 

[1] https://github.com/gardener/gardener/blob/2f79201d69dc27ea738d70785a0aa5223bf3481a/pkg/controllermanager/apis/config/v1alpha1/types.go#L175-L178

petersutter:
This PR adds the configuration flag if gardenlet respects the ignore annotation. This flag should be in sync with `GardenletConfiguration` `controllers.shoot.respectSyncPeriodOverwrite` https://github.com/gardener/gardener/blob/045b03f02d7bf47a9ef39e2981ce92b86e444bda/pkg/gardenlet/apis/config/v1alpha1/types.go#L223-L226

example `values.yaml`:
```yaml
frontendConfig:
  ...
  gardenlet:
    respectSyncPeriodOverwrite: true
```

Fixes #386

**Release note**:

```action operator
In case the flag `controllers.shoot.respectSyncPeriodOverwrite` is disabled on the `GardenletConfiguration`, set the dashboard flag `frontendConfig.gardenlet.respectSyncPeriodOverwrite` also to false
```

### DIFF
--- a/charts/gardener-dashboard/values.yaml
+++ b/charts/gardener-dashboard/values.yaml
@@ -69,6 +69,7 @@ frontendConfig:
       end: 00 08 * * 1,2,3,4,5
     production: ~
   seedCandidateDeterminationStrategy: SameRegion
+  gardenletRespectsSyncPeriodOverwrite: true # this flag should be in sync with gardenlet-configuration controllers.shoot.respectSyncPeriodOverwrite https://github.com/gardener/gardener/blob/045b03f02d7bf47a9ef39e2981ce92b86e444bda/pkg/gardenlet/apis/config/v1alpha1/types.go#L223-L226
   features:
     terminalEnabled: false
     kymaEnabled: false
--- a/charts/gardener-dashboard/values.yaml
+++ b/charts/gardener-dashboard/values.yaml
@@ -70,7 +70,7 @@ frontendConfig:
     production: ~
   seedCandidateDeterminationStrategy: SameRegion
   gardenlet:
-    respectsSyncPeriodOverwrite: true # this flag should be in sync with gardenlet-configuration controllers.shoot.respectSyncPeriodOverwrite https://github.com/gardener/gardener/blob/045b03f02d7bf47a9ef39e2981ce92b86e444bda/pkg/gardenlet/apis/config/v1alpha1/types.go#L223-L226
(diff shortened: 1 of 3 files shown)


charts/gardener-dashboard/values.yaml
```diff
@@ -69,6 +69,7 @@ frontendConfig:
       end: 00 08 * * 1,2,3,4,5
     production: ~
   seedCandidateDeterminationStrategy: SameRegion
+  gardenletRespectsSyncPeriodOverwrite: true # this flag should be in sync with gardenlet-configuration controllers.shoot.respectSyncPeriodOverwrite https://github.com/gardener/gardener/blob/045b03f02d7bf47a9ef39e2981ce92b86e444bda/pkg/gardenlet/apis/config/v1alpha1/types.go#L223-L226
```
holgerkoser: Maybe more gardenlet configuration properties must be added. Rename to `gardenlet.respectSyncPeriodOverwrite`. We should use the same property name.

petersutter: ✅

petersutter:
respectsSyncPeriodOverwrite -> respectSyncPeriodOverwrite

petersutter:
cc @rfranzke

rfranzke:
Which gardenlet? There might be hundreds of them that are each configured differently.

petersutter:
Oh, well I picked up the ticket #386 which was from a time where this setting was part of the GCM and I naively adapted it to gardenlet.
As the config can be different for each gardenlet and there is probably no way for us to figure out the setting during runtime I will revert the change

rfranzke:
Right now it's not possible, no. Maybe we should move this configuration out of the gardenlet config and to the seed taints. Then you can fetch it per seed without the necessity of configuring the dashboard itself. WDYT?

petersutter:
yes something like that would be helpful 👍

PR #649, opened 2020-03-27
Terminals: Use non-deprecated fields

petersutter:
The following applies only for `Terminal` resources:

We should not use anymore the annotation `garden.sapcloud.io/createdBy`. This field was deprecated with `terminal-controller-manager` `v0.8.0`. It will be removed with the upcoming version `v0.9.0`. 
Instead, `gardener.cloud/created-by` annotation should be used

In addition:
- using `dashboard.gardener.cloud/created-by-hash` label instead of `garden.sapcloud.io/createdBy`
- using `dashboard.gardener.cloud/identifier-hash` label instead of `dashboard.gardener.cloud/identifier`

Also unused labels are removed:
- `dashboard.gardener.cloud/hostCluster`
- `dashboard.gardener.cloud/targetCluster`
- `garden.sapcloud.io/name`

Fixes #

Corresponding commit on `terminal-controller-manager`: https://github.com/gardener/terminal-controller-manager/commit/ab92b4fb3a05f56b3fcab25d8f627d3cad38c332

**Release note**:

```action operator
This dashboard version requires `terminal-controller-manager` `v0.8.0` or higher, in case the terminal feature is enabled.
```

### DIFF
--- a/backend/lib/services/terminals/index.js
+++ b/backend/lib/services/terminals/index.js
@@ -139,10 +139,10 @@ async function listTerminals ({ user, namespace, identifier }) {
   const client = user.client
 
   const selectors = [
-    `garden.sapcloud.io/createdBy=${hash(username)}`
+    `dashboard.gardener.cloud/created-by-hash=${hash(username)}`
   ]
   if (identifier) {
-    selectors.push(`dashboard.gardener.cloud/identifier=${hash(identifier)}`)
+    selectors.push(`dashboard.gardener.cloud/identifier-hash=${hash(identifier)}`)
   }
   const query = {
     labelSelector: selectors.join(',')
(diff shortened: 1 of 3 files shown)


backend/lib/services/terminals/index.js
```diff
@@ -333,27 +333,23 @@ function getHostCluster ({ user, namespace, name, target, body, shootResource })
   return getShootHostCluster(client, { namespace, name, target, body, shootResource })
 }
 
-async function createTerminal ({ user, namespace, name, target, hostCluster, targetCluster, body }) {
+async function createTerminal ({ user, namespace, target, hostCluster, targetCluster, body }) {
   const client = user.client
   const isAdmin = user.isAdmin
   const { identifier } = body
   const containerImage = getContainerImage({ isAdmin, preferredContainerImage: hostCluster.containerImage })
+  const containerArgs = hostCluster.containerArgs
```
holgerkoser: Use hierarchy for hierarchy:
```js
{ 
  container: {
    preferredImage': 'prefName', 
    image: 'name',
    args: []
  }
}
```

petersutter: this shouldn't be part of this PR, will remove it again. But I will consider it for the upcoming PR

PR #663, opened 2020-04-10
Rename suse-jeos to suse-chost

vpnachev:
Add support for `suse-chost` which is the replacement of `suse-jeos`.

Fixes #

**Release note**:

```improvement operator
A support for `suse-chost` as alternative name of `suse-jeos` has been added.
```

### DIFF
--- a/frontend/src/store/index.js
+++ b/frontend/src/store/index.js
@@ -126,12 +126,14 @@ const vendorNameFromImageName = imageName => {
     return 'ubuntu'
   } else if (lowerCaseName.includes('suse') && lowerCaseName.includes('jeos')) {
     return 'suse-jeos'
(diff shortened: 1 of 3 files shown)


frontend/src/store/index.js
```diff
@@ -126,6 +126,8 @@ const vendorNameFromImageName = imageName => {
     return 'ubuntu'
   } else if (lowerCaseName.includes('suse') && lowerCaseName.includes('jeos')) {
     return 'suse-jeos'
+  } else if (lowerCaseName.includes('suse') && lowerCaseName.includes('chost')) {
+    return 'suse-chost'
   }
   return undefined
 }
```
petersutter, line 133: `vendorNeedsLicense` needs to be adapted to `suse-chost` as well, right?

vpnachev: good catch

PR #668, opened 2020-04-17
[vuetify-2.2] Filter floating pools that start or end with wildcard

grolu:
Filter floating pools that start or end with wildcard

Fixes #

**Release note**:

```improvement user

```


frontend/src/store/index.js
```diff
@@ -386,7 +386,11 @@ const getters = {
     return (cloudProfileName, region) => {
       const cloudProfile = getters.cloudProfileByName(cloudProfileName)
       const floatingPools = get(cloudProfile, 'data.providerConfig.constraints.floatingPools')
-      const availableFloatingPools = filter(floatingPools, matchesPropertyOrEmpty('region', region))
+      let availableFloatingPools = filter(floatingPools, matchesPropertyOrEmpty('region', region))
+      availableFloatingPools = filter(floatingPools, fp => {
+        const wildcard = startsWith(fp.name, '*') || endsWith(fp.name, '*')
```
petersutter, line 391: what about importing startsWith and endsWith?

grolu: good point

PR #674, opened 2020-04-22
Fix table sorting

petersutter:
This PR fixes the table sorting, that broke with the vuetify 2.2 upgrade

Fixes #

**Release note**:

```improvement user

```

### DIFF
--- a/frontend/src/store/modules/shoots.js
+++ b/frontend/src/store/modules/shoots.js
@@ -457,7 +457,7 @@ const shoots = (state) => {
 const setSortedItems = (state, rootState) => {
   const sortBy = head(get(state, 'sortParams.sortBy'))
   const sortDesc = get(state, 'sortParams.sortDesc', [false])
-  const descending = head(sortDesc) ? 'desc' : 'asc'
+  const sortOrder = head(sortDesc) ? 'desc' : 'asc'
   if (sortBy) {
     const sortbyNameAsc = (a, b) => {
       if (getRawVal(a, 'name') > getRawVal(b, 'name')) {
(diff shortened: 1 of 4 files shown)


frontend/src/store/modules/shoots.js
```diff
@@ -455,8 +455,10 @@ const shoots = (state) => {
 }
 
 const setSortedItems = (state, rootState) => {
-  const sortBy = get(state, 'sortParams.sortBy')
-  const descending = get(state, 'sortParams.descending', false) ? 'desc' : 'asc'
+  const sortBy = head(get(state, 'sortParams.sortBy'))
+  const sortDesc = get(state, 'sortParams.sortDesc', [false])
+  const descending = head(sortDesc) ? 'desc' : 'asc'
+  console.log(sortBy, descending)
```
holgerkoser, line 461: remove `console.log`

petersutter: ✅

frontend/src/store/modules/shoots.js
```diff
@@ -455,8 +455,9 @@ const shoots = (state) => {
 }
 
 const setSortedItems = (state, rootState) => {
-  const sortBy = get(state, 'sortParams.sortBy')
-  const descending = get(state, 'sortParams.descending', false) ? 'desc' : 'asc'
+  const sortBy = head(get(state, 'sortParams.sortBy'))
+  const sortDesc = get(state, 'sortParams.sortDesc', [false])
+  const descending = head(sortDesc) ? 'desc' : 'asc'
```
holgerkoser, line 460: I know it has nothing to do with this PR. But it is completely confusion that a variable with the name `descending` could have a value `'asc'`? We should rename `descending` into `order` or `sortOrder`.

petersutter: ✅

PR #675, opened 2020-04-23
Condense provider, region and zone

petersutter:
The Infrastructure card takes up a lot of space. As an effort to reduce unnecessary rows, the provider, region and zone are now combined in one row.
Screenshot 2020-04-23 at 11 14 30

Fixes #

**Release note**:

```improvement user

```


frontend/src/components/Vendor.vue
```diff
@@ -0,0 +1,83 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+
+<template>
+  <v-tooltip top>
```
holgerkoser, line 18: What is the value of the tooltip. It shows the same information in a different way.

frontend/src/components/ShootDetails/ShootInfrastructureCard.vue
```diff
@@ -25,9 +25,9 @@ limitations under the License.
           <v-icon color="cyan darken-2">cloud_queue</v-icon>
         </v-list-item-icon>
         <v-list-item-content>
-          <v-list-item-subtitle>Provider</v-list-item-subtitle>
+          <v-list-item-subtitle>{{vendorTitle}}</v-list-item-subtitle>
           <v-list-item-title class="pt-1">
-            {{shootCloudProviderKind}}
+            <vendor :shootItem="shootItem" :extended="true"></vendor>
```
holgerkoser, line 30: The vendorTitle and the vendor a coupled. Not things are separated into different components. I would keep one thing in one component.

petersutter: hm, the vender component is also used in the shoot list row, so I can't put
v-list-item-subtitle and v-list-item-title into the vendor component, or are you thinking of a different solution?

frontend/src/components/Vendor.vue
petersutter: yes on the details page it does not give any benefit, however on the shoot list row it does add some value
we could disable the tooltip on the shoot details page. WDYT?

frontend/src/components/ShootDetails/ShootInfrastructureCard.vue
holgerkoser: I was not aware of the fact that the vendor component is used in the ShootListRow. We could add a boolean prop `title` and if true it renders the title
```js
      if (this.shootZones.length > 0) {  
         return `Provider / Region / ${this.zoneTitle}`
      }
      return 'Provider / Region'
```

frontend/src/components/ShootDetails/ShootInfrastructureCard.vue
```diff
@@ -25,9 +25,9 @@ limitations under the License.
           <v-icon color="cyan darken-2">cloud_queue</v-icon>
         </v-list-item-icon>
         <v-list-item-content>
-          <v-list-item-subtitle>Provider</v-list-item-subtitle>
+          <v-list-item-subtitle>{{vendorTitle}}</v-list-item-subtitle>
```
holgerkoser, line 28:
```suggestion
          <v-list-item-subtitle>
            <vendor title extended :shootItem="shootItem">
          </v-list-item-subtitle>
          <v-list-item-title class="pt-1">
            <vendor extended :shootItem="shootItem">
          </v-list-item-title>
```

frontend/src/components/Vendor.vue
holgerkoser: see above

petersutter: ✅

frontend/src/components/ShootDetails/ShootInfrastructureCard.vue
petersutter: ✅

frontend/src/components/ShootDetails/ShootInfrastructureCard.vue
petersutter: ✅ (did not accept suggestion here as the closing tag was missing, however I changed the code like you suggested)

PR #685, opened 2020-05-07
filter regions without zones for all infrastructures

grolu:

Fixes #

**Release note**:

```improvement user
Fixed: Recommended regions derived from available seeds are now filtered by available zones in selected cloud profile
```

### DIFF
--- a/frontend/src/store/index.js
+++ b/frontend/src/store/index.js
@@ -152,11 +152,8 @@ const matchesPropertyOrEmpty = (path, srcValue) => {
 
 const isValidRegion = (getters, cloudProfileName, cloudProviderKind) => {
   return region => {
-    if (cloudProviderKind === 'azure') {
-      // Azure regions may not be zoned, need to filter these out for the dashboard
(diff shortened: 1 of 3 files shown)


frontend/src/store/index.js
```diff
@@ -156,6 +156,13 @@ const isValidRegion = (getters, cloudProfileName, cloudProviderKind) => {
       // Azure regions may not be zoned, need to filter these out for the dashboard
       return !!getters.zonesByCloudProfileNameAndRegion({ cloudProfileName, region }).length
     }
+
+    // Filter regions that are not defined in cloud profile
+    const cloudProfile = getters.cloudProfileByName(cloudProfileName)
+    if (cloudProfile) {
+      return !!(find(cloudProfile.data.regions, { name: region }))
```
petersutter, line 163: we usually do the following (however I haven't tested it), also there was an unnecessary bracket
```suggestion
      return !!find(cloudProfile.data.regions, ['name', region])
```

PR #687, opened 2020-05-07
Allow longer cluster names if possible

### LINKED ISSUE #650: Shoot name length restriction
We have an enhancement request from a dashboard consumer:

**What would you like to be added**/**Why is this needed**::
We have a lot of shoots in our projects and sometimes 10 characters are just not enough to give them a meaningful name. Would you consider adding some "Display Name"?
Alternatively, it would be cool, if I could select a custom-label in the "Column Selection" of the dashboard.

grolu:

Fixes #650

**Release note**:

```improvement user
Maximum cluster name length is now dynamic, depending on the project name
```


grolu:
> the max project name length is still 10 characters?

Yes, longer project names get rejected by gardener

PR #688, opened 2020-05-07
wildcard floating pools

grolu:
Screen Shot 2020-05-12 at 14 58 02

Fixes #

**Release note**:

```improvement user
Support configuration of wildcard floating pools for OpenStack
```

### DIFF
--- a/frontend/src/components/WildcardSelect.vue
+++ b/frontend/src/components/WildcardSelect.vue
@@ -34,7 +34,7 @@ limitations under the License.
         <v-list-item-content>
           <v-list-item-title>
             <span v-if="item.startsWithWildcard">&lt;prefix&gt;</span>
-            <span>{{item.text}}</span>
+            <span>{{item.value}}</span>
             <span v-if="item.endsWithWildcard">&lt;suffix&gt;</span>
           </v-list-item-title>
         </v-list-item-content>
@@ -112,22 +112,25 @@ export default {
       return map(this.wildcardSelectItems, item => {
         let startsWithWildcard = false
         let endsWithWildcard = false
+        const value = trim(item, '*')
+        let pattern = value
 
-        let selectedValue = item
         if (startsWith(item, '*')) {
           startsWithWildcard = true
+          pattern = '.+' + pattern
         }
         if (endsWith(item, '*')) {
           endsWithWildcard = true
+          pattern = pattern + '.+'
         }
-        selectedValue = trim(item, '*')
 
         return {
-          text: selectedValue,
-          value: selectedValue,
+          text: value,
+          value,
           startsWithWildcard,
           endsWithWildcard,
-          isWildcard: startsWithWildcard || endsWithWildcard
+          isWildcard: startsWithWildcard || endsWithWildcard,
+          regex: new RegExp('^' + pattern + '$')
         }
       })
     },
@@ -166,13 +169,7 @@ export default {
     },
     setInternalValue (newValue) {
       const matches = filter(this.wildcardSelectItemObjects, (item) => {
-        if (item.startsWithWildcard) {
-          return new RegExp('^.+' + item.value + '$').test(newValue)
-        }
-        if (item.endsWithWildcard) {
-          return new RegExp('^' + item.value + '.+$').test(newValue)
-        }
-        return new RegExp('^' + item.value + '$').test(newValue)
+        return item.regex.test(newValue)
       })
       matches.sort(function (a, b) {
         return b.value.length - a.value.length
--- a/frontend/src/components/WildcardSelect.vue
+++ b/frontend/src/components/WildcardSelect.vue
@@ -57,8 +57,9 @@ limitations under the License.
 import startsWith from 'lodash/startsWith'
 import endsWith from 'lodash/endsWith'
 import map from 'lodash/map'
-import forEach from 'lodash/forEach'
 import trim from 'lodash/trim'
+import filter from 'lodash/filter'
+import head from 'lodash/head'
 import { getValidationErrors } from '@/utils'
 import { required, requiredIf } from 'vuelidate/lib/validators'
 
@@ -126,7 +127,7 @@ export default {
           value: selectedValue,
           startsWithWildcard,
           endsWithWildcard,
-          isWildcard: startsWithWildcard | endsWithWildcard
+          isWildcard: startsWithWildcard || endsWithWildcard
         }
       })
     },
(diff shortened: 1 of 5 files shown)


holgerkoser:
It is possible to have chips in v-select and on the chips we could have a menu like here https://vuetifyjs.com/en/components/chips/#expandable.

petersutter:
can you also update the screenshot?

frontend/src/components/WildcardSelect.vue
```diff
@@ -0,0 +1,215 @@
...
+        wildcardVariablePart: {
+          required: `${this.wildcardTextFieldLabel} is required`
+        },
+        wildcardSelectedValue: {
+          required: `${this.wildcardSelectLabel} is required`
+        }
+      }
+    },
+    textFieldClass () {
+      return this.wildcardSelectedValue.startsWithWildcard ? 'textFieldStartClass' : 'textFieldEndClass'
```
petersutter, line 108: just to make sure: a value like `*foo*` is not possible, right (starts and ends with wildcard)?

frontend/src/components/WildcardSelect.vue
```diff
@@ -0,0 +1,215 @@
...
+      const rawValue = trim(value, '*')
+      if (!rawValue) {
+        return
+      }
+      const valueLength = rawValue.length
+      let smallestVariablePartLength = valueLength
+      forEach(this.wildcardSelectItemObjects, itemObject => {
+        if (rawValue.includes(itemObject.value)) {
+          const variablePartLength = valueLength - itemObject.value.length
+          if (variablePartLength < smallestVariablePartLength) {
```
petersutter, line 176: early exits for both if's, so that we do not have such a deep nesting

grolu:
> can you also update the screenshot?

done

frontend/src/components/WildcardSelect.vue
grolu: I don't think so, at least not supported by the dashboard. @timuthy is this supported by gardener?

timuthy: No, also not supported by Gardener.

frontend/src/components/WildcardSelect.vue
```diff
@@ -0,0 +1,221 @@
...
+    },
+    textFieldClass () {
+      return this.wildcardSelectedValue.startsWithWildcard ? 'textFieldStartClass' : 'textFieldEndClass'
+    },
+    wildcardSelectItemObjects () {
+      return map(this.wildcardSelectItems, item => {
+        let startsWithWildcard = false
+        let endsWithWildcard = false
+
+        let selectedValue = item
```
petersutter, line 116: assignment unnecessary

frontend/src/components/WildcardSelect.vue
```diff
@@ -0,0 +1,221 @@
...
+        }
+        selectedValue = trim(item, '*')
+
+        return {
+          text: selectedValue,
+          value: selectedValue,
+          startsWithWildcard,
+          endsWithWildcard,
+          isWildcard: startsWithWildcard || endsWithWildcard
+        }
```
petersutter, line 131: why not building the regexp pattern already here. Something like
```js
        const trimmedValue = trim(item, '*')
        let pattern
        if (startsWith(item, '*')) {
          startsWithWildcard = true
          // TODO pattern = '^.+' + trimmedValue + '$'
        } else if (endsWith(item, '*')) {
          endsWithWildcard = true
          // TODO pattern = ...
        } else {
          // TODO pattern = ...
        }

        return {
          text: trimmedValue,
          value: trimmedValue,
          pattern,
          startsWithWildcard,
          endsWithWildcard,
          isWildcard: startsWithWildcard || endsWithWildcard
        }
```

frontend/src/components/WildcardSelect.vue
```diff
@@ -0,0 +1,221 @@
...
+          startsWithWildcard = true
+        }
+        if (endsWith(item, '*')) {
+          endsWithWildcard = true
+        }
+        selectedValue = trim(item, '*')
+
+        return {
+          text: selectedValue,
+          value: selectedValue,
```
petersutter, line 127: why do we need both text and value with the same values?

frontend/src/components/WildcardSelect.vue
```diff
@@ -0,0 +1,221 @@
...
+    },
+    setInternalValue (newValue) {
+      const matches = filter(this.wildcardSelectItemObjects, (item) => {
+        if (item.startsWithWildcard) {
+          return new RegExp('^.+' + item.value + '$').test(newValue)
+        }
+        if (item.endsWithWildcard) {
+          return new RegExp('^' + item.value + '.+$').test(newValue)
+        }
+        return new RegExp('^' + item.value + '$').test(newValue)
```
petersutter, line 175: when introducing the pattern as mentioned above, this could be reduced to
```js
return new RegExp('^' + item.pattern + '$').test(newValue)
```

frontend/src/components/WildcardSelect.vue
```diff
@@ -0,0 +1,221 @@
...
+      })
+
+      const bestMatch = head(matches)
+      if (!bestMatch) {
+        return
+      }
+
+      this.wildcardSelectedValue = bestMatch
+      if (bestMatch.isWildcard) {
+        const value = trim(newValue, '*')
```
petersutter, line 188: do we still need to trim here? please double check

frontend/src/components/WildcardSelect.vue
grolu: leftover from before we had custom slots. v-select expects text / value in case of item objects. will remove it, we can re-use value attribute in the slot

frontend/src/components/WildcardSelect.vue
grolu: yes, I actually thought about this and had in that way before. did some refactoring and ended up not having the regex as part of the item object. I have no real opinion, if you like it better that way, I can change it.

petersutter: If you trim the value in one method and in the other you take this (trimmed) value and build the pattern for the regexp, it makes it harder to understand / follow. So for me it was not clear on first sight that `item.value` was actually the trimmed value.
This makes way makes it more clearer as it is done in one place.

frontend/src/components/WildcardSelect.vue
grolu: ok I eneded up leaving it as is with value and text, as we require text for the select compinent and I don't want to treat `text` as value in my code.

frontend/src/components/WildcardSelect.vue
grolu: Yes, still required, can explain if you want.

petersutter: is newValue the value from the selection? Because the value from the selection does not contain the wildcard, or does it?

PR #696, opened 2020-05-20
Wildcard Select: Support wildcard only values

grolu:
This PR adds support for the case if the value of an item of a wildcard select is `*`only. In this case we want to have a free text input and not a fixed pre or suffix.

Fixes #

**Release note**:

```improvement user

```

### DIFF
--- a/frontend/src/components/WildcardSelect.vue
+++ b/frontend/src/components/WildcardSelect.vue
@@ -18,6 +18,7 @@ limitations under the License.
   <div class="d-flex flex-row">
     <v-select
       class="selectClass"
+      :class="selectClass"
       color="cyan darken-2"
       item-color="cyan darken-2"
       :label="wildcardSelectLabel"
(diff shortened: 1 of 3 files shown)


frontend/src/components/WildcardSelect.vue
```diff
@@ -108,41 +118,63 @@ export default {
...
           startsWithWildcard = true
-          pattern = '.+' + pattern
-        }
-        if (endsWith(item, '*')) {
+          pattern = '.*' + pattern
+        } else if (endsWith(item, '*')) {
           endsWithWildcard = true
-          pattern = pattern + '.+'
+          pattern = pattern + '.*'
         }
```
petersutter, line 144: why `.*` instead of `.+`?

frontend/src/store/index.js
```diff
@@ -400,6 +400,7 @@ const getters = {
       const cloudProfile = getters.cloudProfileByName(cloudProfileName)
       const floatingPools = get(cloudProfile, 'data.providerConfig.constraints.floatingPools')
       const availableFloatingPools = filter(floatingPools, matchesPropertyOrEmpty('region', region))
+      availableFloatingPools.push({ name: '*' })
```
petersutter, line 403: I guess this should be removed, right?

grolu: done

PR #700, opened 2020-05-26
log filtered cloudprofiles

grolu:

Fixes #

**Release note**:

```improvement user

```

### DIFF
--- a/backend/lib/services/cloudprofiles.js
+++ b/backend/lib/services/cloudprofiles.js
@@ -68,7 +68,13 @@ exports.list = function () {
   return _
     .chain(cloudProfiles)
     .map(assignSeedsToCloudProfileIteratee(seeds))
-    .filter('data.seeds')
+    .filter(cloudProfile => {
+      if (!_.isEmpty(cloudProfile.data.seeds)) {
+        return true
+      }
+      console.log(`No matching seed for cloud profile with name ${cloudProfile.metadata.name} found`)
+      return false
+    })
     .value()
 }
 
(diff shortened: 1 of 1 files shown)


backend/lib/services/cloudprofiles.js
```diff
@@ -68,7 +68,13 @@ exports.list = function () {
...
     .chain(cloudProfiles)
     .map(assignSeedsToCloudProfileIteratee(seeds))
-    .filter('data.seeds')
+    .filter(cloudProfile => {
+      if (!_.isEmpty(cloudProfile.data.seeds)) {
+        return true
+      }
+      console.log(`No matching seed for cloud profile with name ${cloudProfile.metadata.name} found`)
+      return false
+    })
```
holgerkoser, line 77: .filter('data.seeds.length')

grolu:
@holgerkoser the point of this PR to (re)add the missing logging for the case where cloud profiles without seeds are removed from the list. This makes it easier to identify why an infrastructure / cloud profile is missing in the frontend.

backend/lib/services/cloudprofiles.js
```diff
@@ -68,7 +68,13 @@ exports.list = function () {
   return _
     .chain(cloudProfiles)
     .map(assignSeedsToCloudProfileIteratee(seeds))
-    .filter('data.seeds')
+    .filter(cloudProfile => {
+      if (!_.isEmpty(cloudProfile.data.seeds)) {
+        return true
+      }
+      console.log(`No matching seed for cloud profile with name ${cloudProfile.metadata.name} found`)
```
petersutter, line 75: use logger

grolu: done

PR #705, opened 2020-05-27
Support Floating Pool domain restrictions

grolu:
This PR adds support to restrict Floating Pools per domain. It also now respects the nonConstraining option. See also https://github.com/gardener/gardener-extension-provider-openstack/pull/76
Furthermore this PR fixes the bug that previously regionalized FPs and LBs did not overwrite the generic ones.

Fixes #

**Release note**:

```improvement user
Added support for optional Floating Pools domain restrictions
```
```improvement user
Fixed: It was possible to select a generic Floating Pool or Load Balancer Provider even if a regionalized exists
```

### DIFF
--- a/frontend/src/store/index.js
+++ b/frontend/src/store/index.js
@@ -402,14 +402,12 @@ const getters = {
       let availableFloatingPools = filter(floatingPools, matchesPropertyOrEmpty('region', region))
       availableFloatingPools = filter(availableFloatingPools, matchesPropertyOrEmpty('domain', secretDomain))
 
-      if (find(availableFloatingPools, fp => {
-        return !!fp.region && !fp.nonConstraining
-      })) {
+      const hasRegionSpecificFloatingPool = find(availableFloatingPools, fp => !!fp.region && !fp.nonConstraining)
+      if (hasRegionSpecificFloatingPool) {
         availableFloatingPools = filter(availableFloatingPools, { region })
       }
-      if (find(availableFloatingPools, fp => {
-        return !!fp.domain && !fp.nonConstraining
-      })) {
+      const hasDomainSpecificFloatingPool = find(availableFloatingPools, fp => !!fp.domain && !fp.nonConstraining)
+      if (hasDomainSpecificFloatingPool) {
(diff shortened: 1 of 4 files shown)


frontend/src/store/index.js
```diff
@@ -395,19 +395,37 @@ const getters = {
...
+
+      if (find(availableFloatingPools, fp => {
+        return !!fp.region && !fp.nonConstraining
+      })) {
+        availableFloatingPools = filter(availableFloatingPools, { region })
+      }
+      if (find(availableFloatingPools, fp => {
+        return !!fp.domain && !fp.nonConstraining
+      })) {
+        availableFloatingPools = filter(availableFloatingPools, { domain: secretDomain })
```
petersutter, line 413: why not call secretDomain just domain?

grolu: I called it that way to make clear where the domain is coming from. I don't know if there are other places where domains can be specified e.g. in the cloud profile.

petersutter: okay let's leave it as is

frontend/src/store/index.js
```diff
@@ -395,19 +395,37 @@ const getters = {
...
+    return ({ cloudProfileName, region }) => {
       const cloudProfile = getters.cloudProfileByName(cloudProfileName)
       const loadBalancerProviders = get(cloudProfile, 'data.providerConfig.constraints.loadBalancerProviders')
-      const availableLoadBalancerProviders = filter(loadBalancerProviders, matchesPropertyOrEmpty('region', region))
+      let availableLoadBalancerProviders = filter(loadBalancerProviders, matchesPropertyOrEmpty('region', region))
+      if (find(availableLoadBalancerProviders, lb => {
+        return !!lb.region
+      })) {
+        availableLoadBalancerProviders = filter(availableLoadBalancerProviders, { region })
+      }
```
petersutter, line 428: can you explain this. Also extract the condition of the if-clause to a variable

```js
find(availableLoadBalancerProviders, lb => !!lb.region)
```
and this should also do the trick, right?

grolu: done

PR #709, opened 2020-05-29
Terminal Bootstrap: Skip seed with missing secretRef

petersutter:
With the introduction of the [gardenlet](https://github.com/gardener/gardener/releases/tag/v0.33.0) the `spec.secretRef` is optional on the seed resource.
If it is missing, the shoot cannot be bootstrapped and needs to be skipped.

In case a shoot is used as seed, the flag `with-secret-ref` has to be added to the `shoot.gardener.cloud/use-as-seed` annotation so that the `spec.secretRef` is set on the seed.

Fixes #

**Release note**:

```improvement user

```

### DIFF
--- a/backend/lib/services/shoots.js
+++ b/backend/lib/services/shoots.js
@@ -330,6 +330,11 @@ exports.seedInfo = async function ({ user, namespace, name }) {
     return data
   }
 
+  if (!seed.spec.secretRef) {
+    logger.info(`Could not fetch info from seed. 'spec.secretRef' on the seed ${seed.metadata.name} is missing. In case a shoot is used as seed, add the flag \`with-secret-ref\` to the \`shoot.gardener.cloud/use-as-seed\` annotation`)
+    return data
+  }
+
   try {
     const seedClient = await client.createKubeconfigClient(seed.spec.secretRef)
     const seedShootNamespace = shoot.status.technicalID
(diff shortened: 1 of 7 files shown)


grolu:
what about asking the gardener team if this change can be reverted. In addition, how do we handle this when fetching information from the backend for the cluster detail page? Do we also log an error about missing secret in this case or only an unspecific seed not reachable error?
Otherwise lgtm

petersutter:
After I thought about it again I think it's just a documentation issue.
@rfranzke is the with-secret-ref flag documented when setting up a seed from a shoot? As long as the seed is reachable from the dashboard `spec.secretRef` is required for the dashboard.

@grolu for the details page we log the error `Failed to retrieve information using seed core client`

grolu:
@petersutter wouldn't it make sense to log a similar message in case retrieving information from the seed fails for this reason, e.g. `Failed to retrieve information using seed core client as 'spec.secretRef' on the seed is missing. In case a shoot is used as seed, add the flag \`with-secret-ref\` to the \`shoot.gardener.cloud/use-as-seed\` annotation`

grolu:
@rfranzke @petersutter Yes, it is ok that the field is optional. But it must be clear that this field is required for the dashboard to work properly.

petersutter:
> @petersutter wouldn't it make sense to log a similar message in case retrieving information from the seed fails for this reason, e.g. `Failed to retrieve information using seed core client as 'spec.secretRef' on the seed is missing. In case a shoot is used as seed, add the flag with-secret-ref to the shoot.gardener.cloud/use-as-seed annotation`

yes makes sense I will also add a proper log message here

rfranzke:
It's not documented yet, should probably be done as part of https://github.com/gardener/gardener/issues/2181

PR #715, opened 2020-06-09
Switch from Journals to Tickets

### LINKED ISSUE #651: User journals / issues
In case of issues with a cluster, users usually contact the gardener operators via mail/slack or other means.
The dashboard should give the users the possibility to create issues for their cluster, similar to the journal feature.

**Why is this needed**:
This gives users and operators a better overview of the created issues.

petersutter:
With this PR the journal feature will be opened up for all project members and isn't anymore exclusive for the gardener operators.

Screenshot 2020-06-09 at 16 18 39

Screenshot 2020-06-09 at 16 27 34

The gardener operator should make sure that all project members have access to the configured github repository in order for the users to create tickets (github issues).

Example `values.yaml`:
```yaml
frontendConfig:
  ticket:
    gitHubRepoUrl: https://foo-github.com/dummyorg/dummyrepo
    hideClustersWithLabel: ignore # hides clusters with label on the 'ALL PROJECTS' page if the respective table option is enabled
...
gitHub:
  apiUrl: https://api.foo-github.com
  org: dummyorg
  repository: dummyrepo
  webhookSecret: foobar # optional if pollIntervalSeconds is defined
  authentication:
    username: dashboard
    token: dummytoken
  # pollIntervalSeconds: 30 # only necessary when dashboard's webhook can't be reached by github and thus polling needs to be done
```

Fixes #651

**Release note**:

```action operator
In case the journal feature is configured you have to adapt your configuration as it changed incompatible. `frontendConfig.gitHubRepoUrl` in `values.yaml` needs to be moved to `frontendConfig.ticket.gitHubRepoUrl`. Otherwise the ticket feature (previously journals) is not shown.
```
```action operator
In case the ticket feature is configured, make sure that all project members have the permission to create and comment issues in the configured github repository.
```
```noteworthy user
You can now create tickets directly from the dashboard in case you have issues with your cluster and need support from the gardener operators. The created tickets for the cluster can be seen on the cluster details page.
```


grolu:
Do we need tests for the charts?

grolu:
I think we already discussed this, if the dashboard is updated without config changes, the old journal function disappears and the new ticket function has to be explicitly enabled as the configuration path is different - so there is no risk of leaking journals to non admin users, right?

petersutter:
> I think we already discussed this, if the dashboard is updated without config changes, the old journal function disappears and the new ticket function has to be explicitly enabled as the configuration path is different - so there is no risk of leaking journals to non admin users, right?

yes correct

petersutter:
> Do we need tests for the charts?

yes I will add tests

PR #723, opened 2020-07-01
Tickets: Use project instead of namespace

petersutter:
Tickets, that are now created will have the project name in the title. Hence also the mapping from ticket to namespace / shoot had to be adapted accordingly.

Fixes #

**Release note**:

```improvement user

```


frontend/src/components/ShootTickets/TicketComment.vue
```diff
@@ -94,6 +94,30 @@ export default {
     word-wrap: break-word;
     /* does not work with firefox */
     word-break: break-word;
+
```
grolu, line 97: why is this css now required?

petersutter: it's not required, however I overwrote the font-size of the headings and adapted them to the github sizes. Previously, I felt they were a bit too large

grolu: okay, I just wanted to confirm because it has nothing to do with the rest of the PR

PR #751, opened 2020-07-29
[master] Improve terminal bootstrapping

### LINKED ISSUE #745: Terminal bootstrapping hangs
**What happened**:
It could happen that the terminal bootstrapping hangs and that no more tasks are run from the queue, in case the queue was not configured with a timeout for the tasks. Only killing / restarting the pod helps to re-start the terminal bootstrapping.

When running a task to bootstrap e.g. a shoot we we could see from the logs that a POST request was initiated to create the ingress for the `kube-apiserver` but it never returned and thus the task never finished.

The current assumption is that the `got` library that we use for the http requests has a bug that causes requests in some cases to not return (e.g. `resolve` / `reject` not called for promise) and we should upgrade to the latest `got` version.

In addition we have to set, by default, a timeout for the tasks on the queue

**Environment**:

petersutter:
This PR improves the terminal bootstrapping
- Adds default timeout of 10 minutes to the bootstrapping task (currently there is a bug in the `got` http request library where requests get stuck and do not return/timeout)
- Stores in-memory which shoots were already processed successfully. Previously all shoots were added to the queue on connection loss / when the watch restarts.
- Cancels the task in case the shoot gets deleted

Fixes #745 

In a second iteration (not part of this PR), the set of successfully bootstrapped shoots could be stored when the dashboard shuts down gracefully and restored once it starts again.

**Release note**:

```improvement user

```

### DIFF
--- a/backend/lib/services/terminals/terminalBootstrap.js
+++ b/backend/lib/services/terminals/terminalBootstrap.js
@@ -19,6 +19,7 @@
 const Queue = require('better-queue')
 const _ = require('lodash')
 const net = require('net')
+const moment = require('moment')
 
 const logger = require('../../logger')
 const config = require('../../config')
@@ -53,39 +54,70 @@ const { getSeed } = require('../../cache')
 
 const TERMINAL_KUBE_APISERVER = 'dashboard-terminal-kube-apiserver'
 
-class BootstrapPendingSet extends Set {
-  _keyForResource (resource) {
-    const kind = resource.kind
-    const { metadata: { name, namespace } } = resource
-    return `${kind}/${namespace}/${name}`
+class BootstrapSet extends Set {
+  static keyForResource () {
+    throw new Error('needs to be overwritten in subclass')
   }
 
   containsResource (resource) {
-    const key = this._keyForResource(resource)
+    const key = this.constructor.keyForResource(resource)
     return this.has(key)
   }
 
   removeResource (resource) {
-    const key = this._keyForResource(resource)
+    const key = this.constructor.keyForResource(resource)
     return this.delete(key)
   }
 
   addResource (resource) {
-    const key = this._keyForResource(resource)
+    const key = this.constructor.keyForResource(resource)
+    return this.addResourceWithKey(key)
+  }
+
+  addResourceWithKey (key) {
     this.add(key)
     return key
   }
 }
 
+class BootstrapPendingSet extends BootstrapSet {
+  static keyForResource (resource) {
+    const kind = resource.kind
+    const { metadata: { name, namespace } } = resource
+    return `${kind}/${namespace}/${name}`
+  }
+}
+
+class BootstrappedSet extends BootstrapSet {
+  static keyForResource (resource) {
+    const { metadata: { uid } } = resource
+    return uid
+  }
+}
+
+function taskIdForResource (resource) {
+  const { metadata: { uid } } = resource
+  return uid
+}
+
 class Handler {
-  constructor (fn, description) {
+  constructor ({ id, fn, description }) {
+    this.id = id
     this.fn = fn
-    this._run = () => fn()
+    this._run = session => fn(session)
     this._description = description
+    this.session = {
+      canceled: false
+    }
   }
 
   run () {
-    return this._run()
+    return this._run(this.session)
+  }
+
+  cancel () {
+    logger.debug(`Cancel called on Handler ${this.description}`)
+    this.session.canceled = true
   }
 
   get description () {
@@ -199,12 +231,11 @@ function replaceServiceKubeApiServer (client, { name = TERMINAL_KUBE_APISERVER, 
   return replaceResource(client.core.services, { namespace, name, body })
 }
 
-async function handleSeed (seed) {
-  const name = seed.metadata.name
+async function handleSeed ({ name }) {
   const namespace = 'garden'
 
   // get latest seed resource from cache
-  seed = getSeed(name)
+  const seed = getSeed(name)
 
   if (!_.isEmpty(seed.metadata.deletionTimestamp)) {
     logger.debug(`Seed ${name} is marked for deletion, bootstrapping aborted`)
@@ -222,8 +253,7 @@ async function handleSeed (seed) {
   }
 }
 
-async function handleShoot (shoot) {
-  const { metadata: { namespace, name } } = shoot
+async function handleShoot ({ name, namespace }) {
   logger.debug(`replacing shoot's apiserver ingress ${namespace}/${name} for webterminals`)
   // read the latest shoot resource version
   const latestShootResource = await dashboardClient['core.gardener.cloud'].shoots.get(namespace, name)
@@ -402,10 +432,11 @@ class Bootstrapper extends Queue {
   constructor () {
     super(Bootstrapper.process, Bootstrapper.options)
     this.bootstrapPending = new BootstrapPendingSet()
+    this.bootstrapped = new BootstrappedSet()
     this.requiredConfigExists = verifyRequiredConfigExists()
     if (this.isBootstrapKindAllowed('gardenTerminalHost')) {
       const description = 'garden host cluster'
-      const handler = new Handler(ensureTrustedCertForGardenTerminalHostApiServer, description)
+      const handler = new Handler({ id: 'gardenTerminalHost', fn: ensureTrustedCertForGardenTerminalHostApiServer, description })
       this.push(handler)
     }
   }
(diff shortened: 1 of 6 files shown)


backend/lib/services/terminals/terminalBootstrap.js
```diff
@@ -53,39 +54,70 @@ const { getSeed } = require('../../cache')
...
+  addResourceWithKey (key) {
     this.add(key)
     return key
   }
 }
 
+class BootstrapPendingSet extends BootstrapSet {
+  static keyForResource (resource) {
+    const kind = resource.kind
+    const { metadata: { name, namespace } } = resource
```
holgerkoser, line 86:
```suggestion
  static keyForResource ({ kind, metadata: { name, namespace } }) {
```

backend/lib/services/terminals/terminalBootstrap.js
```diff
@@ -53,39 +54,70 @@ const { getSeed } = require('../../cache')
 
 const TERMINAL_KUBE_APISERVER = 'dashboard-terminal-kube-apiserver'
 
-class BootstrapPendingSet extends Set {
-  _keyForResource (resource) {
-    const kind = resource.kind
-    const { metadata: { name, namespace } } = resource
-    return `${kind}/${namespace}/${name}`
+class BootstrapSet extends Set {
+  static keyForResource () {
```
holgerkoser, line 58: Pseudo abstract class in javascript? I would use a default implementation.

backend/lib/services/terminals/terminalBootstrap.js
```diff
@@ -53,39 +54,70 @@ const { getSeed } = require('../../cache')
...
+class BootstrappedSet extends BootstrapSet {
+  static keyForResource (resource) {
+    const { metadata: { uid } } = resource
+    return uid
+  }
+}
+
+function taskIdForResource (resource) {
+  const { metadata: { uid } } = resource
+  return uid
```
holgerkoser, line 100: return this.constructor.keyForResource(resource)

holgerkoser, line 100:
```suggestion
return this.constructor.keyForResource(resource)
```

backend/lib/services/terminals/terminalBootstrap.js
```diff
@@ -53,39 +54,70 @@ const { getSeed } = require('../../cache')
...
+}
+
 class Handler {
-  constructor (fn, description) {
+  constructor ({ id, fn, description }) {
+    this.id = id
     this.fn = fn
-    this._run = () => fn()
+    this._run = session => fn(session)
```
holgerkoser, line 107: Why do we store the session on the handler if we pass it to the run method. Or the other way around. Why do we pass the session to the run method if we store it on the handler?

petersutter: so that we can cancel it later on?

backend/lib/services/terminals/terminalBootstrap.js
```diff
@@ -53,39 +54,70 @@ const { getSeed } = require('../../cache')
...
+}
+
+function taskIdForResource (resource) {
+  const { metadata: { uid } } = resource
+  return uid
+}
+
 class Handler {
-  constructor (fn, description) {
+  constructor ({ id, fn, description }) {
```
holgerkoser, line 104: `fn` is not a good name for a named paramter

backend/lib/services/terminals/terminalBootstrap.js
```diff
@@ -432,7 +499,12 @@ class Bootstrapper extends Queue {
   }
 
   bootstrapResource (resource) {
-    const kind = resource.kind
+    const {
+      kind,
+      metadata: { namespace, name, uid }
+    } = resource
+    const description = `${kind} - ${namespace ? namespace + '/' : ''}${name} (${uid})`
```
holgerkoser, line 506:
```suggestion
    const qualifiedName = namespace ? namespace + '/' + name : name
    const description = `${kind} - ${qualifiedName} (${uid})`
```

backend/lib/services/terminals/terminalBootstrap.js
```diff
@@ -455,44 +530,79 @@ class Bootstrapper extends Queue {
...
+      id: taskId, // with the id we make sure that the task for one shoot is not added multiple times (e.g. on another ADDED event when the shoot watch is re-established)
+      fn: async session => {
+        if (session.canceled) {
+          logger.debug(`Canceling handler of ${description} as requested`)
+          return
+        }
+
+        switch (kind) {
+          case 'Seed': {
+            await handleSeed({ name })
```
holgerkoser, line 554: why no try catch

backend/lib/services/terminals/terminalBootstrap.js
```diff
@@ -455,44 +530,79 @@ class Bootstrapper extends Queue {
...
+      try {
+        await handler.run()
+        cb(null, null)
+      } catch (err) {
+        logger.error(`failed to bootstrap ${handler.description}`, err)
+        cb(err, null)
+      }
+    })()
+    return {
+      cancel: function () {
```
holgerkoser, line 602:
```suggestion
      cancel () {
```

backend/lib/services/terminals/terminalBootstrap.js
holgerkoser, line 104:
```suggestion
  constructor (fn, { id, description }) {
```

holgerkoser: the handler function is mandatory. Id and description seem to be optional.

backend/lib/services/terminals/terminalBootstrap.js
```diff
@@ -402,10 +432,11 @@ class Bootstrapper extends Queue {
   constructor () {
     super(Bootstrapper.process, Bootstrapper.options)
     this.bootstrapPending = new BootstrapPendingSet()
+    this.bootstrapped = new BootstrappedSet()
     this.requiredConfigExists = verifyRequiredConfigExists()
     if (this.isBootstrapKindAllowed('gardenTerminalHost')) {
       const description = 'garden host cluster'
-      const handler = new Handler(ensureTrustedCertForGardenTerminalHostApiServer, description)
+      const handler = new Handler({ id: 'gardenTerminalHost', fn: ensureTrustedCertForGardenTerminalHostApiServer, description })
```
holgerkoser, line 439:
```suggestion
      const handler = new Handler(ensureTrustedCertForGardenTerminalHostApiServer, { 
        id: 'gardenTerminalHost', 
        description 
      })
```

backend/lib/services/terminals/terminalBootstrap.js
```diff
@@ -455,44 +530,79 @@ class Bootstrapper extends Queue {
...
+
+        if (session.canceled) { // do not add key to the bootstrapped set when the session is canceled (due to shoot deletion) to prevent leaking memory
+          logger.debug(`Canceling handler of ${description} as requested after handling resource`)
+          return
+        }
+        logger.debug(`Successfully bootstrapped ${description}`)
+        this.bootstrapped.addResourceWithKey(key)
+      },
+      description
+    })
```
holgerkoser, line 575:
```suggestion
    const fn = async session => {
        if (session.canceled) {
          logger.debug(`Canceling handler of ${description} as requested`)
          return
        }

        switch (kind) {
          case 'Seed': {
            await handleSeed({ name })
            break
          }
          case 'Shoot': {
            await handleShoot({ name, namespace })
            break
          }
          default: {
            logger.error(`can't bootstrap unsupported kind ${kind}`)
            return
          }
        }

        if (session.canceled) { // do not add key to the bootstrapped set when the session is canceled (due to shoot deletion) to prevent leaking memory
          logger.debug(`Canceling handler of ${description} as requested after handling resource`)
          return
        }
        logger.debug(`Successfully bootstrapped ${description}`)
        this.bootstrapped.addResourceWithKey(key)
      }
    const handler = new Handler(fn, {
      id: taskId, // with the id we make sure that the task for one shoot is not added multiple times (e.g. on another ADDED event when the shoot watch is re-established)
      description
    })
```

backend/test/services.terminals.spec.js
```diff
@@ -659,7 +662,7 @@ describe('services', function () {
         const seed = getSeed(seedName)
         createConfigStub({ bootstrap })
         const bootstrapper = new Bootstrapper()
-        bootstrapper.push(new Handler(() => {}, 'test'))
+        bootstrapper.push(new Handler({ fn: () => {}, description: 'test' }))
```
holgerkoser, line 665:
```suggestion
        bootstrapper.push(new Handler(() => {}, { description: 'test' }))
```

backend/lib/services/terminals/terminalBootstrap.js
holgerkoser: Or return the handler?

backend/lib/services/terminals/terminalBootstrap.js
```diff
@@ -455,44 +530,79 @@ class Bootstrapper extends Queue {
...
 
-  static async process (handler, cb) {
-    try {
-      await handler.run()
-      cb(null, null)
-    } catch (err) {
-      logger.error(`failed to bootstrap ${handler.description}`, err)
-      cb(err, null)
+  static process (handler, cb) {
+    (async () => {
```
holgerkoser, line 592: Move the background processing into handler run?

PR #752, opened 2020-07-29
Avoid false bottoms

grolu:
Currently our dialogs tend to have false bottoms. For users it is often not clear that there is more content, that needs to be scrolled into view. For Mac users this is even more an issue as scrollbars are hidden on macOS by default. With this PR we show the scrollbar briefly when a dialog is opened. This indicates that there is more content and hopefully helps to mitigate this problem.

Fixes #

**Release note**:

```improvement user

```

### DIFF
--- a/frontend/src/components/dialogs/GDialog.vue
+++ b/frontend/src/components/dialogs/GDialog.vue
@@ -15,9 +15,9 @@ limitations under the License.
  -->
 
 <template>
-  <v-dialog v-model="visible" v-if="visible" scrollable persistent :max-width="maxWidth" @keydown.esc="resolveAction(false)">
+  <v-dialog v-model="visible" scrollable persistent :max-width="maxWidth" @keydown.esc="resolveAction(false)">
     <v-card>
-      <v-toolbar flat :class="titleColorClass">
+      <v-toolbar flat :class="[{ 'elevation-4': topBarElevation }, titleColorClass]">
         <v-toolbar-title class="dialog-title align-center justify-start">
           <slot name="caption">
             Confirm Dialog
@@ -28,7 +28,7 @@ limitations under the License.
           </template>
         </v-toolbar-title>
       </v-toolbar>
-      <v-card-text class="pa-3" :style="{'max-height': maxHeight}" ref="contentCard" @scroll="onContentScrolled()">
+      <v-card-text class="pa-3" :style="{'max-height': maxHeight}" ref="contentCard">
         <slot name="message">
           This is a generic dialog template.
         </slot>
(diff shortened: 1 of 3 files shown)


frontend/src/components/dialogs/GDialog.vue
```diff
@@ -202,6 +203,23 @@ export default {
         resolve(value)
       }
       this.visible = false
+    },
+    showScrollBar () {
+      const contentCardRef = this.$refs.contentCard
+      if (!contentCardRef || !contentCardRef.clientHeight) {
+        this.$nextTick(() => this.showScrollBar())
```
petersutter, line 210: Potential endless loop?

frontend/src/components/dialogs/GDialog.vue
```diff
@@ -108,7 +108,8 @@ export default {
     return {
       userInput: '',
       visible: false,
-      resolve: noop
+      resolve: noop,
+      intervalId: undefined
```
petersutter, line 112: Unused?

frontend/src/components/dialogs/GDialog.vue
grolu: hmm yes could be if the ref is never set. However, this should not be the case. Anyway, if you find a better solution it would be highly appreciated

frontend/src/components/dialogs/GDialog.vue
grolu: removed

frontend/src/components/dialogs/GDialog.vue
```diff
@@ -202,6 +202,23 @@ export default {
...
+      }
+      const scrollTopVal = contentCardRef.scrollTop
+      contentCardRef.scrollTop = scrollTopVal + 10
+      contentCardRef.scrollTop = scrollTopVal - 10
+    }
+  },
+  watch: {
+    visible (value) {
+      if (value) {
+        this.showScrollBar()
```
petersutter, line 220: why not call this.showScrollBar() surrounded with nextTick here instead of having the nextTick within the showScrollBar.
If you want to keep it in the showScrollBar then an abort condition would be good by passing either a boolean or a counter to the showScrollBar function

frontend/src/components/dialogs/GDialog.vue
petersutter: see comment below

frontend/src/components/dialogs/GDialog.vue
grolu: The reason ist that one next tick is sometimes not sufficient, it takes some time until all refs are set.

petersutter: then we should use a counter/tries parameter

grolu: done

frontend/src/components/dialogs/GDialog.vue
grolu: added circuit breaker

frontend/src/components/dialogs/GDialog.vue
```diff
@@ -202,6 +202,27 @@ export default {
         resolve(value)
       }
       this.visible = false
+    },
+    showScrollBar (retryCount = 0) {
+      if (retryCount > 10) {
```
holgerkoser, line 207:
```suggestion
      if (!this.visible || retryCount > 10) {
```

PR #755, opened 2020-07-29
Fixed secret handling for cases where secretBindingName != secretName

### LINKED ISSUE #701: Secret shown as 'currently unused' when there are Shoots using it
**What happened**:
Cloud provider Secret shown as 'currently unused' when there are Shoots using it. This happens when the SecretBinding `.metadata.name` != Secret `.metadata.name`.

1. Create `Secret`
```yaml
apiVersion: v1
data:
  accessKeyID: "accessKeyID"
  secretAccessKey: "secretAccessKey"
kind: Secret
metadata:
  name: shoot-aws-secret-bar
  namespace: garden-baz
type: Opaque
```

2. Create `SecretBinding`
```yaml
kind: SecretBinding
metadata:
  name: shoot-aws-secret-foo
  namespace: garden-baz
  labels:
    cloudprofile.garden.sapcloud.io/name: aws
secretRef:
  name: shoot-aws-secret-bar
  namespace: garden-baz
```

3. Create `Shoot`
```yaml
kind: Shoot
metadata:
  name: test
  namespace: garden-baz
spec:
  # [...]
  secretBindingName: shoot-aws-secret-foo
```

4. Ensure that the `shoot-aws-secret-bar` Secret in the dashboard is shown as `shoot-aws-secret-bar (currently unused)`

**Environment**:
Dashboard version: `1.42.0-dev-22049d66`

grolu:
Secret Bindings where name is not equal to Secret name did not show correct number of used clusters. Furthermore these secrets did not open on navigation (even if own secrets).
Also, I fixed the delete secret dialog which showed up blank (without any message), probably we broke the dialog with the latest vuetify upgrade.

Fixes #701 

**Release note**:

```improvement user
Fixed: Some Secrets were shown as 'currently unused' even if used by clusters. This happened for Secrets where the Secret Binding name is different from the Secret name
```

### DIFF
--- a/frontend/src/components/dialogs/SecretDialogDelete.vue
+++ b/frontend/src/components/dialogs/SecretDialogDelete.vue
@@ -36,12 +36,8 @@ limitations under the License.
 
       <v-card-text>
         <v-container fluid>
-          <template v-slot:message>
-            <div>
-              Are you sure to delete the secret <span class="font-weight-bold">{{name}}</span>?<br/>
-              <span class="red--text font-weight-bold">The operation can not be undone.</span>
-            </div>
-          </template>
+          Are you sure to delete the secret <span class="font-weight-bold">{{name}}</span>?<br/>
+          <span class="red--text font-weight-bold">The operation can not be undone.</span>
         </v-container>
         <g-alert color="error" :message.sync="errorMessage" :detailedMessage.sync="detailedErrorMessage"></g-alert>
       </v-card-text>
--- a/frontend/src/components/SecretRow.vue
+++ b/frontend/src/components/SecretRow.vue
@@ -97,9 +97,9 @@ export default {
       return this.shootsByInfrastructureSecret.length
     },
     shootsByInfrastructureSecret () {
-      const secretName = this.secret.metadata.name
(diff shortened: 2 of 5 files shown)


frontend/src/components/dialogs/SecretDialogDelete.vue
```diff
@@ -36,12 +36,8 @@ limitations under the License.
 
       <v-card-text>
         <v-container fluid>
-          <template v-slot:message>
-            <div>
-              Are you sure to delete the secret <span class="font-weight-bold">{{name}}</span>?<br/>
-              <span class="red--text font-weight-bold">The operation can not be undone.</span>
-            </div>
-          </template>
+          Are you sure to delete the secret <span class="font-weight-bold">{{name}}</span>?<br/>
```
petersutter, line 39: what happened to the `<template v-slot:message>`? It's not needed anymore?

grolu: I asked myself the same question. I don't know but the slot doesn't exist anymore. This was the reason why the text did not show up. I also deleted the div as it is not required, but the template slot thingy was the reason for the issue

PR #756, opened 2020-07-30
[hotfix-1.43] Upgrade http client library 'got' to latest version

holgerkoser:
We have observed situations where the `kube-apiserver` is throttling requests from the dashboard. In the log we see that a request does not return and the reflector loop is blocked.
log
There are some closed iusses in `got` (e.g. [1341](https://github.com/sindresorhus/got/issues/1341), [1295](https://github.com/sindresorhus/got/issues/1295) and [1287](https://github.com/sindresorhus/got/issues/1287)) which describe problems with hanging requests. These issues could be the reason for cache expiration problems in the dashboard backend.

Fixes #

**Release note**:

```improvement operator
Upgrade http client library `got` to latest version to get rid of cache expiration problems.
```


VERSION
```diff
@@ -1 +1 @@
-1.43.3-dev
\ No newline at end of file
+1.43.2
```
petersutter, line 1: revert change

holgerkoser: Ja.Habich auch gerade gesehen. 👍

backend/lib/kubernetes-client/cache/Reflector.js
```diff
@@ -133,29 +121,52 @@ class Reflector {
...
+      logger.debug('Closing socket of %s has been triggered', this.expectedTypeName)
+      await pEvent(this.socket, 'close', { timeout })
+      logger.debug('Closed socket of %s', this.expectedTypeName)
+    } catch (err) {
+      logger.debug('Terminating socket of %s', this.expectedTypeName)
+      this.socket.terminate()
+      logger.debug('Terminated socket of %s', this.expectedTypeName)
+    }
+  }
+
```
petersutter, line 142: I guess we don't need the excessive log here in this function anymore, right?
Also you could move `await pEvent..` on top of `this.socket.close`, just to be sure (as you had it previously)

backend/lib/kubernetes-client/cache/Reflector.js
```diff
@@ -272,7 +288,9 @@ class Reflector {
           return
         }
       } finally {
-        await closeOrTerminateSocket(this.socket, this.gracePeriod.asMilliseconds())
+        logger.info('Trying to gracefully close socket of %s', this.expectedTypeName)
+        await this.closeOrTerminateSocket()
+        logger.info('Closed or terminated socket of %s', this.expectedTypeName)
```
petersutter, line 293: remove logs or at least one of them

backend/lib/kubernetes-client/cache/Reflector.js
holgerkoser: ✔️

backend/lib/kubernetes-client/cache/Reflector.js
holgerkoser: ✔️

PR #765, opened 2020-08-03
Cluster list actions

### LINKED ISSUE #707: Please add support for hibernate/wake-up as action in the clusters list
It would be nice to have hibernate/wake-up as action in the clusters list. There I can find only the kubeconfig and deletion, but hibernate/wake-up I do more often.

Screenshot 2020-05-29 at 06 31 08

**Why is this needed**:
Convenience, no hard reason.

**Remark**
There is also https://github.com/gardener/dashboard/issues/631, but it's about mass operations. Still, I link it because it would require to have the operation available first. The users example was also about hibernation.

grolu:
This PR adds the ability to trigger cluster actions directly from the cluster list without navigating to the cluster details page.

Screen Shot 2020-08-12 at 16 12 53

Fixes #707

**Release note**:

```noteworthy user
Cluster actions can now be accessed directly from the cluster list without navigating to the cluster details page
```


holgerkoser:
According to [Material Design](https://material.io/components/menus#anatomy) a Text and icon list menu should look like this
image
Icon on the left, text not uppercase, no headline in the menu.

holgerkoser:
I would add at least one separator before `Delete Cluster`

frontend/src/components/DeleteCluster.vue
```diff
@@ -15,7 +15,7 @@ limitations under the License.
 -->
 
 <template>
-  <action-icon-dialog
+  <action-button-dialog
```
holgerkoser, line 18: The name should be something like `action-dropdown-menu`. It is not a dialog and in the menu shouldn't occur buttons.

frontend/src/components/ShootListRowActions.vue
```diff
@@ -0,0 +1,96 @@
...
+        <template v-slot:activator="{ on: tooltip }">
+          <v-btn v-on="{ ...menu, ...tooltip}" icon class="cyan--text text--darken-2">
+            <v-icon class="cursor-pointer">more_vert</v-icon>
+          </v-btn>
+        </template>
+        Cluster Actions
+      </v-tooltip>
+    </template>
+    <v-list subheader dense class="actionMenuItem" @click.native.capture="actionMenu=false">
+      <v-subheader>Cluster Actions</v-subheader>
```
holgerkoser, line 30: Remove subheader

frontend/src/components/DeleteCluster.vue
holgerkoser, line 18: Do not use this component in the menu or add more config options.

frontend/src/components/dialogs/ActionButtonDialog.vue
```diff
@@ -16,23 +16,29 @@ limitations under the License.
 
 <template>
   <div v-if="canPatchShoots">
```
holgerkoser, line 18: I think it is strange to decide inside a component that it should not render something. This should be done outside even if it redundant. What makes sense inside the component would be to disable the action.

grolu: Okay but this has nothing to do with this PR. This is logic related to the member roles / permissions. I would suggest to change this when we introduce the uam role, which I will be working on next.

PR #768, opened 2020-08-05
Support configuring aws ios value via UI

### LINKED ISSUE #677: Allow setting of IOPS for AWS Worker-groups
Currently, the Dashboard only allows the configuration of the Volume Type (`gp2` or `io1`) for workers on AWS shoot. For both volume types, there is however the possiblility to provide a value for the desired IOPS using the shoot-spec.
It'd be nice if this configuration could also be done through the Dashboard's `Configure Workers` dialog.

**Why is this needed**:
Makes the configuration of these values easier for end-users. Also, in case of `io1` a value has to be given, so additional hints/verifications could be added here.

grolu:
This PR adds support for configuring IOPS value for aws volumes.

Screen Shot 2020-08-05 at 16 11 26

Fixes #677 

**Release note**:

```improvement user
IOPs value for aws volumes can now be configured via the Dashboard UI
```

### DIFF
--- a/frontend/src/components/ShootWorkers/VolumeType.vue
+++ b/frontend/src/components/ShootWorkers/VolumeType.vue
@@ -90,7 +90,7 @@ export default {
           }
         },
         internalIops: {
-          required: requiredIf(function () {
+          required: requiredIf(() => {
             return this.isAWS && this.worker.volume.type === 'io1'
           }),
           minValue: minValue(100),
--- a/frontend/src/components/ShootWorkers/VolumeType.vue
+++ b/frontend/src/components/ShootWorkers/VolumeType.vue
@@ -32,7 +32,6 @@
       v-model.number="workerIops"
       type="number"
       min="100"
-      :max="maxIops"
       label="IOPS">
     </v-text-field>
   </div>
@@ -42,7 +41,7 @@
 
 import { mapGetters } from 'vuex'
 import HintColorizer from '@/components/HintColorizer'
-import { required, requiredIf, minValue, maxValue } from 'vuelidate/lib/validators'
+import { required, requiredIf, minValue } from 'vuelidate/lib/validators'
 import { getValidationErrors } from '@/utils'
 import { getWorkerProviderConfig } from '@/utils/createShoot'
 import find from 'lodash/find'
--- a/frontend/src/components/ShootWorkers/VolumeType.vue
+++ b/frontend/src/components/ShootWorkers/VolumeType.vue
@@ -29,7 +29,7 @@
       :error-messages="getErrorMessages('internalIops')"
       @input="onInputIops"
       @blur="$v.internalIops.$touch()"
-      v-model="workerIops"
+      v-model.number="workerIops"
       type="number"
       min="100"
       :max="maxIops"
--- a/frontend/src/components/ShootWorkers/VolumeType.vue
+++ b/frontend/src/components/ShootWorkers/VolumeType.vue
@@ -1,53 +1,54 @@
 <template>
-  <hint-colorizer hintColor="orange">
-    <v-select
+  <div class="d-flex flex-row">
+    <hint-colorizer hintColor="orange">
+      <v-select
+        color="cyan darken-2"
+        item-color="cyan darken-2"
+        :items="volumeTypeItems"
+        item-text="name"
+        item-value="name"
+        v-model="worker.volume.type"
+        :error-messages="getErrorMessages('worker.volume.type')"
+        @input="onInputVolumeType"
+        @blur="$v.worker.volume.type.$touch()"
+        label="Volume Type"
+        :hint="hint"
+        persistent-hint>
+        <template v-slot:item="{ item }">
+          <v-list-item-content>
+            <v-list-item-title>{{item.name}}</v-list-item-title>
+            <v-list-item-subtitle v-if="item.class">Class: {{item.class}}</v-list-item-subtitle>
+          </v-list-item-content>
+        </template>
+      </v-select>
+    </hint-colorizer>
+    <v-text-field
+      v-if="isAWS"
+      class="ml-1"
       color="cyan darken-2"
-      item-color="cyan darken-2"
-      :items="volumeTypeItems"
-      item-text="name"
-      item-value="name"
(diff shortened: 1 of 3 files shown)


frontend/src/components/ShootWorkers/VolumeType.vue
```diff
@@ -88,16 +132,51 @@ export default {
...
+          if (!this.worker.providerConfig) {
+            this.worker.providerConfig = getWorkerProviderConfig('aws')
+          }
+          set(this.worker.providerConfig, 'volume.iops', parseInt(value))
+        } else {
+          unset(this.worker.providerConfig, 'volume.iops')
+        }
+      }
+    },
+    maxIops () {
```
holgerkoser, line 156: hardcoded volume types?

frontend/src/components/ShootWorkers/VolumeType.vue
```diff
@@ -61,15 +62,58 @@ export default {
...
+        worker: {
+          volume: {
+            type: {
+              required
+            }
+          }
+        },
+        internalIops: {
+          required: requiredIf(function () {
+            return this.isAWS && this.worker.volume.type === 'io1'
```
holgerkoser, line 94: Why is `this` the component here?

frontend/src/components/ShootWorkers/VolumeType.vue
```diff
@@ -88,16 +132,51 @@ export default {
...
+      const cloudProfile = this.cloudProfileByName(this.cloudProfileName)
+      return get(cloudProfile, 'metadata.cloudProviderKind') === 'aws'
+    },
+    workerIops: {
+      get () {
+        return this.internalIops
+      },
+      set (value) {
+        this.internalIops = value
+        if (value > 0) {
```
holgerkoser, line 146: why not parseInt at the beginning

frontend/src/components/ShootWorkers/VolumeType.vue
```diff
@@ -88,16 +132,51 @@ export default {
...
     },
     onInputVolumeType () {
       this.$v.worker.volume.type.$touch()
-      this.$emit('updateVolumeType', this.worker.volume.type)
+      this.$emit('updateVolumeType')
+      this.validateInput()
+    },
+    onInputIops () {
+      this.$v.internalIops.$touch()
+      this.$emit('updateVolumeType')
```
holgerkoser, line 179: why `updateVolumeType` event

frontend/src/components/ShootWorkers/VolumeType.vue
```diff
@@ -61,15 +62,58 @@ export default {
...
+    cloudProfileName: {
+      type: String
     }
   },
   data () {
     return {
-      validationErrors,
-      valid: undefined
+      valid: undefined,
+      internalIops: undefined
```
holgerkoser, line 73: Why do we need the `internalIops`. We already have an eventHandler for @input.

frontend/src/components/ShootWorkers/VolumeType.vue
```diff
@@ -1,53 +1,54 @@
...
-          <v-list-item-title>{{item.name}}</v-list-item-title>
-          <v-list-item-subtitle v-if="item.class">Class: {{item.class}}</v-list-item-subtitle>
-        </v-list-item-content>
-      </template>
-    </v-select>
-  </hint-colorizer>
+      :error-messages="getErrorMessages('internalIops')"
+      @input="onInputIops"
+      @blur="$v.internalIops.$touch()"
+      v-model="workerIops"
```
holgerkoser, line 32:
```suggestion
      v-model.number="workerIops"
```

holgerkoser: Use the https://vuejs.org/v2/guide/forms.html#number modifier.

frontend/src/components/ShootWorkers/VolumeType.vue
grolu: Yes according to the aws specifications, this values should not change, there is no point in making them configurable and we can also not read them from anywhere, right? @rfranzke

frontend/src/components/ShootWorkers/VolumeType.vue
grolu: we need to inform the parent component, that the cild component changed so that a re-validation takes place. We could also raise a different event here, but I see no value in differentiating between volume type or iops value change as no data is transferred via the vent and the only purpose is to inform that _something_ changed.

frontend/src/components/ShootWorkers/VolumeType.vue
grolu: Actually, the whole point of the internalIops has a reactivity background. We cannot operate on `worker.providerConfig.volume.iops` as the path may not exists initially and therefore validation does not work correctly. If you can show me a way that makes this work without the ìnternal` thingy, please let me know.

frontend/src/components/ShootWorkers/VolumeType.vue
grolu: You mean why does it work as it is no arrow function? Actually a good question, I don't know. But we can change it to an arrow function if you want.

frontend/src/components/ShootWorkers/VolumeType.vue
rfranzke: There are some limits imposed by AWS, https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ebs-volume-types.html, but now taking a look at it it seems that they have increased them.

I think for `gp2` its `[100,16000]` meanwhile, and for `io1` its `[100,64000]`.

frontend/src/components/ShootWorkers/VolumeType.vue
grolu: If I would do that, no value would result in 0, which would lead to a validation error (minimum value) even for gp2 volumes that do not require a value at all.

frontend/src/components/ShootWorkers/VolumeType.vue
grolu: =>changed to arrow function

PR #770, opened 2020-08-07
Show link only for shooted Seeds

grolu:
This is an implementation proposal that solves the issue that we currently show links to all seeds, even if no corresponding shoot exists in the garden namespace.

Fixes #

**Release note**:

```improvement user

```

### DIFF
--- a/backend/lib/services/cloudprofiles.js
+++ b/backend/lib/services/cloudprofiles.js
@@ -18,6 +18,7 @@
 
 const { NotFound } = require('../errors')
 const logger = require('../logger')
+const shoots = require('./shoots')
 const _ = require('lodash')
 const { getCloudProfiles, getVisibleAndNotProtectedSeeds } = require('../cache')
 
@@ -31,19 +32,22 @@ function fromResource ({ cloudProfile: { metadata, spec }, seeds }) {
   return { metadata, data }
 }
 
-function fromSeedResource ({ metadata, spec }) {
+function fromSeedResource ({ metadata, spec }, gardenShoots) {
   metadata = _.pick(metadata, ['name'])
   const provider = _.get(spec, 'provider')
   const volume = _.get(spec, 'volume')
   const data = { volume, ...provider }
-  return { metadata, data }
+  const canGetShootForSeed = _.some(gardenShoots, shoot => {
+    return shoot.metadata.name === metadata.name
+  })
+  return { metadata, data, canGetShootForSeed }
 }
 
 function emptyToUndefined (value) {
   return _.isEmpty(value) ? undefined : value
 }
 
-function assignSeedsToCloudProfileIteratee (seeds) {
+function assignSeedsToCloudProfileIteratee (seeds, gardenShoots) {
   return cloudProfileResource => {
     const providerType = cloudProfileResource.spec.type
     const matchLabels = _.get(cloudProfileResource, 'spec.seedSelector.matchLabels', {})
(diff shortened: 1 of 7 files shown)


backend/lib/services/cloudprofiles.js
```diff
@@ -18,6 +18,7 @@
 
 const { NotFound } = require('../errors')
 const logger = require('../logger')
+const shoots = require('./shoots')
```
holgerkoser, line 21: The api of the services modules is designed to be used from the routes directly. We should be careful with cross service reuse because of cycling dependancies.

backend/lib/services/cloudprofiles.js
```diff
@@ -79,15 +84,16 @@ exports.list = function () {
...
+exports.read = async function ({ name, user }) {
   const cloudProfiles = getCloudProfiles()
   const cloudProfileResource = _.find(cloudProfiles, ['metadata.name', name])
   if (!cloudProfileResource) {
     throw new NotFound(`Cloud profile with name ${name} not found`)
   }
 
   const seeds = getVisibleAndNotProtectedSeeds()
-  const cloudProfile = assignSeedsToCloudProfileIteratee(seeds)(cloudProfileResource)
+  const { items: gardenShoots } = await shoots.list({ user, namespace: 'garden' })
```
holgerkoser, line 95:
```suggestion
  const client = user.client
  const { items: gardenShoots } = await client['core.gardener.cloud'].shoots.list('garden')
```

backend/lib/services/cloudprofiles.js
```diff
@@ -31,19 +32,22 @@ function fromResource ({ cloudProfile: { metadata, spec }, seeds }) {
...
+function fromSeedResource ({ metadata, spec }, gardenShoots) {
   metadata = _.pick(metadata, ['name'])
   const provider = _.get(spec, 'provider')
   const volume = _.get(spec, 'volume')
   const data = { volume, ...provider }
-  return { metadata, data }
+  const canGetShootForSeed = _.some(gardenShoots, shoot => {
+    return shoot.metadata.name === metadata.name
+  })
+  return { metadata, data, canGetShootForSeed }
```
holgerkoser, line 43: Why do we introduce a new property on top level. The idea has been to have metadata and data. Why not have this in metadata `isShootedSeed` or something like this.

backend/lib/services/cloudprofiles.js
holgerkoser: If the user has no authorization for ns garden?

holgerkoser:
The cloudProfile has been a resource which could be considered as configuration. Therefore the information could be easily cached for all kind of users. With this PR is get's a user specific dynamic behavior which in many cases is not needed when only the static cloudProfile Information is needed. I would not add the information if the seed is a shoot and can be accessed by the user to the cloudProfiles backend resource.

backend/lib/services/cloudprofiles.js
grolu: we can move it to metadata. The reason for the naming is that `isShootedSeed` indicates that this value gives information about this. Actually, as we make the call with the user token, an admin may receive `true` while a regular user would receive `false`, as he cannot access the `garden` namespace. So the only real information is, if we can _access_ the shoot resource in the `garden` namespace.

backend/lib/services/cloudprofiles.js
grolu: see comment above

holgerkoser:
Replaced by #779

PR #773, opened 2020-08-10
Small UI Bug Fixes

grolu:
This is another bunch of small UI fixes

Fixes #

**Release note**:

```improvement user

```

### DIFF
--- a/frontend/src/components/StatusTag.vue
+++ b/frontend/src/components/StatusTag.vue
@@ -200,7 +200,7 @@ export default {
         return 'error'
       }
       if (this.isUnknown) {
-        return 'grey lighten-1'
+        return 'grey lighten'
       }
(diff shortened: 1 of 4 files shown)


frontend/src/components/StatusTag.vue
```diff
@@ -200,7 +200,7 @@ export default {
         return 'error'
       }
       if (this.isUnknown) {
-        return 'grey lighten-1'
+        return 'grey lighten'
```
holgerkoser, line 203: There is no class `lighten`

grolu: Sorry, wrong class name, should be `grey` only. We don't want to lighten the grey (as your last feedback was to make the grey darker). I fixed this.

PR #776, opened 2020-08-11
Removed kyma addon support

grolu:

Fixes #

**Release note**:

```noteworthy user
The experimental Kyma addon has been removed from Gardener Dashboard. You cannot use the Dashboard to add the Kyma addon anymore. The Kyma access card has also been removed from the cluster details page
```


backend/lib/services/shoots.js
```diff
@@ -135,58 +131,14 @@ exports.replacePurpose = async function ({ user, namespace, name, body }) {
 
 exports.replaceAddons = async function ({ user, namespace, name, body }) {
   const client = user.client
-  const { kyma = {}, ...addons } = body
+  const { ...addons } = body
```
holgerkoser, line 134:
```suggestion
  const addons = body
```

PR #777, opened 2020-08-11
Show warning for stale projects

### LINKED ISSUE #721: Display stale information on project
Projects that are not actively used for a certain amount of time are marked as stale. This information should be displayed on the UI.
See also https://github.com/gardener/gardener/blob/master/docs/concepts/controller-manager.md#project-controller for more details

**Why is this needed**:
So that a user is informed before a project is cleaned up.

Related gardener PR https://github.com/gardener/gardener/pull/2446

grolu:
Gardener flags inactive projects as stale and auto removes them after some time. This PR adds support for this information to the Dashboard. We show information about stale projects and when the deletion will take place.

Screen Shot 2020-08-12 at 16 34 38
Screen Shot 2020-08-12 at 16 33 02

Fixes #721

**Release note**:

```improvement user
The Dashboard now shows if projects are flagged as stale. If projects have an auto deletion timestamp assigned, information about when the deletion will happen is also shown
```

### DIFF
--- a/backend/lib/services/projects.js
+++ b/backend/lib/services/projects.js
@@ -28,10 +28,11 @@ const cache = require('../cache')
 
 const PROJECT_INITIALIZATION_TIMEOUT = 30 * 1000
 
-function fromResource ({ metadata, spec = {} }) {
+function fromResource ({ metadata, spec = {}, status = {} }) {
   const role = 'project'
   const { name, resourceVersion, creationTimestamp, annotations } = metadata
   const { namespace, createdBy, owner, description, purpose } = spec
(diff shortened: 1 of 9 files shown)


holgerkoser:
I would use `mdi-clock-alert-outline` if the project is stale and `mdi-delete-clock` if the deletion timestamp is assigned without colors. Use these icons in the administration page and no icon in the text.

grolu:
Actually, I tried to use the `mdi-delete-clock` icon but it is not available. Probably because the mdi icon font has not been updated since a year or so. The mdi library is far ahead.

PR #779, opened 2020-08-12
Show link only for shooted Seeds

holgerkoser:
This is an implementation proposal that solves the issue that we currently show links to all seeds, even if no corresponding shoot exists in the garden namespace.

Fixes #

**Release note**:

```improvement user

```


petersutter:
cc @mandelsoft @MartinWeindel

PR #780, opened 2020-08-13
Added UAM role and respect new manage-member verb

grolu:

Fixes #

**Release note**:

```improvement user
Added support for managing UAM role via the Gardener Dashboard. Respect the new `manage-members` rbac verb to decide if buttons for project member management are visible
```

### DIFF
--- a/frontend/src/views/Administration.vue
+++ b/frontend/src/views/Administration.vue
@@ -84,7 +84,7 @@ limitations under the License.
                     <v-list-item-subtitle>Technical Contact</v-list-item-subtitle>
                     <v-list-item-title>
                       <editable-account
-                        :read-only="!canPatchProject"
+                        :read-only="!canManageMembers"
                         :color="color"
                         :value="technicalContact"
                         :items="userList"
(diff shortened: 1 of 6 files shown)


frontend/src/views/Administration.vue
```diff
@@ -84,7 +84,7 @@ limitations under the License.
                     <v-list-item-subtitle>Technical Contact</v-list-item-subtitle>
                     <v-list-item-title>
                       <editable-account
-                        :read-only="!canPatchProject"
+                        :read-only="!canManageMembers"
```
holgerkoser, line 87: Why is this part of `manager-members` and not `patch-project`?

grolu: Because this changes the owner of the project, which is by the way also represented as role. Moreover thr owner explicitly gets manage-members rights so an admin could escalate his privileges. I tested it, Gardener also rejects the request without manage-member permission.

holgerkoser: So as an admin I cannot change the project members. This is a very strange naming but not our problem.

grolu: Yes exactly, but this is per design. The owner always has the right, but the `uam` role is _next_ to the admin permission, to allow external uam management. Admins shall not (per default) get the permission to take over user management so I guess one can argument that an `admin` in a project has still all rights in regards to the resources within the project, but as the project management can be external, this role is required in addition.

PR #789, opened 2020-08-26
Small changes

grolu:
Some small date / text / icon changes, as discussed.

Fixes #

**Release note**:

```improvement user

```

### DIFF
--- a/frontend/src/components/RotateKubeconfigStart.vue
+++ b/frontend/src/components/RotateKubeconfigStart.vue
@@ -75,7 +75,7 @@ export default {
       if (this.isActionToBeScheduled) {
         return 'Requesting to schedule kubeconfig credentials rotation'
       }
-      return this.buttonText
+      return this.buttonTitle
     },
     buttonTitle () {
       return 'Start Kubeconfig Rotation'
@@ -84,7 +84,7 @@ export default {
       if (!this.text) {
         return
       }
-      return this.buttonTitle
+      return 'Rotate Kubeconfig'
     }
   },
   methods: {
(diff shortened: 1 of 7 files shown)


frontend/src/components/RotateKubeconfigStart.vue
```diff
@@ -84,7 +84,7 @@ export default {
       if (!this.text) {
         return
       }
-      return this.buttonTitle
+      return 'Rotate Kubeconfig'
```
petersutter, line 87: fix reconcile deactivated issue

frontend/src/utils/index.js
```diff
@@ -726,6 +726,7 @@ export function k8sVersionExpirationForShoot (shootK8sVersion, shootCloudProfile
   }
   return {
     expirationDate: version.expirationDate,
+    expirationDateString: getDateFormatted(version.expirationDate),
```
holgerkoser, line 729: Why do we add a formatted date String here and not do the formatted in the component template via computed property or method?

frontend/src/components/RotateKubeconfigStart.vue
grolu: done

frontend/src/utils/index.js
grolu: Yes, this would be possible. The method is in utils to make testing easier. It is used in a computed property anyway, so we could map and enrich the result there. If you want, I can change it but I thought it would be cleaner to have it all in one place.

holgerkoser (approved):
/lgtm

petersutter (approved):
/lgtm

PR #792, opened 2020-08-28
Set apiServerServiceRef in case of target shoot

petersutter:
If gardener's `APIServerSNI` feature gate was enabled by the gardener administrators, it was not possible to reach the kube-apiserver within the terminal pod when running a cluster terminal on the seed (operators only). This is now fixed by using the local endpoint of the kube-apiserver instead of the public endpoint.
The local endpoint `kube-apiserver` is used in case the terminal pod is running on the seed.
The local endpoint `kubernetes.default` is used in case the terminal pod is running on the shoot's worker node

Fixes #

Corresponding API change on `terminal-controller-manager` https://github.com/gardener/terminal-controller-manager/pull/40

**Release note**:

```action operator
This dashboard version requires terminal-controller-manager v0.12.0 or higher, in case the terminal feature is enabled
```

```improvement operator
If gardener's `APIServerSNI` feature gate was enabled by the gardener administrators, it was not possible to reach the kube-apiserver within the terminal pod when running a cluster terminal on the seed (operators only). This is now fixed by using the local endpoint of the kube-apiserver (`kube-apiserver` / `kubernetes.default`) instead of the public endpoint.
```

### DIFF
--- a/backend/lib/services/terminals/index.js
+++ b/backend/lib/services/terminals/index.js
@@ -198,7 +198,7 @@ function getShootResource ({ user, namespace, name, target }) {
   return client.getShoot({ namespace, name })
 }
 
-async function getTargetCluster ({ user, namespace, name, target, shootResource }) {
+async function getTargetCluster ({ user, namespace, name, target, preferredHost, shootResource }) {
   const client = user.client
   const isAdmin = user.isAdmin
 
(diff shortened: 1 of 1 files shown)


grolu (approved):
/lgtm

PR #797, opened 2020-09-09
Support seedSelector providerType and enhance project member check

petersutter:
This PR includes two improvements:
- `spec.seedSelector.providerTypes` on the cloudprofile is now supported
- is-member check on the project was enhanced to also consider the user's groups

Cloudprofiles need to have a matching seed, otherwise it is filtered out and not visible / selectable on the dashboard. The following describes how a cloudprofile is matched to a seed:
- **Provider type**
  - `spec.type` of the cloud profile has to match `spec.provider.type` of a seed
  - or one of  `spec.seedSelector.providerTypes` (array) of the cloud profile matches `spec.provider.type` of a seed
- **Optional: label selector**
  - in case `spec.seedSelector.matchLabels` is set of a cloud profile, there must be a seed with these `metadata.lables`

Fixes #

**Release note**:

```improvement user

```

### DIFF
--- a/backend/lib/services/cloudprofiles.js
+++ b/backend/lib/services/cloudprofiles.js
@@ -45,20 +45,22 @@ function emptyToUndefined (value) {
 
 function assignSeedsToCloudProfileIteratee (seeds) {
   return cloudProfileResource => {
+    function filterProviderType (seed) {
+      if (!_.isEmpty(providerTypes)) {
+        const seedProviderType = _.get(seed, 'spec.provider.type')
+        return _.includes(providerTypes, seedProviderType)
+      }
+
+      return ['spec.provider.type', providerType]
+    }
+
     const providerType = cloudProfileResource.spec.type
     const matchLabels = _.get(cloudProfileResource, 'spec.seedSelector.matchLabels', {})
     const providerTypes = _.get(cloudProfileResource, 'spec.seedSelector.providerTypes', [])
 
-    let chain = _.chain(seeds)
-    if (_.isEmpty(providerTypes)) {
-      chain = chain.filter(['spec.provider.type', providerType])
-    } else {
-      chain = chain.filter(seed => {
-        const seedProviderType = _.get(seed, 'spec.provider.type')
-        return _.includes(providerTypes, seedProviderType)
-      })
-    }
-    const seedsForCloudProfile = chain
+    const seedsForCloudProfile = _
+      .chain(seeds)
+      .filter(filterProviderType)
       .filter({ metadata: { labels: matchLabels } })
       .map(fromSeedResource)
       .thru(emptyToUndefined)
--- a/backend/lib/services/cloudprofiles.js
+++ b/backend/lib/services/cloudprofiles.js
@@ -47,10 +47,18 @@ function assignSeedsToCloudProfileIteratee (seeds) {
   return cloudProfileResource => {
     const providerType = cloudProfileResource.spec.type
     const matchLabels = _.get(cloudProfileResource, 'spec.seedSelector.matchLabels', {})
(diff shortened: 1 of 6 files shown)


backend/lib/services/cloudprofiles.js
```diff
@@ -47,10 +47,18 @@ function assignSeedsToCloudProfileIteratee (seeds) {
   return cloudProfileResource => {
     const providerType = cloudProfileResource.spec.type
     const matchLabels = _.get(cloudProfileResource, 'spec.seedSelector.matchLabels', {})
+    const providerTypes = _.get(cloudProfileResource, 'spec.seedSelector.providerTypes', [])
 
-    const seedsForCloudProfile = _
-      .chain(seeds)
-      .filter(['spec.provider.type', providerType])
+    let chain = _.chain(seeds)
+    if (_.isEmpty(providerTypes)) {
```
grolu, line 53: move if to filter function

petersutter: ✅

backend/lib/services/cloudprofiles.js
```diff
@@ -45,12 +45,24 @@ function emptyToUndefined (value) {
...
+      if (!_.isEmpty(providerTypes)) {
+        return seed => {
+          const seedProviderType = _.get(seed, 'spec.provider.type')
+          return _.includes(providerTypes, seedProviderType)
+        }
+      }
+
+      return ['spec.provider.type', providerType]
+    }
+
```
holgerkoser, line 58:
```suggestion
    function filterProviderType (seed) {
        const seedProviderType = _.get(seed, 'spec.provider.type')
        return _.includes(providerTypes, seedProviderType)
    }
```

backend/lib/services/cloudprofiles.js
```diff
@@ -45,12 +45,24 @@ function emptyToUndefined (value) {
...
     const providerType = cloudProfileResource.spec.type
     const matchLabels = _.get(cloudProfileResource, 'spec.seedSelector.matchLabels', {})
+    const providerTypes = _.get(cloudProfileResource, 'spec.seedSelector.providerTypes', [])
```
holgerkoser, line 61:
```suggestion
    const providerTypes = _.get(cloudProfileResource, 'spec.seedSelector.providerTypes', [cloudProfileResource.spec.type])
```

backend/lib/services/cloudprofiles.js
```diff
@@ -45,12 +45,24 @@ function emptyToUndefined (value) {
...
 
     const seedsForCloudProfile = _
       .chain(seeds)
-      .filter(['spec.provider.type', providerType])
+      .filter(filterProviderType())
```
holgerkoser, line 65:
```suggestion
      .filter(filterProviderType)
```

backend/lib/services/projects.js
```diff
@@ -138,6 +138,14 @@ exports.list = async function ({ user, qs = {} }) {
     .gte(0)
     .value()
 
+  const isGroupMemberOf = project => !_
```
holgerkoser, line 141: I would change the isMemberOf function in a way that it also considers groups. The user is member of the project if one of his groups is member of the project. 

```js
const predicate = members => {
  const hasGroupMembership = _
    .chain(members)
    .filter(['kind', 'Group'])
    .map('name')
    .intersection(user.groups)
    .size()
    .gt(0)
    .value()
  const hasUserMembership = _
    .chain(members)
    .filter(['kind', 'User'])
    .map('name')
    .includes(user.id)
    .value()
  return hasGroupMembership || hasUserMembership
}

holgerkoser (approved):
/lgtm

holgerkoser (approved):
/lgtm

PR #829, opened 2020-10-13
Fix storage plugin not returning item

### LINKED ISSUE #828: Sort preferences are no longer persisted
**What happened**:
Sort preferences are no longer persisted.

Sort preferences to be persisted as I don't want to sort them every time I open the dashboard.

1. Open a dashboard

2. Sort the Shoots by for example the infrastructure column

3. Refresh the dashboard and ensure that the sort preference is cleaned up.

**Environment**:

petersutter:
Fix storage plugin not returning item

Fixes #828 

**Release note**:

```improvement user

```

### DIFF
--- a/frontend/src/plugins/storage.js
+++ b/frontend/src/plugins/storage.js
@@ -23,7 +23,7 @@ const VueStorage = {
         window.localStorage.setItem(key, value)
       },
       getItem (key) {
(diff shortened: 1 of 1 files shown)


holgerkoser (approved):
/lgtm

PR #831, opened 2020-10-14
Handle unreachable seeds

### LINKED ISSUE #826: No error is shown in case of error when fetching the seed info
**What happened**:
No error is shown in case of error when fetching the seed info. The error was catched and logged in the backend.

It should be clear to the user that fetching the information (e.g. monitoring credentials) has failed

As operator, navigate on the shoot details page of a cluster that is using an infrastructure that is not reachable from the gardener dashboard

**Environment**:

petersutter:
This PR handles unreachable seeds
- Terminal feature is disabled in case the seed is not reachable by the dashboard
- A hint is shown that the monitoring credentials can't be fetched in case the seed is not reachable by the dashboard (only applies for operators)
Screenshot 2020-10-14 at 21 31 03

example `values.yaml`
```yaml
unreachableSeeds:
  matchLabels:
    seed.gardener.cloud/network: private
```

example seed:
```yaml
kind: Seed
metadata:
  labels:
    seed.gardener.cloud/network: private
    gardener.cloud/role: seed
  name: my-seed
spec:
  ...
```

Fixes #826 

**Release note**:

```improvement user
We are now handling the case when seed is not reachable by the dashboard backend
- Terminal feature is disabled
- A hint is shown that the monitoring credentials can't be fetched in case the seed is not reachable by the dashboard (only applies for operators)
```

```action operator
Seeds that are unreachable by the dashboard need to be flagged as such with a corresponding label. This label needs be manifested in the dashboard configuration under the property `unreachableSeeds.matchLabels` in the `values.yaml` file
```

### DIFF
--- a/frontend/src/components/UsernamePasswordListTile.vue
+++ b/frontend/src/components/UsernamePasswordListTile.vue
@@ -15,8 +15,8 @@ limitations under the License.
 -->
 
 <template>
-  <div v-show="((!!username || !!email) && !!password) || showNotAvailablePlaceholder">
-    <template v-if="(!!username || !!email) && !!password">
+  <div v-if="showCredentials || showNotAvailablePlaceholder">
+    <template v-if="showCredentials">
       <v-list-item v-if="username">
         <v-list-item-icon>
           <v-icon color="cyan darken-2">{{icon}}</v-icon>
@@ -136,6 +136,9 @@ export default {
     },
     visibilityIcon () {
       return this.showPassword ? 'mdi-eye-off' : 'mdi-eye'
+    },
+    showCredentials () {
+      return (!!this.username || !!this.email) && !!this.password
     }
   },
   watch: {
--- a/frontend/src/components/UsernamePasswordListTile.vue
+++ b/frontend/src/components/UsernamePasswordListTile.vue
@@ -15,50 +15,66 @@ limitations under the License.
 -->
 
 <template>
-  <div v-show="(!!username || !!email) && !!password">
-    <v-list-item v-if="username">
+  <div v-show="((!!username || !!email) && !!password) || showNotAvailablePlaceholder">
+    <template v-if="(!!username || !!email) && !!password">
+      <v-list-item v-if="username">
+        <v-list-item-icon>
+          <v-icon color="cyan darken-2">{{icon}}</v-icon>
+        </v-list-item-icon>
+        <v-list-item-content>
+          <v-list-item-subtitle>{{usernameTitle}}</v-list-item-subtitle>
+          <v-list-item-title class="pt-1">{{username}}</v-list-item-title>
+        </v-list-item-content>
+        <v-list-item-action>
+          <copy-btn :clipboard-text="username"></copy-btn>
+        </v-list-item-action>
+      </v-list-item>
+      <v-list-item v-if="email">
+        <v-list-item-icon>
+          <v-icon v-if="!username" color="cyan darken-2">{{icon}}</v-icon>
+        </v-list-item-icon>
+        <v-list-item-content class="pt-0">
+          <v-list-item-subtitle>{{emailTitle}}</v-list-item-subtitle>
+          <v-list-item-title class="pt-1">{{email}}</v-list-item-title>
+        </v-list-item-content>
+        <v-list-item-action>
+          <copy-btn :clipboard-text="email"></copy-btn>
+        </v-list-item-action>
+      </v-list-item>
+      <v-list-item>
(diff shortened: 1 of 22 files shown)


frontend/src/store/modules/seeds.js
```diff
@@ -0,0 +1,58 @@
...
+  all: []
+}
+
+// getters
+const getters = {
+  items: state => state.all,
+  seedByName: (state) => (name) => {
+    const predicate = item => item.metadata.name === name
+    return find(state.all, predicate)
+  }
```
holgerkoser, line 31:
```suggestion
  items (state) {
    return state.all
  },
  seedByName (state) {
    return name => find(state.all, ['metadata.name', name])
  }
```

frontend/src/store/modules/seeds.js
```diff
@@ -0,0 +1,58 @@
...
+
+// actions
+const actions = {
+  getAll: ({ commit, rootState }) => {
+    return getSeeds()
+      .then(res => {
+        commit('RECEIVE', res.data)
+        return state.all
+      })
+  }
```
holgerkoser, line 42:
```suggestion
  async getAll ({ commit, state }) {
    const { data } = await getSeeds()
    commit('RECEIVE', res.data)
    return state.all
  }
```

frontend/src/store/index.js
```diff
@@ -936,6 +957,12 @@ const actions = {
         dispatch('setError', err)
       })
   },
+  fetchSeeds ({ dispatch }) {
+    return dispatch('seeds/getAll')
+      .catch(err => {
+        dispatch('setError', err)
+      })
+  },
```
holgerkoser, line 965:
```suggestion
  async fetchSeeds ({ dispatch }) {
    try {
      await dispatch('seeds/getAll')
    } catch (err) {
      dispatch('setError', err)
    }
  },
```

frontend/src/store/index.js
```diff
@@ -571,15 +586,21 @@ const getters = {
...
+      const seeds = getters.seedsByNames(seedNames)
       const uniqueSeedRegions = uniq(map(seeds, 'data.region'))
       const uniqueSeedRegionsWithZones = filter(uniqueSeedRegions, isValidRegion(getters, cloudProfileName, cloudProfile.metadata.cloudProviderKind))
       return uniqueSeedRegionsWithZones
     }
   },
+  seedsByNames (state, getters) {
+    return (seedNames) => {
+      return map(seedNames, getters.seedByName)
+    }
```
holgerkoser, line 602:
```suggestion
    return seedNames => map(seedNames, getters.seedByName)
```

frontend/src/store/index.js
```diff
@@ -316,11 +317,25 @@ const getters = {
...
+  },
   cloudProfileByName (state, getters) {
     return (name) => {
       return getters['cloudProfiles/cloudProfileByName'](name)
     }
   },
+  seedByName (state, getters) {
+    return (name) => {
+      return getters['seeds/seedByName'](name)
+    }
```
holgerkoser, line 331:
```suggestion
    return name) => getters['seeds/seedByName'](name)
```

frontend/src/store/index.js
```diff
@@ -316,11 +317,25 @@ const getters = {
...
+  },
+  isSeedUnreachableByName (state, getters) {
+    return (name) => {
```
holgerkoser, line 334:
```suggestion
    return name => {
```

frontend/src/store/index.js
holgerkoser, line 331:
```suggestion
    return name => getters['seeds/seedByName'](name)
```

frontend/src/components/UsernamePasswordListTile.vue
```diff
@@ -15,50 +15,66 @@ limitations under the License.
 -->
 
 <template>
-  <div v-show="(!!username || !!email) && !!password">
-    <v-list-item v-if="username">
+  <div v-show="((!!username || !!email) && !!password) || showNotAvailablePlaceholder">
+    <template v-if="(!!username || !!email) && !!password">
```
holgerkoser, line 19: The code became to complex for html template from my point of view. We should create a computed property `showCredentials` or something like this.

```suggestion
  <div v-show="showCredentials || showNotAvailablePlaceholder">
    <template v-if="showCredentials">
```

petersutter: done

holgerkoser (approved):
/lgtm

PR #838, opened 2020-10-22
Import more Roboto font variants ...

holgerkoser:
1. This PR imports more [Roboto](https://github.com/fontsource/fontsource/tree/master/packages/roboto#readme) font variants with the full CSS unicode-range.
* `normal`:  [100,300,400,500,700,900]
* `italic`: [400]

2. If the the request to `/config.json` fails during bootstrap the frontend was stuck in an endlessloop. This is also fixed with this PR.

Fixes #

**Release note**:

```improvement user

```

### DIFF
--- a/frontend/src/plugins/styles.js
+++ b/frontend/src/plugins/styles.js
@@ -14,12 +14,18 @@
 // limitations under the License.
 //
 
-import './styles'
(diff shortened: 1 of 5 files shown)


petersutter (approved):
/lgtm

PR #839, opened 2020-10-22
Fix terminal runtime selection

petersutter:
- As operator, when changing the terminal settings and selecting `Cluster` as runtime for the terminal pod and afterwards try to change the runtime to `Infrastructure (Seed)` it wouldn't work. This is now fixed.
- Also the behavior was changed when closing snackbar (that is displayed on connection loss / connection error). It will also close the terminal window

Fixes #

**Release note**:

```improvement operator
Fixed issue that when changing the terminal settings and selecting `Cluster` as runtime for the terminal pod and afterwards try to change the runtime to `Infrastructure (Seed)` it wouldn't work
```

### DIFF
--- a/frontend/src/components/GTerminal.vue
+++ b/frontend/src/components/GTerminal.vue
@@ -39,10 +39,10 @@ limitations under the License.
           </v-btn>
         </template>
       </g-popper>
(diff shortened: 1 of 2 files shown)


holgerkoser (approved):
/lgtm

PR #843, opened 2020-10-23
Fix unreachable seed check

petersutter:
Fixed issue that seed is considered as unreachable if the `unreachableSeeds.matchLabels` configuration was missing

Fixes #

Only an issue on the master branch (bug is not on released version)

**Release note**:

```improvement user

```


holgerkoser (approved):
\lgtm

PR #846, opened 2020-10-26
Fix multiline strings in chart

petersutter:
In case e.g. `frontendConfig.ticket.issueDescriptionTemplate` contains multiline strings with two or more consecutive newlines, it could be that `data["config.yaml"]` of the `gardener-dashboard-configmap` will be treated as one large json string which makes it hard to read and edit the content. 

We are now using toJson for multiline strings

Fixes #

**Release note**:

```improvement user

```


grolu (approved):
/lgtm

PR #847, opened 2020-10-26
Confirm dialog using "enter" key press

petersutter:
Do not call non existing function. `okClicked` function does not exist anymore
Instead, resolve the action in case it is valid. This should restore the previous behaviour of the component.

Fixes #

**Release note**:

```improvement user
Fixed: Pressing enter on the confirmation dialog now behaves the same as if the OK button is pressed
```

### DIFF
--- a/frontend/src/components/dialogs/GDialog.vue
+++ b/frontend/src/components/dialogs/GDialog.vue
@@ -37,7 +37,7 @@ limitations under the License.
       <v-alert tile :color="confirmAlertColor" v-if="confirmValue && !confirmDisabled">
         <span class="text-body-2" v-if="!!confirmMessage">{{confirmMessage}}</span>
         <v-text-field
-          @keyup.enter="okClicked()"
+          @keyup.enter="resolveAction(true)"
           ref="deleteDialogInput"
           :hint="hint"
           persistent-hint
@@ -205,6 +205,10 @@ export default {
       }
     },
     async resolveAction (value) {
+      if (value && !this.valid) {
+        return
+      }
+
       if (isFunction(this.resolve)) {
         if (value) {
           if (this.confirmationInterceptor) {


holgerkoser (approved):
/lgtm

frontend/src/components/dialogs/GDialog.vue
```diff
@@ -37,7 +37,6 @@ limitations under the License.
       <v-alert tile :color="confirmAlertColor" v-if="confirmValue && !confirmDisabled">
         <span class="text-body-2" v-if="!!confirmMessage">{{confirmMessage}}</span>
         <v-text-field
```
grolu, line 39: I saw this error a couple of times already in the log. Do you know why we had this here and did this even work at some time?

petersutter: Yes, initially we had a ConfirmInputDialog with a button
`<v-btn @click.native.stop="okClicked()" :disabled="confirmDisabled || hasError" class="red--text" flat>Confirm</v-btn>`

and having an ok and cancel function passed as prop.
```js
    props: {
      value: {
        type: Boolean,
        required: true
      },
      ok: {
        type: Function,
        required: true
      },
      cancel: {
        type: Function,
        required: true
      },
      ..
```
which was called in the okClicked / cancelClicked function
```js
    methods: {
      cancelClicked () {
        if (this.cancel) {
          this.cancel()
        }
      },
      okClicked () {
        if (this.ok) {
          this.ok()
        }
      },
```

at some point in time we agreed that passing a function as prop is not a good idea, hence it was removed..

grolu: ah okay. So now the component that is using the dialog must take over this functionality (if required) by listening to the key events? Then I guess we can approve this. We should however align the general behavior of dialogs. I think some of them can be confirmed via keys while others do not allow this.

petersutter: Good point. I now resolve the action if it is valid. This should restore the old behavior
For other dialogs, we should discuss. E.g. for the create cluster dialog I do not think it's a good idea to "create" on enter press

holgerkoser: I should have noticed this already in the first review.

holgerkoser (approved):
/lgtm

PR #860, opened 2020-11-04
Use secret binding name as infrastructure secret name

### LINKED ISSUE #859: Duplicate names for infrastructure secrets
**What happened**:
On the secrets page two infrastructure secrets with the same name are shown

The infrastructure secrets should have different names

Create two secret bindings and let them point (via secretRef) to the same secret

**Environment**:

petersutter:
Use secret binding name as infrastructure secret name

Fixes #859 

**Release note**:

```improvement user

```


holgerkoser (approved):
/lgtm

PR #862, opened 2020-11-04
Disallow the use of console

holgerkoser:
This PR adds an `eslint` rule which disallows the use of console for all packages. The only exception is the frontend the usage of `console.error` is still allowed.

Fixes #

**Release note**:

```improvement user

```


petersutter (approved):
/lgtm

PR #868, opened 2020-11-05
Use the Gardener logo instead of the SAP logo in the footer

berendt:
Perhaps this needs to be discussed in advance. But I think it makes sense not to have the SAP logo in the footer.


gardener-robot-ci-1:
Thank you @berendt for your contribution. Before I can start building your PR, a member of the organization must set the required label(s) {'reviewed/ok-to-test'}. Once started, you can check the build status in the PR checks section below.

petersutter:
Hi @berendt, currently we have `this.cfg.footerLogoUrl` that tries to read the footer logo url from the config, however we haven't included this configuration option in the gardener-dashboard helm chart. Hence this only works when patching the configmap and the volume mounts of the dashboard deployment afterwards.
Also there would be two gardener logos on the page which does not look good.
For now, let's remove the footer logo completely.

When working at the theming umbrella issue #875, respectively #874 we can work on making the logos configurable and support it also in the helm chart.

petersutter:
> For now, let's remove the footer logo completely.
>
> When working at the theming umbrella issue #875, respectively #874 we can work on making the logos configurable and support it also in the helm chart.

Hi @berendt, with #901 I have completely removed the footer logo.
Will close this PR in favour of #901. Please let me know if you have any concerns

PR #872, opened 2020-11-06
Member Management Part2

### LINKED ISSUE #788: Member Management Improvements
The current implementation of the project member management page has some limitations that we know about. Now with the requirement to connect service accounts from different namespaces, these limitations need to be resolved.

Screen Shot 2020-08-26 at 17 53 12

# Planned changes

## Project User Card
- [x]  Change the _trash_ button to _remove_ (as no actual deletion takes place this may causes confusion)
- [x]  Prevent to add service accounts via the add user dialog (prevent names that start with _system:serviceaccount_), the reason for this is that it may confuses users when they add a service account using this dialog and the connected account appears on a different card
- [x]  Show owner role (grayed out, not editable, with tooltip)

## Service Accounts Card
Currently, the name of tis card is a bit misleading as we actually do not show (all) service accounts. We only show service accounts that have a role in the project, more precisely service accounts that have en entry of kind _User_ in the member section of the _project_ resource. This can lead to name clashes even if no service account with the name is listed.

- [x]  Show all service accounts, even those that have no role in the project
- [x]  Possibility to add service accounts without a role
- [x]  Deletion of service accounts need to be confirmed by entering the name
- [x]  Possibility to connect service accounts from different namespaces by using the full qualified name. Those service accounts need to have a role and cannot be deleted, so they also get a _remove_ button
- [x]  Add service accounts to project resource with kind _ServiceAccount_
- [x]  Add possibility to rotate service account secrets
- [x] Show service account details in row
- [x] Support descriptions for Service Accounts (annotation)

## Other Changes
- [x]  Add information about role types to help dialog
- [x]  Rename _Technical Contact_ to owner on Administration page
- [x]  Remove _Technical Contact_ Card

grolu:
This PR adds the missing functionality for the member management improvements

Fixes #788 

**Release note**:

```feature user
Member Management Improvements
- Added feature that allows to rotate secrets of service accounts
- Members are now sortable and shown in table layout
- Added help texts for member roles
- Service Account Members whose Service Account has been deleted, are now flagged in the Dashboard and can be cleaned up
```


backend/lib/services/members/MemberManager.js
```diff
@@ -97,6 +97,21 @@ class MemberManager {
...
   }
 
+  async deleteSecret (id) {
+    const item = this.subjectList.get(id)
+    if (!item) {
+      return this.subjectList.members
+    }
+
+    if (!item.kind === 'ServiceAccount') {
+      throw new MethodNotAllowed('Member is not a ServiceAccount')
```
holgerkoser, line 107:
```suggestion
      throw new UnprocessableEntity('Member is not a ServiceAccount')
```

backend/lib/routes/members.js
```diff
@@ -66,3 +66,15 @@ router.route('/:name')
       next(err)
     }
   })
+
+router.route('/:name/secret')
+  .delete(async (req, res, next) => {
+    try {
+      const user = req.user
+      const namespace = req.params.namespace
+      const name = req.params.name
```
holgerkoser, line 75:
```suggestion
      const { namespace, name } = req.params
```

backend/lib/routes/members.js
```diff
@@ -66,3 +66,15 @@ router.route('/:name')
       next(err)
     }
   })
+
+router.route('/:name/secret')
```
holgerkoser, line 70: there is no route get secret. I would use verb `post` on the member route.

backend/lib/services/members/MemberManager.js
```diff
@@ -97,6 +97,21 @@ class MemberManager {
     return this.subjectList.members
   }
 
+  async deleteSecret (id) {
+    const item = this.subjectList.get(id)
+    if (!item) {
+      return this.subjectList.members
```
holgerkoser, line 103: We do not modify the member list. I would not return a body?

backend/lib/services/members/MemberManager.js
```diff
@@ -97,6 +97,21 @@ class MemberManager {
...
+    }
+
+    if (!item.kind === 'ServiceAccount') {
+      throw new MethodNotAllowed('Member is not a ServiceAccount')
+    }
+
+    await this.deleteServiceAccountSecret(item)
+
+    return this.subjectList.members
```
holgerkoser, line 112: We do not modify the member list. I would not return a body?

backend/lib/services/members/MemberManager.js
```diff
@@ -97,6 +97,21 @@ class MemberManager {
     return this.subjectList.members
   }
 
+  async deleteSecret (id) {
```
holgerkoser, line 100:
```suggestion
  async rotateServiceAccountSecret (id) {
```

backend/lib/services/members/MemberManager.js
```diff
@@ -169,6 +184,21 @@ class MemberManager {
     this.subjectList.delete(item.id)
   }
 
+  async deleteServiceAccountSecret (item) {
+    const { namespace } = Member.parseUsername(item.id)
+    if (namespace !== this.namespace) {
+      return
```
holgerkoser, line 190: This should be an error I think?!

backend/lib/services/members/MemberManager.js
```diff
@@ -169,6 +184,21 @@ class MemberManager {
...
+    }
+
+    const secretName = _
+      .chain(item)
+      .get('extensions.secrets')
+      .head()
```
holgerkoser, line 196: What if we have more than one secret?

backend/lib/services/members/MemberManager.js
```diff
@@ -169,6 +184,21 @@ class MemberManager {
...
+      .get('name')
+      .value()
+    return await this.client.core.secrets.delete(namespace, secretName)
```
holgerkoser, line 199:
```suggestion
    const name = _
      .chain(item)
      .get('extensions.secrets')
      .head()
      .get('name')
      .value()
    return await this.client.core.secrets.delete(namespace, name)
```

frontend/src/components/ProjectServiceAccountRow.vue
```diff
@@ -129,41 +122,11 @@ export default {
...
-    created: {
-      type: String
-    },
-    roles: {
-      type: Array,
-      required: true
-    },
-    roleDisplayNames: {
-      type: Array,
+    item: {
```
holgerkoser, line 129: Why do we change the explicit behavior to an implicit untyped?

frontend/src/store/modules/members.js
```diff
@@ -41,6 +41,11 @@ const actions = {
     const namespace = rootState.namespace
     const res = await deleteMember({ namespace, name })
     commit('RECEIVE', res.data)
+  },
+  async rotateServiceAccountSecret ({ commit, rootState }, name) {
+    const namespace = rootState.namespace
+    const res = await deleteServiceAccountSecret({ namespace, name })
+    commit('RECEIVE', res.data)
```
holgerkoser, line 48: We do not change the list

frontend/src/views/Members.vue
```diff
@@ -43,24 +43,22 @@ SPDX-License-Identifier: Apache-2.0
           Access to resources within your project can be configured by assigning roles.
         </p>
       </v-card-text>
-      <v-list two-line subheader v-else>
-        <template v-for="(user, index) in sortedAndFilteredUserList">
-          <v-divider v-if="index !== 0" inset :key="index"></v-divider>
+      <v-data-table
```
holgerkoser, line 46: not sortable, no pagination, no filter ==> <v-simple-table>?

frontend/src/views/Members.vue
```diff
@@ -321,7 +374,9 @@ export default {
     ...mapActions([
       'addMember',
       'deleteMember',
-      'setError'
+      'rotateServiceAccountSecret',
+      'setError',
+      'set'
```
holgerkoser, line 379: There is no action `set` on the store!

backend/lib/services/members/MemberManager.js
grolu: You are right. I think I just did it as we always did in the past. I will have another look.

backend/lib/services/members/MemberManager.js
grolu: Yes, we can throw an error here. Also 422 in this case?

backend/lib/services/members/MemberManager.js
grolu: Hmm, well I asked myself the same question. The way we generate kubeconfigs is that we take the first secret so I guessed, deleting the first one will do the job. I don't know how we should handle service accounts with multiple secrets. We could of course delete all, but I don't know if this what we want. WDYT?

frontend/src/components/ProjectServiceAccountRow.vue
grolu: @petersutter refactored the code and added this component for reuse in the terminal feature context. However, he missed to re-use the component in the user context but it is the very same code so re-use here makes 100% sense. However, the name `ServiceAccount` makes no sense if we use it ofr all account types (User and Service Accounts)

frontend/src/views/Members.vue
grolu: IDK where this came from. rm

frontend/src/store/modules/members.js
grolu: I removed the commit, you are right ... same as above

frontend/src/views/Members.vue
grolu: Filter is in place, sorting can be added if we want but I currently do not see a real use case for it... maybe for the roles but I don't know... I will add pagination and adjust the row spacing and alignment, as we discussed with @petersutter

backend/lib/services/members/MemberManager.js
grolu: we can throw an error if there is more than one secret (422)

backend/lib/routes/members.js
```diff
@@ -56,24 +56,27 @@ router.route('/:name')
...
-      const name = req.params.name
-      res.send(await members.remove({ user, namespace, name }))
+      const { namespace, name } = req.params
+      const { method } = req.body
+
+      if (method === 'rotateSecret') {
+        res.send(await members.rotateSecret({ user, namespace, name }))
+      } else {
+        throw new Error(`${method} not allowed for members`)
+      }
```
holgerkoser, line 69:
```suggestion
      switch (method) {
        case 'rotateSecret':
          res.send(await members.rotateSecret({ user, namespace, name }))
          break
        default:
          throw new UnprocessableEntity(`${method} not allowed for members`)
      }
```

backend/lib/services/members/MemberManager.js
```diff
@@ -159,7 +157,7 @@ class MemberManager {
   async updateServiceAccount (item, { description }) {
     const { namespace, name } = Member.parseUsername(item.id)
     if (namespace !== this.namespace) {
-      return
+      throw new UnprocessableEntity(`ServiceAccount namespace ${namespace} does not match project namespace this.namespace`)
```
holgerkoser, line 160:
```suggestion
      throw new UnprocessableEntity('It is not possible to modify ServiceAccount from another namespace')
```

backend/lib/services/members/MemberManager.js
```diff
@@ -187,16 +185,20 @@ class MemberManager {
   async deleteServiceAccountSecret (item) {
     const { namespace } = Member.parseUsername(item.id)
     if (namespace !== this.namespace) {
-      return
+      throw new UnprocessableEntity(`ServiceAccount namespace ${namespace} does not match project namespace this.namespace`)
```
holgerkoser, line 188:
```suggestion
      throw new UnprocessableEntity(`ServiceAccount namespace ${namespace} does not match project namespace this.namespace`)
```
```suggestion
   throw new UnprocessableEntity('It is not possible to modify ServiceAccount from another namespace')
```

backend/lib/services/members/MemberManager.js
```diff
@@ -187,16 +185,20 @@ class MemberManager {
...
     }
 
-    const secretName = _
+    if (_.get(item, 'extensions.secrets.length') > 1) {
```
holgerkoser, line 191:
```suggestion
    if (_.size(item, 'extensions.secrets') > 1) {
```

backend/lib/services/members/MemberManager.js
```diff
@@ -187,16 +185,20 @@ class MemberManager {
...
+      throw new UnprocessableEntity(`ServiceAccount ${namespace} has more than one secret`)
```
holgerkoser, line 192: I am not sure if this should be an client error

backend/test/services.members.spec.js
```diff
@@ -393,6 +412,15 @@ describe('services', function () {
         })
       })
 
+      describe('#deleteServiceAccount', function () {
+        it('should delete a serviceaccount', async function () {
+          const id = 'system:serviceaccount:garden-foo:robot-sa'
+          const item = memberManager.subjectList.get(id)
+          await memberManager.deleteServiceAccount(item)
+          expect(client.core.serviceaccounts.delete).to.have.been.calledOnce
```
holgerkoser, line 420: also check arguments

backend/test/services.members.spec.js
```diff
@@ -403,11 +431,50 @@ describe('services', function () {
...
+          }
+        })
+      })
+
+      describe('#deleteServiceAccountSecret', function () {
+        it('should delete a serviceaccount secret', async function () {
+          const id = 'system:serviceaccount:garden-foo:robot-sa'
+          const item = memberManager.subjectList.get(id)
+          await memberManager.deleteServiceAccountSecret(item)
+          expect(client.core.secrets.delete).to.have.been.calledOnce
```
holgerkoser, line 451: check arguments

backend/test/services.members.spec.js
```diff
@@ -403,11 +431,50 @@ describe('services', function () {
...
       describe('#updateServiceAccount ', function () {
-        it('should not delete a serviceaccount from a different namespace', async function () {
+        it('should not update a serviceaccount from a different namespace', async function () {
+          const id = 'system:serviceaccount:garden-foreign:robot-foreign-namespace'
+          const item = memberManager.subjectList.get(id)
+          try {
+            await memberManager.updateServiceAccount(item, {})
+            expect.fail('should throw an error')
+          } catch (err) {
+            expect(err).to.be.instanceof(UnprocessableEntity)
```
holgerkoser, line 441: We have many UnprocessableEntity errors check significant part of message with regex

frontend/src/components/ProjectServiceAccountRow.vue
```diff
@@ -5,121 +5,114 @@ SPDX-License-Identifier: Apache-2.0
...
+
+          <span class="body-2">{{item.username}}</span>
+        </div>
+      </div>
+    </td>
+    <td>
+      <account-avatar :account-name="item.createdBy" :size="16"></account-avatar>
+    </td>
+    <td>
+      <div>
```
holgerkoser, line 33: Why this div

petersutter:
- The edit button of service accounts has the tooltip "Change Service Account Roles", even though you can edit the description there
- Should we use the `EditableText` component do edit the description in place (in the table)?

frontend/src/components/ProjectServiceAccountRow.vue
grolu: switched back to v-list

backend/lib/services/members/MemberManager.js
```diff
@@ -97,6 +97,19 @@ class MemberManager {
     return this.subjectList.members
   }
 
+  async rotateServiceAccountSecret (id) {
+    const item = this.subjectList.get(id)
+    if (!item) {
+      return
+    }
+
+    if (!item.kind === 'ServiceAccount') {
```
holgerkoser, line 106: Very strange kind of boolean algebra
```suggestion
    if (item.kind !== 'ServiceAccount') {
```

holgerkoser: Don't we have a unit test for this case?

backend/lib/services/members/MemberManager.js
```diff
@@ -162,13 +175,32 @@ class MemberManager {
...
+      throw new UnprocessableEntity('It is not possible to modify a ServiceAccount from another namespace')
+    }
+
+    if (_.size(_.get(item, 'extensions.secrets')) > 1) {
+      throw new UnprocessableEntity(`ServiceAccount ${namespace} has more than one secret`)
+    }
+
+    const name = _
+      .chain(item)
+      .get('extensions.secrets')
```
holgerkoser, line 197:
```suggestion
      .get('extensions.secrets', [])
      .tap(secrets => {
        if (secrets.length > 1) {
          throw new UnprocessableEntity(`ServiceAccount ${namespace} has more than one secret`)
        }
      })
```

backend/lib/services/members/MemberManager.js
```diff
@@ -162,13 +175,32 @@ class MemberManager {
...
 
+  async deleteServiceAccountSecret (item) {
+    const { namespace } = Member.parseUsername(item.id)
+    if (namespace !== this.namespace) {
+      throw new UnprocessableEntity('It is not possible to modify a ServiceAccount from another namespace')
+    }
+
+    if (_.size(_.get(item, 'extensions.secrets')) > 1) {
+      throw new UnprocessableEntity(`ServiceAccount ${namespace} has more than one secret`)
+    }
```
holgerkoser, line 193:
```suggestion
```

frontend/src/components/ProjectServiceAccountRow.vue
```diff
@@ -5,164 +5,134 @@ SPDX-License-Identifier: Apache-2.0
...
+              </v-btn>
+            </template>
+            <span>Change Service Account</span>
+          </v-tooltip>
+        </div>
+        <div v-if="canManageServiceAccountMembers" class="ml-1">
+          <v-tooltip top>
+            <template v-slot:activator="{ on }">
+              <v-btn v-on="on" icon color="red" @click.native.stop="onDelete">
+                <v-icon v-if="isServiceAccountFromCurrentNamespace">mdi-delete</v-icon>
```
holgerkoser, line 99: Why not a computed property?

frontend/src/components/ProjectServiceAccountRow.vue
```diff
@@ -5,164 +5,134 @@ SPDX-License-Identifier: Apache-2.0
...
+                <v-icon v-else>mdi-close</v-icon>
+              </v-btn>
+            </template>
+            <span v-if="isServiceAccountFromCurrentNamespace">Delete Service Account</span>
+            <span v-else>Remove Foreign Service Account from Project</span>
```
holgerkoser, line 104: Why not a computed property?

frontend/src/components/ProjectServiceAccountRow.vue
```diff
@@ -173,41 +143,47 @@ export default {
...
       return namespace
+    },
+    selectedHeaders () {
+      return mapTableHeader(this.headers, 'selected')
     }
   },
   methods: {
     onDownload () {
-      this.$emit('download', this.username)
+      this.$emit('download', this.item.username)
```
holgerkoser, line 165: Always emit the item

frontend/src/components/ProjectUserRow.vue
```diff
@@ -5,109 +5,88 @@ SPDX-License-Identifier: Apache-2.0
...
+    </td>
+    <td width="250px" v-if="selectedHeaders.actions">
+      <div class="d-flex flex-row justify-end">
+        <div v-if="canManageMembers" class="ml-1">
+          <v-tooltip top>
             <template v-slot:activator="{ on }">
-              <v-chip v-on="on" class="mr-3" small :color="notEditable ? 'grey' : 'black'" outlined>
-                {{displayName}}
-              </v-chip>
+              <v-btn v-on="on" icon @click.native.stop="onEdit">
```
holgerkoser, line 31:
```suggestion
              <v-btn v-on="on" icon @click.stop="onEdit">
```

backend/lib/services/members/MemberManager.js
grolu: we have a test and if the expression above is wrong the test fails

frontend/src/components/ProjectUserRow.vue
```diff
@@ -5,109 +5,88 @@ SPDX-License-Identifier: Apache-2.0
...
-    ])
+    ]),
+    selectedHeaders () {
+      return mapTableHeader(this.headers, 'selected')
+    }
   },
   methods: {
     onDelete (username) {
-      this.$emit('delete', this.username)
+      this.$emit('delete', this.item.username)
```
holgerkoser, line 86: emit item

frontend/src/components/TableColumnSelection.vue
```diff
@@ -0,0 +1,129 @@
...
+    }
+  },
+  data () {
+    return {
+      columnSelectionMenu: false
+    }
+  },
+  methods: {
+    setSelectedColumn (header) {
+      this.$emit('setSelectedColumn', header)
```
holgerkoser, line 105: The names of the events should not be the name of method that the listener will have to call. It should be something like `selectionChange` or `reset` or `filterChange`.

frontend/src/views/Members.vue
```diff
@@ -43,24 +48,26 @@ SPDX-License-Identifier: Apache-2.0
...
       </v-card-text>
-      <v-list two-line subheader v-else>
-        <template v-for="(user, index) in sortedAndFilteredUserList">
-          <v-divider v-if="index !== 0" inset :key="index"></v-divider>
+      <v-data-table
+        v-else
+        :headers="visibleUserAccountTableHeaders"
+        :items="userList"
+        :footer-props="{ 'items-per-page-options': [5,10,20] }"
+        :options.sync="userAccountTableOptions"
```
holgerkoser, line 56: Here we use a sync property to distribute shared state across components. I think we should only use one possible solution for one type of problem. Why don't we share the state via the store. Even for reuse components we could have a path which directs the component to the correct store.state property. Also the localStorage handling should be done centrally in the store.

backend/lib/routes/members.js
grolu: done

frontend/src/components/ProjectUserRow.vue
holgerkoser: username parameter always empty ==> delete

frontend/src/components/ProjectServiceAccountRow.vue
grolu: optimized to not use templating conditions

frontend/src/components/ProjectServiceAccountRow.vue
grolu: optimized to not use templating conditions

frontend/src/components/ProjectServiceAccountRow.vue
grolu: done

frontend/src/components/ProjectUserRow.vue
grolu: done

frontend/src/components/TableColumnSelection.vue
grolu: renamed as suggested

grolu: we agreed to do this in a separate PR, see also https://github.com/gardener/dashboard/issues/905

frontend/src/views/Members.vue
grolu: we agreed to do this in a separate PR, see also https://github.com/gardener/dashboard/issues/905

backend/lib/services/members/MemberManager.js
```diff
@@ -162,13 +175,33 @@ class MemberManager {
   async deleteServiceAccount (item) {
     const { namespace, name } = Member.parseUsername(item.id)
     if (namespace !== this.namespace) {
-      return
+      return // do not throw an error as it is allowed to add foreign service accounts
```
petersutter, line 178: is it really allowed to delete a service account from another namespace?
if yes, the comment needs to be adapted anyhow: `add` -> `delete`

frontend/src/views/ShootList.vue
```diff
@@ -26,125 +26,13 @@ SPDX-License-Identifier: Apache-2.0
...
-                      <div v-for="label in hideClustersWithLabels" :key="label">- {{label}}</div>
-                    </v-tooltip>
-                    ticket labels
-                  </v-list-item-title>
-                </v-list-item-content>
-              </v-list-item>
-            </template>
-          </v-list>
-        </v-menu>
+        <table-column-selection
```
petersutter, line 29: bug: I cannot select any columns under the table options menu

frontend/src/components/ProjectServiceAccountRow.vue
```diff
@@ -5,164 +5,132 @@ SPDX-License-Identifier: Apache-2.0
...
+          </v-tooltip>
+        </div>
+        <div v-if="canManageServiceAccountMembers" class="ml-1">
+          <v-tooltip top>
+            <template v-slot:activator="{ on }">
+              <v-btn v-on="on" icon @click.native.stop="onEdit">
+                <v-icon>mdi-pencil</v-icon>
+              </v-btn>
+            </template>
+            <span>Change Service Account</span>
```
petersutter, line 92: When I change the role of a service account and press `Update` it takes very long until the dialog closes. Did anything change here? At first I thought it didn't react on my button press but then after ~3 seconds the dialog closed

petersutter (changes requested):
there are some bugs that needs to be fixed, and see my other comments

frontend/src/views/Members.vue
```diff
@@ -454,12 +552,23 @@ export default {
         confirmValue: name
       })
     },
-    onEditUser (username, roles) {
+    confirmRotateServiceAccountSecret (name) {
+      const memberName = escape(displayName(name))
+      const messageHtml = `Do you want to rotate the secret of service account <i>${memberName}</i>?<br /> Attention: The current kubeconfig credentials will be revoked`
```
petersutter, line 557: change to vue templates

backend/lib/services/members/MemberManager.js
grolu: renamed comment to make it clearer

frontend/src/components/ProjectServiceAccountRow.vue
grolu: I did not change the way this works. Probably some network / server load issue. But there is certainly room for improvement when handling these async actions on the UI, e.g. disabling the button, adding a spinner...
Probably not scope of this PR though

frontend/src/views/Members.vue
grolu: done

frontend/src/views/ShootList.vue
grolu: not reducable

petersutter (approved):
/lgtm

backend/lib/routes/members.js
```diff
@@ -56,6 +57,23 @@ router.route('/:name')
...
   })
+  .post(async (req, res, next) => {
+    try {
+      const user = req.user
+      const { namespace, name } = req.params
+      const { method } = req.body
+
+      switch (method) {
+        case 'rotateSecret':
+          res.send(await members.rotateSecret({ user, namespace, name }))
```
holgerkoser, line 68:
```suggestion
          await members.rotateSecret({ user, namespace, name })
          res.end()
```

backend/lib/services/members/MemberManager.js
```diff
@@ -162,13 +175,33 @@ class MemberManager {
...
+      .get('extensions.secrets', [])
+      .tap(secrets => {
+        if (secrets.length > 1) {
+          throw new UnprocessableEntity(`ServiceAccount ${namespace} has more than one secret`)
+        }
+      })
+      .head()
+      .get('name')
+      .value()
+    return await this.client.core.secrets.delete(namespace, name)
```
holgerkoser, line 202: Remove return or await

backend/lib/services/members/index.js
```diff
@@ -32,3 +32,8 @@ exports.remove = async function ({ user, namespace, name }) {
   const memberManager = await MemberManager.create(user, namespace)
   return memberManager.delete(name)
 }
+
+exports.rotateSecret = async function ({ user, namespace, name }) {
+  const memberManager = await MemberManager.create(user, namespace)
+  memberManager.rotateServiceAccountSecret(name)
```
holgerkoser, line 38:
```suggestion
  await memberManager.rotateServiceAccountSecret(name)
```

backend/test/acceptance/api.members.spec.js
```diff
@@ -393,5 +393,25 @@ describe('api', function () {
...
+      mockRequest.mockImplementationOnce(fixtures.projects.mocks.get())
+      mockRequest.mockImplementationOnce(fixtures.serviceaccounts.mocks.list())
+      mockRequest.mockImplementationOnce(fixtures.serviceaccounts.mocks.delete())
+
+      const res = await agent
+        .post(`/api/namespaces/${namespace}/members/${name}`)
+        .set('cookie', await user.cookie)
+        .send({
+          method: 'rotateSecret'
+        })
```
holgerkoser, line 409:
```suggestion
        })
        .expect(200)
```

backend/test/acceptance/api.members.spec.js
```diff
@@ -393,5 +393,25 @@ describe('api', function () {
 
       expect(res.body).toMatchSnapshot()
     })
+
+    it('should rotate a service account secret', async function () {
+      const name = 'system:serviceaccount:garden-foo:robot'
+
+      mockRequest.mockImplementationOnce(fixtures.projects.mocks.get())
+      mockRequest.mockImplementationOnce(fixtures.serviceaccounts.mocks.list())
+      mockRequest.mockImplementationOnce(fixtures.serviceaccounts.mocks.delete())
```
holgerkoser, line 402:
```suggestion
      mockRequest.mockImplementationOnce(fixtures.secrets.mocks.delete())
```

backend/test/acceptance/api.members.spec.js
```diff
@@ -393,5 +393,25 @@ describe('api', function () {
...
+        .post(`/api/namespaces/${namespace}/members/${name}`)
+        .set('cookie', await user.cookie)
+        .send({
+          method: 'rotateSecret'
+        })
+
+      expect(mockRequest).toBeCalledTimes(3)
+      expect(mockRequest.mock.calls).toMatchSnapshot()
+
+      expect(res.body).toMatchSnapshot()
```
holgerkoser, line 414:
```suggestion
```

holgerkoser (approved):
/lgtm

holgerkoser (approved):
/lgtm

PR #880, opened 2020-11-10
Show error page if no permission to list secrets

### LINKED ISSUE #870: No error message is shown if you have no permission for secrets page
**What happened**:
With this fix another issue surfaced where the secrets page is shown after switching even though you have viewer permission to view them.
How to reproduce:
You are member of two projects.
In project A you have admin role
In project B you have viewer role

Navigate to the secrets page of project A.
Now switch to project B
You are still on the secrets page but you only have the viewer role. Of course no secrets are shown but you also do not see an error message

You see an error page when trying to navigate to `/namespace/garden-myproject/secrets` with viewer permission
When following the above procedure, after switching to project B it is expected to be redirected to a fallback page, like the cluster list

**Environment**:

holgerkoser:
With this PR you will see an error page when trying to navigate to `/namespace/garden-<project>/secrets` with only viewer permission.

Forbidden

Fixes #870 

**Release note**:

```improvement user

```


frontend/src/views/ProjectError.vue
```diff
@@ -0,0 +1,69 @@
...
+          params: {
+            namespace
+          }
+        }
+      }
+      return {
+        name: 'Home'
+      }
+    }
+  },
```
petersutter, line 58: this does not work anymore:
- login e.g. with service account token that is not a member of any project
- navigate to `/namespace/garden-notexisting/shoots/notexisting`
- Clicking the `Get me out of here` button will not bring you to the `Home` screen

petersutter (changes requested):

frontend/src/views/ProjectError.vue
holgerkoser: ✔️

petersutter (approved):
/lgtm

grolu (approved):
/lgtm

PR #881, opened 2020-11-11
ControllerRegistration support / List Networking Types

grolu:
Added support to read controllerregistration resources
Show dropdown for registered network extensions
Show network extension list (for operators)

Fixes #

**Release note**:
```user feature
Available Networking Extensions are listed and can be selected during cluster creation
```


backend/test/acceptance/api.controllerregistrations.spec.js
```diff
@@ -0,0 +1,31 @@
...
+    common.stub.getControllerRegistrations(sandbox)
+
+    const res = await agent
+      .get('/api/controllerregistrations')
+      .set('cookie', await user.cookie)
+
+    expect(res).to.have.status(200)
+    expect(res).to.be.json
+    expect(res.body).to.have.length(2)
+    expect(res.body[0].name).eql('foo')
```
holgerkoser, line 28: compare primitive values with eql and objects and arrays with eql. Why not compare to hole body.

backend/lib/services/controllerregistrations.js
```diff
@@ -0,0 +1,26 @@
...
+
+function fromResource (controllerregistrations) {
+  return _.map(controllerregistrations, ({ metadata, spec }) => {
+    return {
+      name: metadata.name,
+      version: get(spec, 'deployment.providerConfig.values.image.tag'),
+      resources: spec.resources
+    }
+  })
+}
```
holgerkoser, line 21:
```suggestion
const _ = require('lodash')

function fromResource (controllerregistrations) {
  return _.map(controllerregistrations, ({ metadata, spec }) => {
    return {
      name: metadata.name,
      version: _.get(spec, 'deployment.providerConfig.values.image.tag'),
      resources: spec.resources
    }
  })
}
```

backend/test/cache.spec.js
```diff
@@ -116,12 +130,14 @@ describe('cache', function () {
       let syncQuotasSpy
       let syncSeedsSpy
       let syncProjectsSpy
+      let syncControllerregistrationsSpy
```
holgerkoser, line 133:
```suggestion
      let syncControllerRegistrationsSpy
```

backend/test/cache.spec.js
```diff
@@ -116,12 +130,14 @@ describe('cache', function () {
...
 
       beforeEach(function () {
         syncCloudprofilesSpy = sandbox.spy(gardenerCore.cloudprofiles, 'syncList')
         syncQuotasSpy = sandbox.spy(gardenerCore.quotas, 'syncListAllNamespaces')
         syncSeedsSpy = sandbox.spy(gardenerCore.seeds, 'syncList')
         syncProjectsSpy = sandbox.spy(gardenerCore.projects, 'syncList')
+        syncControllerregistrationsSpy = sandbox.spy(gardenerCore.controllerregistrations, 'syncList')
```
holgerkoser, line 140:
```suggestion
        syncControllerRegistrationsSpy = sandbox.spy(gardenerCore.controllerregistrations, 'syncList')
```

backend/test/cache.spec.js
```diff
@@ -131,6 +147,7 @@ describe('cache', function () {
         expect(syncQuotasSpy).to.be.calledOnce
         expect(syncSeedsSpy).to.be.calledOnce
         expect(syncProjectsSpy).to.be.calledOnce
+        expect(syncControllerregistrationsSpy).to.be.calledOnce
```
holgerkoser, line 150:
```suggestion
        expect(syncControllerRegistrationsSpy).to.be.calledOnce
```

backend/test/support/common.js
```diff
@@ -131,6 +131,36 @@ const quotaList = [
   getQuota({ name: 'foo-quota2', namespace: 'garden-foo' })
 ]
 
+const controllerregistrationsList = [
```
holgerkoser, line 134:
```suggestion
const controllerRegistrationList = [
```

backend/test/support/common.js
```diff
@@ -152,6 +182,10 @@ const stub = {
     const ticketCache = createTicketCache()
     getTicketCacheStub.returns(ticketCache)
     return ticketCache
+  },
+  getControllerRegistrations (sandbox) {
+    const getControllerRegistrationsStub = sandbox.stub(cache, 'getControllerRegistrations')
+    getControllerRegistrationsStub.returns(controllerregistrationsList)
```
holgerkoser, line 188:
```suggestion
    getControllerRegistrationsStub.returns(controllerRegistrationList)
```

frontend/src/components/MainToolbar.vue
```diff
@@ -28,13 +28,30 @@ SPDX-License-Identifier: Apache-2.0
...
         </template>
-        <v-card tile>
+        <v-card tile width="300px">
           <v-card-title primary-title>
             <div class="content">
               <div class="title mb-2">Gardener</div>
               <v-progress-circular size="18" indeterminate v-if="!dashboardVersion"></v-progress-circular>
               <div class="caption" v-if="!!gardenerVersion">API version {{gardenerVersion}}</div>
               <div class="caption" v-if="!!dashboardVersion">Dashboard version {{dashboardVersion}}</div>
+              <v-expansion-panels flat v-if="extensionCount && isAdmin">
```
holgerkoser, line 38: I think this info should not be displayed in the help menu.

frontend/src/components/MainToolbar.vue
```diff
@@ -211,6 +245,8 @@ export default {
           this.setError({
             message: `Failed to fetch version information. ${err.message}`
           })
+          this.dashboardVersion = this.dashboardVersion || 'Unknown'
```
holgerkoser, line 248: Why do we need this line?

frontend/src/components/MainToolbar.vue
```diff
@@ -236,4 +272,14 @@ export default {
     color: white;
     font-weight: bold;
   }
+
+  .extension-expansion-panel {
+    max-height: 20vh;
```
holgerkoser, line 277: Why unit vh?

frontend/src/store/index.js
```diff
@@ -941,6 +955,13 @@ const actions = {
         dispatch('setError', err)
       })
   },
+  fetchControllerRegistrations ({ dispatch }) {
+    try {
+      return dispatch('controllerRegistrations/getAll')
```
holgerkoser, line 960:
```suggestion
      return await dispatch('controllerRegistrations/getAll')
```

holgerkoser: Without await we do not catch an error here

backend/lib/services/controllerregistrations.js
```diff
@@ -0,0 +1,26 @@
...
+
+exports.list = function () {
+  const controllerregistrations = getControllerRegistrations()
```
petersutter, line 24: add authorization check for can list controllerregistrations

frontend/src/components/MainToolbar.vue
grolu: Yes, I agree. We can also hide this completely for the moment and make an additional PR where we add a decent about dialog. Alternatively, I can include i in this PR WDYT?

frontend/src/components/MainToolbar.vue
grolu: If we are not able to fetch the version information I had an endless spinner. I think it is better to show unknown in this case. I don't know I think this worked before (at least the spinner was hidden after timeout)

frontend/src/components/MainToolbar.vue
grolu: I chose ch to expand the main menu height dynamically, depending on the browser size. Anyway, this will be removed when we move the extension information to another dialog

frontend/src/components/MainToolbar.vue
holgerkoser: I understand this for the gardener version but not for the dashboard version

grolu: the dashboard version is also retrieved from the backend

frontend/src/store/index.js
```diff
@@ -941,6 +955,13 @@ const actions = {
         dispatch('setError', err)
       })
   },
+  fetchControllerRegistrations ({ dispatch }) {
```
holgerkoser, line 958:
```suggestion
  async fetchControllerRegistrations ({ dispatch }) {
```

frontend/src/components/MainToolbar.vue
holgerkoser: Fix version endpoint in backend. Should always return dashboard version.

frontend/src/components/MainToolbar.vue
grolu: removed extension list from this PR, will file an additional PR after this one is merged where we build an about screen to display this kind of information

backend/test/acceptance/api.controllerregistrations.spec.js
grolu: done

frontend/src/components/MainToolbar.vue
grolu: fixed, always return dashboard version

frontend/src/components/MainToolbar.vue
grolu: removed completely as we no longer show the list in this PR

backend/lib/services/controllerregistrations.js
grolu: done

backend/test/acceptance/api.controllerregistrations.spec.js
```diff
@@ -25,7 +28,6 @@ module.exports = function ({ agent, sandbox, auth, k8s }) {
     expect(res).to.have.status(200)
     expect(res).to.be.json
     expect(res.body).to.have.length(2)
-    expect(res.body[0].name).eql('foo')
-    expect(res.body[0].version).eql('v1.0.0')
+    expect(res.body[0]).to.include({ name: 'foo', version: 'v1.0.0' })
```
holgerkoser, line 31: I would expect
```
expect(res.body).to.include({ name: 'foo', version: 'v1.0.0' })
```
or
```
expect(res.body[0]).to.eql({ name: 'foo', version: 'v1.0.0' })
```

grolu: `expect(res.body).to.include({ name: 'foo', version: 'v1.0.0' })`

does not work,

`expect(res.body[0]).to.eql({ name: 'foo', version: 'v1.0.0' })`

also does not work as the object has other properties.

holgerkoser: Ok

holgerkoser (approved):
/lgtm

frontend/src/router/guards.js
```diff
@@ -180,3 +181,9 @@ function ensureKubeconfigDataLoaded (store) {
     return store.dispatch('fetchKubeconfigData')
   }
 }
+
+function ensureControllerRegistrationsLoaded (store) {
+  if (isEmpty(store.getters.controllerRegistrationList)) {
+    return store.dispatch('fetchControllerRegistrations')
```
petersutter, line 187: only call `fetchControllerRegistrations` if you have the authorization to do so, similar like
```js
if (store.getters.canUseProjectTerminalShortcuts) {
  promises.push(store.dispatch('ensureProjectTerminalShortcutsLoaded'))
}
```

grolu: Yes, you are right of course. We need to discuss this PR in general - as we know know that the registrations cannot be accessed by regular users, we need to establish a way for non admins to fetch the network extensions

backend/lib/services/controllerregistrations.js
```diff
@@ -30,3 +30,14 @@ exports.list = async function ({ user }) {
...
+
+exports.listNetworkingTpes = async function () {
+  const controllerregistrations = getControllerRegistrations()
+  const networkExtensions = _.filter(controllerregistrations, ({ spec }) => {
+    return _.find(spec.resources, { kind: 'Network' })
+  })
+
+  return _.flatMap(networkExtensions, ({ spec }) => {
+    return _.map(spec.resources, 'type')
+  })
```
petersutter, line 42:
```suggestion
  return _
    .chain(controllerregistrations)
    .map('spec.resources')
    .flatMap()
    .filter(['kind', 'Network'])
    .map('type')
    .value()
```

frontend/src/router/guards.js
```diff
@@ -115,7 +114,8 @@ function ensureDataLoaded (store, localStorage) {
         case 'NewShoot':
         case 'NewShootEditor': {
           const promises = [
-            store.dispatch('subscribeShoots')
+            store.dispatch('subscribeShoots'),
+            store.dispatch('fetchNetworkingTypes')
```
petersutter, line 118: both needed also for the new shoot editor?

petersutter (changes requested):

backend/lib/routes/controllerregistrations.js
```diff
@@ -0,0 +1,31 @@
...
+  .get(async (req, res, next) => {
+    try {
+      const user = req.user
+      res.send(await controllerregistrations.list({ user }))
+    } catch (err) {
+      next(err)
+    }
+  })
+
+router.route('/resources/kind/network/types')
```
holgerkoser, line 24: I don't like this pattern since everything in between will return 404.

holgerkoser, line 24:
```suggestion
router.route('/network-types')
```

backend/lib/services/controllerregistrations.js
```diff
@@ -30,3 +30,14 @@ exports.list = async function ({ user }) {
   const controllerregistrations = getControllerRegistrations()
   return fromResource(controllerregistrations)
 }
+
+exports.listNetworkingTpes = async function () {
+  const controllerregistrations = getControllerRegistrations()
```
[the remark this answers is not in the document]
grolu: Yes, every authenticated used should be able to. I mean - we could also check for the right to create shoots, because otherwise the user doesn't need this information after all. However, idk if we should implement such checks. We discussed this - we agreed, that this is in any case not an ideal solution, as the dashboard backend returns information that a regular (non admin) user is not able to get by accessing the gardener api. However, there is currently no other way...

backend/lib/routes/info.js
```diff
@@ -50,9 +49,6 @@ async function fetchGardenerVersion () {
     return version
   } catch (err) {
     logger.warn(`Could not fetch gardener version. Error: ${err.message}`)
-    if (isHttpError(err, 404)) {
-      return undefined
-    }
-    throw err
+    return undefined
```
holgerkoser, line 52:
```suggestion
```

backend/lib/services/controllerregistrations.js
grolu: ah yes, good point to use chain. I moved this from the frontend where we cannot use chaining and forgot to change :)

backend/lib/routes/controllerregistrations.js
grolu: Yes, I thought that you will raise this point. What would you suggest? Do not use a path and hust have an independent route like `/networktypes`?

grolu: ah just saw your other comment, ok. Let's do it like that

frontend/src/router/guards.js
grolu: Yes, we need this to set the initial network type also for the editor (when accessing the editor page directly, without navigating from UI component view)

backend/lib/services/controllerregistrations.js
```diff
@@ -0,0 +1,43 @@
...
+  return fromResource(controllerregistrations)
+}
+
+exports.listNetworkingTpes = async function () {
+  const controllerregistrations = getControllerRegistrations()
+  return _
+    .chain(controllerregistrations)
+    .map('spec.resources')
+    .flatMap()
```
holgerkoser, line 39:
```suggestion
    .flatMap('spec.resources')
```

backend/lib/routes/index.js
```diff
@@ -16,6 +16,7 @@ module.exports = {
   '/cloudprofiles': require('./cloudprofiles'),
   '/seeds': require('./seeds'),
   '/shoots': require('./shoots'),
+  '/controllerregistrations': require('./controllerregistrations'),
```
holgerkoser, line 19: I think we should have to separate routes. 
* `controllerregistrations` or better `extensions` or `gardener-extensions`
* `networking-types` 
The networking-types is independent since we do not check any authorization here. The authorization check should not be less strict for a subresource. That is strange.

grolu: I agree in that case I need to introduce a new route.
Regrading renaming - I don't like to call it extensions, I know we talk about extensions when a referring to them but as we are more or less returning the `controllerregistrations` resource I think we should keep the name.

backend/lib/routes/controllerregistrations.js
```diff
@@ -0,0 +1,31 @@
...
+      res.send(await controllerregistrations.list({ user }))
+    } catch (err) {
+      next(err)
+    }
+  })
+
+router.route('/network-types')
+  .get(async (req, res, next) => {
+    try {
+      res.send(await controllerregistrations.listNetworkingTpes())
```
holgerkoser, line 27:
```suggestion
      res.send(await controllerregistrations.listNetworkingTypes())
```

backend/lib/routes/index.js
grolu: done, split into two routes:
`/gardener-extensions`
`/networking-types`

frontend/src/router/guards.js
petersutter: but when you switch from the `NewShootEditor` to the `NewShoot`, this code will be executed again, or not? Then you can only fetch it in case of `NewShoot`

petersutter (approved):
only minor suggestion to optimise (if possible), otherwise /lgtm

frontend/src/router/guards.js
grolu: If the user enters the route with the editor, the network types must also be loaded. so how to achieve this then?

PR #884, opened 2020-11-12
Save shoot list filter options in local storage

### LINKED ISSUE #866: Filter prefences are lost on page refresh
**What happened**:
If I change the filters on  https://dashboard.example.com/namespace/_all/shoots/ after refres (ctrl+R/F5) the filters are reset to the default one.

The filters settings to be persisted and to survive page refresh.

Open the https://dashboard.example.com/namespace/_all/shoots/ page, change some of the filters settings, refresh the page and see that the filters are reset to the default values.

**Environment**:
Dashboard Version: 1.46.1
Browser: Firefox 82.0 on Ubuntu 18.04

petersutter:
Save shoot list filter options in local storage

Fixes #866 

**Release note**:

```improvement user

```

### DIFF
--- a/frontend/src/store/index.js
+++ b/frontend/src/store/index.js
@@ -81,7 +81,6 @@ const state = {
     incomplete: false,
     evaluationError: null
   },
-  onlyShootsWithIssues: true,
   sidebar: true,
   user: null,
   redirectPath: null,
@@ -132,8 +131,11 @@ class Shortcut {
   }
 }
 
-const getFilterValue = (state) => {
-  return state.namespace === '_all' && state.onlyShootsWithIssues ? 'issues' : null
+const getFilterValue = (getters) => {
+  if (state.namespace === '_all' && getters.onlyShootsWithIssues) {
+    return 'issues'
+  }
+  return null
 }
 
 const vendorNameFromImageName = imageName => {
@@ -922,6 +924,9 @@ const getters = {
   },
   isKubeconfigEnabled (state) {
     return !!(get(state, 'kubeconfigData.oidc.clientId') && get(state, 'kubeconfigData.oidc.clientSecret'))
+  },
+  onlyShootsWithIssues (state, getters) {
+    return getters['shoots/onlyShootsWithIssues']
   }
 }
 
@@ -1067,9 +1072,9 @@ const actions = {
         dispatch('setError', err)
       })
   },
-  async subscribeShoots ({ dispatch, commit, state }) {
+  async subscribeShoots ({ dispatch, commit, state, getters }) {
     try {
-      EmitterWrapper.shootsEmitter.subscribeShoots({ namespace: state.namespace, filter: getFilterValue(state) })
+      EmitterWrapper.shootsEmitter.subscribeShoots({ namespace: state.namespace, filter: getFilterValue(getters) })
     } catch (err) { /* ignore error */ }
   },
   async subscribeComments ({ dispatch, commit }, { name, namespace }) {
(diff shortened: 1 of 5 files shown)


frontend/src/store/index.js
```diff
@@ -132,8 +131,11 @@ class Shortcut {
   }
 }
 
-const getFilterValue = (state) => {
-  return state.namespace === '_all' && state.onlyShootsWithIssues ? 'issues' : null
+const getFilterValue = (getters) => {
+  if (state.namespace === '_all' && getters.onlyShootsWithIssues) {
+    return 'issues'
+  }
+  return null
```
holgerkoser, line 138:
```suggestion
function getFilterValue (state, getters) {
  if (state.namespace === '_all' && getters.onlyShootsWithIssues) {
    return 'issues'
  }
  return null
```

petersutter: ✅

grolu (approved):
/lgtm

frontend/src/store/modules/shoots.js
```diff
@@ -90,6 +90,9 @@ const getters = {
   getShootListFilters () {
     return state.shootListFilters
   },
+  onlyShootsWithIssues (state, getters) {
```
holgerkoser, line 93: genau richtig 👏

frontend/src/views/ShootList.vue
```diff
@@ -314,9 +314,12 @@ export default {
       }
     },
     toggleFilter (key) {
-      if (this.showOnlyShootsWithIssues) {
-        const filters = this.getShootListFilters
-        this.setShootListFilter({ filter: key, value: !filters[key] })
+      const filters = this.getShootListFilters
+      this.setShootListFilter({ filter: key, value: !filters[key] })
+      this.$localStorage.setObject('shootListFilter', pick(filters, ['onlyShootsWithIssues', 'progressing', 'userIssues', 'deactivatedReconciliation', 'hideTicketsWithLabel']))
```
holgerkoser, line 319: I don't understand why this works? Is the correct value stored?

frontend/src/store/index.js
```diff
@@ -1094,17 +1099,19 @@ const actions = {
...
       })
   },
-  setShootListFilters ({ dispatch, commit }, value) {
-    return dispatch('shoots/setShootListFilters', value)
-      .catch(err => {
-        dispatch('setError', err)
-      })
+  async setShootListFilters ({ dispatch, getters }, value) {
+    try {
+      await dispatch('shoots/setShootListFilters', value)
```
holgerkoser, line 1104: I know it is no part of this PR but we don't need an action here. We could directly call the mutation in the shoot module.
```
  commit('shoots/SET_SHOOT_LIST_FILTERS', value)
```

petersutter: `shoots/SET_SHOOT_LIST_FILTERS` is working with the rootState.
`modules/shoots.js`:
```js
  SET_SHOOT_LIST_FILTERS (state, { rootState, value }) {
    state.shootListFilters = value
    setFilteredAndSortedItems(state, rootState)
  },
```

 Not sure if I may pass state as rootState like e.g.
```js
commit('shoots/SET_SHOOT_LIST_FILTERS', { rootState: state, value })
```

holgerkoser: You are right. We need at least the action on the root module. As I said, it not part of this PR. We will refactor that anyway and may get ride of the rootState at all. You can already remove the actions in the shoots module or keep them.

petersutter: as discussed offline we will address this in the upcoming, planned refactoring

frontend/src/views/ShootList.vue
petersutter: maybe it's now clearer?

PR #899, opened 2020-12-07
Permanently hide alert banner

### LINKED ISSUE #887: Closing the Warning box it's not remembered
**What happened**:
I'm closing this alert box on every login, and still pops up.
image

I would expect that my action is remembered, and no alert is shown next login

Login, and notice the Orange Alert on the top. Close the alert. Log out and Login again.
Alert will still be visible

**Environment**:
garden.canary.k8s.ondemand.com

petersutter:
This PR adds an option to hide the configured alert banner permanently.

- `GAlert` component will be renamed to `GMessage`
- A new component `GAlert` will be introduced that only displays the alert in case the message was not closed by the user

When configuring an alert as admin, you can now set an `identifier` for the message. This identifier will be used to identify which messages the user wants to hide permanently.
This gives you the ability to adapt the message text and use the same identifier without that the user needs to read and hide the message again.
On the other hand, you can force the display of the message by changing the identifier.

Example `values.yaml`

```yaml
frontendConfig:
  alert:
    type: error
    message: This is an **alert** banner
    # identifer: alert-identifier # Optional. If not set, a hash of the alert.message will be used as identifier. The identifier is used to store the user's decision to not show the alert message with this identifier again.
```

Fixes #887

**Release note**:

```bugfix user
The alert banner is now permanently hidden when clicking the close icon
```

### DIFF
--- a/frontend/src/components/GAlert.vue
+++ b/frontend/src/components/GAlert.vue
@@ -5,69 +5,118 @@ SPDX-License-Identifier: Apache-2.0
 -->
 
 <template>
-  <v-alert :color="color" :tile="tile" :dark="dark" dismissible v-model="alertVisible">
-    <div class="subtitle-1">
-      {{message}}
-      <v-btn dark outlined small v-if="!!detailedMessage" @click="detailedMessageVisible = !detailedMessageVisible">
-        Details
-      </v-btn>
-    </div>
-    <transition name="fade">
-      <div v-if="!!detailedMessageVisible">
-        <code>{{detailedMessage}}</code>
-      </div>
-    </transition>
-  </v-alert>
+  <div>
+    <v-alert
+      class="alertBanner"
+      :type="type"
+      v-model="alertVisible"
+      :color="color"
+      :transition="transition"
+    >
+      <v-row align="center">
+        <v-col class="grow pa-0">
+          <div v-if="message" class="alert-banner-message" v-html="messageHtml"></div>
+          <slot v-else name="message"></slot>
+        </v-col>
+        <v-col class="shrink py-0">
+          <v-btn small icon @click="closeBanner"><v-icon>mdi-close-circle</v-icon></v-btn>
+        </v-col>
+      </v-row>
+    </v-alert>
+    <confirm-dialog ref="confirmDialog"></confirm-dialog>
+  </div>
 </template>
 
 <script>
+import { mapActions } from 'vuex'
+import { transformHtml } from '@/utils'
+import ConfirmDialog from '@/components/dialogs/ConfirmDialog'
+
+const LOCAL_STORE_ALERT_BANNER_HIDDEN_MESSAGES = 'global/alert-banner/hidden-messages'
+
 export default {
-  name: 'galert',
+  components: {
+    ConfirmDialog
+  },
   props: {
-    message: {
+    message: { // alternatively, use message slot
       type: String
     },
-    detailedMessage: {
+    type: {
+      type: String,
+      default: 'error'
+    },
+    identifier: {
       type: String
     },
     color: {
-      type: String,
-      required: true
+      type: String
     },
-    tile: {
-      type: Boolean
-    },
-    dark: {
-      type: Boolean,
-      default: true
+    transition: {
+      type: String
     }
   },
   data () {
     return {
-      detailedMessageVisible: false
+      alertVisible: false
     }
   },
   computed: {
-    alertVisible: {
-      get () {
-        return !!this.message
-      },
-      set (value) {
-        if (!value) {
-          this.$emit('update:message', undefined)
-          this.$emit('update:detailedMessage', undefined)
-        }
+    messageHtml () {
+      return transformHtml(this.message)
+    }
+  },
+  methods: {
+    ...mapActions([
+      'setAlertBanner'
+    ]),
+    async closeBanner () {
+      const result = await this.$refs.confirmDialog.waitForConfirmation({
+        confirmButtonText: 'Hide',
+        captionText: 'Confirm Hide Message',
+        messageHtml: 'Do you want to hide this message?',
+        showDoNotAskAgain: true
+      })
+      if (!result) {
+        return
       }
+      const { doNotAskAgain } = result
+      if (doNotAskAgain) {
+        const permanentlyHiddenIds = this.getPermanentlyHiddenIds()
+        permanentlyHiddenIds[this.identifier] = true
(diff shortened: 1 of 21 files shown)


grolu:
I'm a bit confused regarding the caption of the textbox. What does `do not ask me again` mean? From the code it looks like activating the checkbox permanently hides the message? In that case it should rather be labelled as `Permanently hide message`.

petersutter:
Good point, I will change it as you suggested 👍

petersutter:
After second thought, I changed the caption to „Hide Message“ as it is only permanently if you  select the checkbox

grolu:
IDK ... I don't think it makes it clear. For me the checkbox still means that I don't want to see this confirmation popup again but I would never expect that this controls if the alert message is permanently hidden or not.

petersutter:
As we discussed offline, I will remove the popup altogether and the alert will be permanently hidden when clicking on the close icon

frontend/src/components/GAlert.vue
```diff
@@ -5,69 +5,101 @@ SPDX-License-Identifier: Apache-2.0
...
+      const permanentlyHiddenIds = this.$localStorage.getObject(LOCAL_STORE_ALERT_BANNER_HIDDEN_MESSAGES) || {}
+      return permanentlyHiddenIds[this.identifier] === true
+    },
+    updateAlertVisibility () {
+      const visible = !this.isPermanentlyHidden(this.identifier)
+      this.setAlertVisibility(visible)
+    },
+    setAlertVisibility (visible) {
+      this.alertVisible = visible
+      this.$emit('value', visible)
```
holgerkoser, line 87: Why do we emit visibility?
Why is the name value? Shouldn't it be a value prop with two ways data binding and emit `input`?

frontend/src/components/GMessage.vue
```diff
@@ -0,0 +1,72 @@
...
+    }
+  },
+  computed: {
+    alertVisible: {
+      get () {
+        return !!this.message
+      },
+      set (value) {
+        if (!value) {
+          this.$emit('update:message', undefined)
```
holgerkoser, line 56: I think the message prop does not change only the visiblility?

frontend/src/components/dialogs/GDialog.vue
```diff
@@ -22,7 +22,7 @@ SPDX-License-Identifier: Apache-2.0
         <slot name="message">
           This is a generic dialog template.
         </slot>
-        <g-alert color="error" class="mt-4" :message.sync="message" :detailedMessage.sync="detailedMessage"></g-alert>
+        <g-message color="error" class="mt-4" :message.sync="message" :detailedMessage.sync="detailedMessage"></g-message>
```
holgerkoser, line 25: I cannot understand why message and detailedMessage are sync properties?

frontend/src/components/GAlert.vue
```diff
@@ -5,69 +5,101 @@ SPDX-License-Identifier: Apache-2.0
...
+    ]),
+    async closeBanner () {
+      if (this.identifier) { // hide permanently
+        this.hidePermanently(this.identifier)
       }
+
+      this.setAlertVisibility(false)
+    },
+    hidePermanently (identifier) {
+      const permanentlyHiddenIds = this.$localStorage.getObject(LOCAL_STORE_ALERT_BANNER_HIDDEN_MESSAGES) || {}
```
holgerkoser, line 69: Why don't we write these values into the vuex store and decide there centrally (via localStorage plugin) what to persist in localStroage. I think this would be much cleaner.

holgerkoser:
```js
import debounce from 'lodash/debounce'

const ALERT_BANNER_HIDDEN_MESSAGES = 'global/alert-banner/hidden-messages'
const SET_ALERT_BANNER_HIDDEN_MESSAGES = 'SET_ALERT_BANNER_HIDDEN_MESSAGES'


const setAlertBannerHiddenMessages = debounce(value => {
  localStorage.setItem(ALERT_BANNER_HIDDEN_MESSAGES, JSON.stringify(value))
}, 200)

const LocalStoragePlugin= store => {
  const value = localStorage.getItem(ALERT_BANNER_HIDDEN_MESSAGES)
  if (value) {
    store.commit(SET_ALERT_BANNER_HIDDEN_MESSAGES, JSON.parse(value))
  }
  store.subscribe((mutation, state) => {
    switch (mutation.type) {
      case SET_ALERT_BANNER_HIDDEN_MESSAGES: 
        setAlertBannerHiddenMessages(mutation.payload)
        break
    }
  })
}

const store = new Vuex.Store({
  state,
  getters,
  actions,
  mutations,
  plugins: [LocalStoragePlugin]
})
```

holgerkoser: Let's create a separate ticket for this.

frontend/src/components/GAlert.vue
holgerkoser: Either add a value (v-model) property and emit input or just emit input with value prop.

frontend/src/store/index.js
```diff
@@ -849,17 +849,24 @@ const getters = {
     const user = state.user
     return user ? user.name || fullDisplayName(user.id) : ''
   },
-  alertMessage (state) {
```
holgerkoser, line 852: This is needed. Do not remove.

frontend/src/components/GAlert.vue
petersutter: https://github.com/gardener/dashboard/issues/905

frontend/src/components/GAlert.vue
petersutter: I removed the emit completely. The component handles the visibility which depends if the user already dismissed the message already or not

frontend/src/components/dialogs/GDialog.vue
petersutter: as discussed offline we will not start this refactoring in this pr

frontend/src/store/index.js
petersutter: ✅

frontend/src/components/GMessage.vue
petersutter: see https://github.com/gardener/dashboard/pull/899#discussion_r544370302

frontend/src/store/index.js
petersutter: ✅

holgerkoser (approved):
/lgtm

PR #900, opened 2020-12-08
Backend tests MUST be transport independent

holgerkoser:
This PR make all backend tests independent from the underlying http transport. This is a prerequisite to migrate the `kube-client` transport from http/1.1 to http/2. With http/2 sessions we are able to perform "liveness" checks on the underlying tcp connection and we will get rid of the watch with `resourceVersion` problem (watch caches of different apiservers have different resourceVersions).  

In addition the test framework has been migrated from multi tool solution [mocha](https://mochajs.org/), [sinon](https://sinonjs.org/), [chai](https://www.chaijs.com/api/bdd/), [nock](https://github.com/nock/nock) and [istanbuljs/nyc](https://github.com/istanbuljs/nyc) to the single tool solution [jest](https://jestjs.io/). All other gardener-dashboard packages already use jest as test runner.

Fixes #

**Release note**:

```bugfix user

```


petersutter (approved):
/lgtm

PR #909, opened 2020-12-16
EditableText should not allow to save if an error exists

### LINKED ISSUE #908: Invalid cost center can be entered
**What happened**:
If you enter an invalid cost center on the project administration page, it can be applied despite an error message. The Apply button is disabled but by pressing the Enter key the wrong cost center is saved in the project anyway.
Screen Shot 2020-12-16 at 12 41 01
An invalid cost center is assigned to the project.
Screen Shot 2020-12-16 at 12 41 20

It is expected that when pressing the Enter key the dialog will remain open and the error message will persist.

**Environment**:

holgerkoser:
With this PR it is no longer possible to save invalid data of an `EditableText` when the ENTER key is pressed. 

Fixes #908

**Release note**:

```bugfix user
An invalid Cost Object cannot be saved by pressing the ENTER key. The popover remains open and the error message persists.
```

### DIFF
--- a/frontend/src/components/editable/EditableText.vue
+++ b/frontend/src/components/editable/EditableText.vue
@@ -173,6 +173,9 @@ export default {
       this.isActive = false
     },
     async onSave (value) {
+      if (this.error) {
+        return
+      }
       this.loading = this.color
       try {
         await this.save(this.internalValue)


petersutter (approved):
/lgtm

PR #912, opened 2020-12-17
Do not directly set innerHTML of html elements

### LINKED ISSUE #910: Do not directly set innerHTML of html elements
Do not set innerHTML of html elements. Instead create all element with `document.createElement` and use `element.textContent` instead.

**Why is this needed**:
This is important that we do not forget to escape text in generated html which could lead to XSS vulnerability.

holgerkoser:
Do not use `element.innerHTML` of DOM elements. Instead create all elements with `document.createElement` and use `element.textContent` instead.

Fixes #910

**Release note**:

```other developer

```


petersutter (approved):
/lgtm

PR #913, opened 2020-12-17
Improve mock implementations

holgerkoser:
Reject with HttpError 503  in all mock implementations of mockRequest if the `header[':path']` does not match the pattern. 

Fixes #

**Release note**:

```bugfix user

```


grolu (approved):
/lgtm

PR #919, opened 2020-12-21
Add support to re-login with an redirectPath given

### LINKED ISSUE #865: Re-login on expired sesion does not navigate to the last page
**What happened**:
My session has expired and I was logged out. Before this I was on https://dashboard.example.com/namespace/_all/shoots/, but after I log in I was redirected to my "home" page, in my case it is https://dashboard.example.com/namespace/garden/shoots

To be re-directed to the last page on re-loging.

**Environment**:
Dashboard Version: 1.46.1
Browser: Firefox 82.0 on Ubuntu 18.04

NotTheEvilOne:
This PR adds support to (force) signout a user, redirect him to the login view but remember the original view (path) presented before a session expired.

Fixes #865 

The intended `redirectPath` is stored in local storage. For signouts without an error the item is cleaned up.


gardener-robot-ci-2:
Thank you @NotTheEvilOne for your contribution. Before I can start building your PR, a member of the organization must set the required label(s) {'reviewed/ok-to-test'}. Once started, you can check the build status in the PR checks section below.

holgerkoser:
When a session is terminated by logging out, whether by the user themselves or by a session timeout, any state of the session should be cleared. If you want to preserve the last page visited, this should not be passed through the logout endpoint. If at all, it should be stored in a user specific way / store. We plan to store user specific data in the future and also plan to do an automatic refresh of the token. This would allow the token to expire only in case of prolonged inactivity. Wouldn't that be the right solution to the problem?

NotTheEvilOne:
Sounds feasable. I'll force-push this branch to reuse this.$localstorage soon.

NotTheEvilOne:
I've updated the implementation to use the mentioned local storage approach. Beside automatic token refresh this should implement the intended use case without breaking current default behavior.

frontend/src/views/Login.vue
```diff
@@ -97,7 +98,11 @@ export default {
       'cfg'
     ]),
     redirectPath () {
-      return get(this.$route.query, 'redirectPath', '/')
+      let redirectPath = get(this.$route.query, 'redirectPath', '/')
+      if (redirectPath === '/' && this.$localStorage.getItem(LOCAL_STORE_AUTH_LOGIN_REDIRECTION_PATH) !== null) {
+        redirectPath = this.$localStorage.getItem(LOCAL_STORE_AUTH_LOGIN_REDIRECTION_PATH)
```
holgerkoser, line 103:
```suggestion
        redirectPath = this.$localStorage.getItem(LOCAL_STORE_AUTH_LOGIN_REDIRECTION_PATH)
        this.$localStorage.removeItem(LOCAL_STORE_AUTH_LOGIN_REDIRECTION_PATH)
```

holgerkoser: Remove the store redirect path once it has been used.

holgerkoser:
My apologies for only now responding to the revised PR. I had unfortunately completely lost sight of it. Anyway, thanks for including the changes with the localStorage. However, I don't think the PR really solves the problem, as on automatic logout (https://github.com/gardener/dashboard/blob/master/frontend/src/router/index.js#L40), the current path is not saved in localStorage. I think the line https://github.com/gardener/dashboard/blob/master/frontend/src/router/guards.js#L79 where you store the current path is only an absolute special case and occurs only very rarely when the `getPrivileges` call returns with 401.

NotTheEvilOne:
> My apologies for only now responding to the revised PR. I had unfortunately completely lost sight of it. [...]

I'm sorry for my late reply as well. RL got me focusing on other things these days.

> However, I don't think the PR realy solves the problem, as on automatic logout (https://github.com/gardener/dashboard/blob/master/frontend/src/router/index.js#L40), the current path is not saved in localStorage. I think the line https://github.com/gardener/dashboard/blob/master/frontend/src/router/guards.js#L79 where you store the current path is only an absolute special case and occurs only very rarely when the `getPrivileges` call returns with 401.

I do agree that the special handling is only happening when the token used is invalidated. As far as I can see that happens before the automatic logout as the token is already invalid at that time. Unfortunately it is not easy to reproduce this corner case.

For normal logout other mechanism seems to be in place as if I log out and log in again I get redirected to the last viewed page right away. That's why I only implemented it for the 401.

> By the way, next I will start implementing the OIDC id_token refresh and then there should be no more automatic logout.

Feel free to close this PR in that case :)

grolu:
We will start implementing OIDC id_token refresh and this should solve this issue.
You can track the progress here: https://github.com/gardener/dashboard/issues/976

PR #922, opened 2020-12-21
Fixed Control Plane Migration Indication

### LINKED ISSUE #892: Fix Control Panel Migration Indication
Currently this is not working properly as we check

`this.shootStatusSeedName && this.shootSeedName && this.shootStatusSeedName !== this.shootSeedName`

However, during the restore phase of CP Migration the Shoot.Status.SeedName = nil and the Shoot.Spec.SeedName = New Seed  so just from those values it looks as normal reconciliation
We can try using the LastOperation from the `Shoot.Status`. The `LastOperation.Type` will be `migrate` when the resources from the old seed are being deleted, then the `LastOperation.Type` will be switched to `restore` when the control plane components are being created on the new seed 

Gardener actually uses the restore operation to determine whether it should restore the control plane or do a regular reconciliation

grolu:

Fixes #892

**Release note**:

```bugfix operator
The status indicator for control plane migration is now shown correctly
```

### DIFF
--- a/frontend/src/components/ShootSeedName.vue
+++ b/frontend/src/components/ShootSeedName.vue
@@ -6,27 +6,17 @@ SPDX-License-Identifier: Apache-2.0
 
 <template>
   <v-row align="center" class="ma-0">
-    <v-tooltip top :disabled="!isControlPlaneMigrating">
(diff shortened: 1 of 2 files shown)


holgerkoser (approved):
/lgtm

PR #928, opened 2020-12-30
Improve wording in shoot action tooltip

istvanballok:

As a new operator, we were confused by the wording: "Actions disabled for cluster purpose infrastructure".
We think that by adding a "with" preposition, it becomes clear that purpose is a property of the cluster and infrastructure is the value of the purpose property.

image

@BeckerMax 

**Release note**:

```other operator
NONE
```


grolu (approved):
/lgtm

PR #931, opened 2021-01-11
Convert camel case html attributes to kebab-case

grolu:
**What this PR does / why we need it**:
Prop names should always use camelCase during declaration, but kebab-case in templates and JSX.
https://vuejs.org/v2/style-guide/#Prop-name-casing-strongly-recommended

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Release note**:

```bugfix user

```

PR #936, opened 2021-01-15
Use kebab case for custom html events

grolu:
This PR addresses some issues introduced with #931

Fixes #

**Release note**:

```bugfix user

```

### DIFF
--- a/frontend/src/components/icons/IconBase.vue
+++ b/frontend/src/components/icons/IconBase.vue
@@ -8,7 +8,7 @@ SPDX-License-Identifier: Apache-2.0
   <svg xmlns="http://www.w3.org/2000/svg"
     :width="width"
     :height="height"
-    :view-box="viewBox"
+    :viewBox="viewBox"
     :aria-labelledby="iconName"
     role="presentation"
   >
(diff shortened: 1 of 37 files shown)


frontend/src/components/icons/IconBase.vue
```diff
@@ -8,7 +8,7 @@ SPDX-License-Identifier: Apache-2.0
   <svg xmlns="http://www.w3.org/2000/svg"
     :width="width"
     :height="height"
-    :view-box="viewBox"
+    :viewBox="viewBox"
```
holgerkoser, line 11: I think we should add a comment that this must be camelCase.

holgerkoser:
The lifecycle hooks https://github.com/gardener/dashboard/blob/1538be54ab96c76b32f8c5ad556a195554d24056/frontend/src/mixins/asyncRef.js#L48-L49 must also be camelCase. I am not sure I you don't touched them by intention or by accident.

holgerkoser (approved):
/lgtm

frontend/src/components/icons/IconBase.vue
grolu: agreed & done

PR #937, opened 2021-01-19
Show Secrets as Table List

grolu:
We now use a Table to show the Secrets. This aligns the Secret page with the other pages.
It is also a preparation for the new Secret types that we plan to implement, e.g. DNS Provider Secrets.

Screen Shot 2021-01-25 at 12 15 36

Fixes #

With this PR we remove the **demo providers / disabled cloud providers** (Digital Ocean, China Telecom, Nutanix) from the code.

**Release note**:

```feature user
Secrets can now be filtered and are sortable using a table layout
```


frontend/src/views/Secrets.vue
```diff
@@ -213,40 +241,166 @@ export default {
...
+      switch (secret.metadata.cloudProviderKind) {
+        case 'openstack':
+          return [
+            {
+              label: 'Domain Name',
+              value: secretData.domainName
+            },
+            {
+              label: 'Tenant Name',
+              value: 'secretData.tenantName'
```
petersutter, line 271:
```suggestion
              value: secretData.tenantName
```

frontend/src/components/SecretRow.vue
```diff
@@ -5,112 +5,109 @@ SPDX-License-Identifier: Apache-2.0
...
+            <infra-icon v-on="on" v-model="item.infrastructure"></infra-icon>
+            <span class="ml-2">{{item.cloudProfileName}}</span>
           </div>
         </template>
-        <span v-if="!isOwnSecret">You can only delete secrets that are owned by you</span>
-        <span v-else-if="relatedShootCount > 0">You can only delete secrets that are currently unused</span>
-        <span v-else>Delete Secret</span>
+        <div><span class="font-weight-bold">Infrastructure:</span> {{item.infrastructure}}</div>
+        <div><span class="font-weight-bold">Cloud Profile:</span> {{item.cloudProfileName}}</div>
       </v-tooltip>
```
petersutter, line 30: align tooltip with the one on the shoot list

frontend/src/components/NewShoot/NewShootSelectInfrastructure.vue
```diff
@@ -12,11 +12,12 @@ SPDX-License-Identifier: Apache-2.0
       :class="{ 'select_infra_card_active elevation-4' : infrastructureKind == selectedInfrastructure }"
       @click="selectInfrastructure(infrastructureKind)"
       :key="infrastructureKind"
+      :color="darkMode ? 'grey darken-2' : undefined"
```
petersutter, line 15: in dark mode it is hard to tell which infrastructure is selected

frontend/src/components/TableColumnSelection.vue
```diff
@@ -7,10 +7,10 @@ SPDX-License-Identifier: Apache-2.0
 <template>
   <v-menu :nudge-bottom="20" :nudge-right="20" left v-model="columnSelectionMenu" absolute>
     <template v-slot:activator="{ on: menu }">
-      <v-tooltip open-delay="500" top>
+      <v-tooltip  top>
```
petersutter, line 10:
```suggestion
      <v-tooltip top>
```

frontend/src/components/VendorIcon.vue
```diff
@@ -5,33 +5,39 @@ SPDX-License-Identifier: Apache-2.0
...
+      type: Number,
+      default: 24
     },
-    height: {
-      type: Number
-    },
-    contentClass: {
-      type: String,
-      default: ''
+    noDarkModeBackground: {
```
petersutter, line 33: lightBackground?

frontend/src/components/VendorIcon.vue
```diff
@@ -5,33 +5,39 @@ SPDX-License-Identifier: Apache-2.0
...
 </template>
 
 <script>
 import startsWith from 'lodash/startsWith'
+import { mapState } from 'vuex'
 
 export default {
+  data () {
+    return {
+      backgroundDivSize: undefined
```
petersutter, line 22: rm

frontend/src/views/Secrets.vue
grolu: done

frontend/src/components/SecretRow.vue
grolu: Do you mean that I should use the Vendor component? Yes, that was my first idea but this tooltip shows shoot information (zone, region) .. so this does not possible (out of the box). It would be an option to pass the cloud information as parameters and not the shootItem but I don't know if reuse makes sense here as we show completely different information (`Infrastructure / Cloud Profile` vs `Infrastructure / Region / Zone`)

petersutter: No I was not saying to (re)use the component, just to mimic the same visual style

frontend/src/components/NewShoot/NewShootSelectInfrastructure.vue
grolu: I worked on the select infrastructure cards. Still not perfect, maybe we can have another look

frontend/src/components/TableColumnSelection.vue
grolu: done

frontend/src/components/VendorIcon.vue
grolu: done

frontend/src/components/VendorIcon.vue
grolu: renamed. default is false anyway?

frontend/src/components/SecretRow.vue
grolu: Hmm. But then we need to rethink this in general. I had a discussion with @holgerkoser a few weeks ago about the visual style of this tooltip does not match the other ones (white background).

petersutter: I changed it to white, otherwise the icons would not be nicely visible, but meanwhile we show a box around icons so I guess we can at least change the background

grolu: switched back to white and re-use on secrets list

petersutter:
just noted that now the secret help dialogs are not accessible anymore

frontend/src/views/Secrets.vue
```diff
@@ -6,127 +6,82 @@ SPDX-License-Identifier: Apache-2.0
...
-      infrastructure-name="Nutanix"
-      icon="mdi-xamarin"
-      description="Before you can provision and access a Kubernetes cluster on Nutanix, you need to add account credentials."
-        ></disabled-secret>
-
-    </template>
+      <v-card-text v-if="!secretList.length">
+        <div class="title grey--text text--darken-1 my-4">Add Secrets to your project.</div>
+        <p class="body-1">
+          Add Secrets to your project. You need to create at least one secret to be able to create a cluster.
```
petersutter, line 63: Screenshot 2021-01-29 at 16 11 09

> Before you can provision and access a Kubernetes cluster, you need to add account credentials.
\<CREATE\>

Maybe we should also hide the table when there are no secrets?

grolu: Yes, good point. I will implement it in the same way as we have it for the member section.

grolu:
>just noted that now the secret help dialogs are not accessible anymore

You are right, but how do you think best to integrate the help dialogs? We could work with tabs, something like this

Screen Shot 2021-02-01 at 12 26 51

WDYT?

grolu:
Secret Help side by side in secret dialog:
Screen Shot 2021-02-02 at 12 50 56

frontend/src/components/AccountAvatar.vue
```diff
@@ -5,7 +5,7 @@ SPDX-License-Identifier: Apache-2.0
 -->
 
 <template>
-  <div>
+  <div style="white-space:nowrap">
```
holgerkoser, line 8: move to class

holgerkoser, line 8:
```suggestion
  <div class="nowrap">
```

frontend/src/views/Secrets.vue
grolu: done

petersutter:
Improve text when no secrets are configured (taken from the aws help dialog)
```
Add Secrets to your project  (<--title)
Before you can provision and access a Kubernetes cluster, you need to add account credentials. The Gardener needs the credentials to provision and operate the AWS infrastructure for your Kubernetes cluster.
```

petersutter:
use the same style for `currently unused` as for `unknown` (greyed out)

petersutter:
We should also fix the inconsistent casing
Screenshot 2021-02-23 at 11 57 50
Screenshot 2021-02-23 at 11 58 56

petersutter:
when adding a metal secret the HMAC is always showing as unknown. Do we even want to show it as the input field is of type password

grolu:
>when adding a metal secret the HMAC is always showing as unknown. Do we even want to show it as the input field is of type password
You are right, the key is wrong. As we cannot show anything here (even API URL ist starred out) I removed it for metal

frontend/src/components/SecretRow.vue
```diff
@@ -5,112 +5,113 @@ SPDX-License-Identifier: Apache-2.0
...
+        <v-tooltip top v-if="canPatchSecrets">
+          <template v-slot:activator="{ on }">
+            <div v-on="on">
+              <v-btn :disabled="!item.isOwnSecret || !item.isSupportedInfrastructure" icon @click="onUpdate">
+                <v-icon class="action-button--text">mdi-pencil</v-icon>
+              </v-btn>
+            </div>
+          </template>
+          <span v-if="!item.isOwnSecret">You can only edit secrets that are owned by you</span>
+          <span v-else-if="!item.isSupportedInfrastructure">This infrastructure type is currently not supported by the dashboard</span>
```
petersutter, line 60:
```js
          <span v-else-if="!item.isSupportedInfrastructure">This infrastructure type is currently not supported by the dashboard</span>
```
this is not a limitation of the dashboard itself but rather a configuration issue, or not?

grolu: If an extension for an infrastructure kind is installed on a landscape that the Dashboard is not aware of / has not yet implemented, then secrets (and also clusters) of this infrastructure still appear in the dashboard. It happened on our dev landscapes, for instance. While viewing cluster details works fine (most of the time), editing such a secret is not possible as we do not know which fields we need to maintain.

petersutter:
```js
isSupportedInfrastructure: includes(this.sortedCloudProviderKindList, secret.metadata.cloudProviderKind)
```
nevermind, I confused `sortedCloudProviderKindList` with the `cloudProfiles`

frontend/src/components/SecretRow.vue
```diff
@@ -5,112 +5,113 @@ SPDX-License-Identifier: Apache-2.0
...
+  <tr>
+    <td v-if="selectedHeaders.name">
+      <div class="d-flex">
+        {{item.name}}
+        <v-tooltip v-if="!item.isOwnSecret" top>
           <template v-slot:activator="{ on }">
             <v-icon v-on="on" small class="mx-1">mdi-account-arrow-left</v-icon>
           </template>
-          <span>Secret shared by {{secretOwner}}</span>
+          <span>Secret shared by {{item.ownernamespace}}</span>
```
holgerkoser, line 16: `ownernamespace` should be camelCase `ownerNamespace`

frontend/src/components/Vendor.vue
```diff
@@ -62,30 +76,48 @@ export default {
       default: false
     }
   },
-  mixins: [shootItem],
   computed: {
-    shootZonesText () {
-      return join(this.shootZones, ', ')
+    zonesText () {
```
holgerkoser, line 80: zoneTitle but ZonesText

holgerkoser: rename to zoneText

frontend/src/components/VendorIcon.vue
```diff
@@ -65,17 +62,12 @@ export default {
...
-    getWidth () {
-      if (!this.width && !this.height) {
-        return 20
-      }
-      return this.width
-    },
     isMdiIcon () {
       return startsWith(this.value, 'mdi-')
+    },
+    iconStyle () {
```
holgerkoser, line 68: Use classes if possible. Static attributes should be static not bound. Use Object Syntax (inline if possible) not String syntax https://vuejs.org/v2/guide/class-and-style.html#Object-Syntax-1

frontend/src/components/dialogs/SecretDialog.vue
```diff
@@ -198,6 +208,14 @@ export default {
     shootsByInfrastructureSecret () {
       const name = get(this.secret, 'metadata.name')
       return filter(this.shootList, ['spec.secretBindingName', name])
+    },
+    helpStyle () {
```
holgerkoser, line 212: Use classes if possible. Static attributes should be static not bound. Use Object Syntax (inline if possible) not String syntax https://vuejs.org/v2/guide/class-and-style.html#Object-Syntax-1

frontend/src/views/Secrets.vue
```diff
@@ -188,21 +149,91 @@ export default {
...
+      }))
+    },
+    visibleSecretTableHeaders () {
+      return filter(this.secretTableHeaders, ['selected', true])
+    },
+    secretList () {
+      return map(this.infrastructureSecretList, secret => ({
+        name: secret.metadata.name,
+        isOwnSecret: isOwnSecret(secret),
+        ownernamespace: secret.metadata.secretRef.namespace,
```
holgerkoser, line 227: camelCase

frontend/src/components/SecretRow.vue
grolu: done

frontend/src/components/Vendor.vue
grolu: done

frontend/src/views/Secrets.vue
```diff
@@ -188,21 +149,91 @@ export default {
...
+        infrastructure: `${secret.metadata.cloudProviderKind}${secret.metadata.cloudProfileName}`,
```
petersutter, line 228: `infrastructure` acts as key, or? I guess we should add a delimiter, like `--`

frontend/src/components/VendorIcon.vue
grolu: done

frontend/src/components/dialogs/SecretDialog.vue
grolu: done

frontend/src/views/Secrets.vue
grolu: done

frontend/src/views/Secrets.vue
grolu: `infrastructure`is used as value for the _Infrastructure_ column. The value is used for the column sorting by the data table

petersutter: still, to be on the safe side I would add the delimiter, even though a name clash is highly unlikely

frontend/src/views/Secrets.vue
```diff
@@ -6,127 +6,83 @@ SPDX-License-Identifier: Apache-2.0
...
-      <disabled-secret
-      class="mt-4"
-      infrastructure-name="Nutanix"
-      icon="mdi-xamarin"
-      description="Before you can provision and access a Kubernetes cluster on Nutanix, you need to add account credentials."
-        ></disabled-secret>
-
-    </template>
+      <v-card-text v-if="!secretList.length">
+        <div class="title grey--text text--darken-1 my-4">Add Secrets to your project.</div>
```
petersutter, line 61:
```suggestion
        <div class="title grey--text text--darken-1 my-4">Add Secrets to your project</div>
```

frontend/src/views/Secrets.vue
```diff
@@ -6,127 +6,83 @@ SPDX-License-Identifier: Apache-2.0
...
+        <p class="body-1">
+          Before you can provision and access a Kubernetes cluster, you need to add account credentials. The Gardener needs the credentials to provision and operate the infrastructure for your Kubernetes cluster.
```
petersutter, line 63:
```suggestion
          Before you can provision and access a Kubernetes cluster, you need to add infrastructure account credentials. The Gardener needs the credentials to provision and operate the infrastructure for your Kubernetes cluster.
```

petersutter:
when changing changing the secret table options, e.g. when I hide columns and press reload, the changes are reset

frontend/src/components/VendorIcon.vue
```diff
@@ -67,7 +67,11 @@ export default {
     },
     iconStyle () {
       const maxIconSize = this.size - 4
-      return `max-height: ${maxIconSize}px; max-width: ${maxIconSize}px; border-radius: 0px;`
+      return {
+        maxHeight: `${maxIconSize}px`,
+        maxWidth: `${maxIconSize}px`,
+        borderRadius: '0px'
```
holgerkoser, line 73: add a class `no-border-radius` to main.scss

frontend/src/components/dialogs/SecretDialog.vue
```diff
@@ -215,7 +215,11 @@ export default {
       if (detailsRef) {
         detailsHeight = detailsRef.getBoundingClientRect().height
       }
-      return `max-width: 80%; max-height: ${detailsHeight}px; overflow-y: scroll;`
+      return {
+        maxWidth: '80%',
+        maxHeight: `${detailsHeight}px`,
+        overflowY: 'scroll'
```
holgerkoser, line 221: define maxWidth and overflow statically via css.

frontend/src/views/Secrets.vue
grolu: I do not fully understand what you want to achieve by adding a delimiter. How does this make the value more unique. Other than that, multiple entries with the same value are likely. However, I agree that sorting is not perfect, it would be better to search for `cloudProviderKind` and then for `cloudProfileName` and not combined. However we need to implement a custom search function in that case

frontend/src/components/dialogs/SecretDialog.vue
grolu: done

frontend/src/components/VendorIcon.vue
grolu: done, used already defined `rounded-0` class

holgerkoser (approved):
\lgtm

petersutter (approved):
/lgtm however I still can't properly see which infrastructure is selected on the create cluster page with dark mode

grolu:
> /lgtm however I still can't properly see which infrastructure is selected on the create cluster page with dark mode

Ok, let's discuss this in the theming PR where we also fine-tune the colors

PR #941, opened 2021-01-27
Bug/fix shoot table sorting

### LINKED ISSUE #903: Shoot Table does not use v-data-table API correctly
The table does not use the `custom-sort` (https://vuetifyjs.com/en/api/v-data-table/#custom-sort) method for the custom sort implementation. Instead, the data is sorted before it is passed to the table. This currently works, however probably only because of timing and the content gets sorted twice. We should switch to the correct implementation, see also https://github.com/gardener/dashboard/pull/872

grolu:

Fixes #903

**Release note**:

```bugfix user
Fixed: Persist cluster sort order correctly when sorting by an non custom field if cluster list was sorted by a custom field before
```


frontend/src/utils/shootList.js
```diff
@@ -0,0 +1,222 @@
...
+    }
+
+    return some(searchableCustomFields, ({ key }) => includes(getRawVal(storeGetters, item, key), value))
+  })
+}
+
+export function sortShoots (storeGetters, items, sortByArr, sortDescArr) {
+  const sortBy = head(sortByArr)
+  const sortOrder = head(sortDescArr) ? 'desc' : 'asc'
+  if (sortBy) {
```
petersutter, line 71: early exit
```js
  if (!sortBy) {
    return items
  }
  ...
```

frontend/src/views/ShootList.vue
```diff
@@ -222,6 +220,22 @@ export default {
...
+      }
+      return searchShoots(storeGetters, value, search, item)
+    },
+    sortShoots (items, sortByArr, sortDescArr) {
+      const storeGetters = {
+        ticketsLabels: this.ticketsLabels,
+        shootCustomFields: this.shootCustomFields,
+        latestUpdatedTicketByNameAndNamespace: this.latestUpdatedTicketByNameAndNamespace
+      }
+      return sortShoots(storeGetters, items, sortByArr, sortDescArr)
```
holgerkoser, line 238: I would implement these two functions:

Either as getters in the shoots store module
```js
filterItems (state, getters, rootState, rootGetters) {
  return (value, search, item) => {
    ...
  }
},
sortItems (state, getters, rootState, rootGetters) {
  return (value, search, item) => {
    ...
  }
},
```

or as component mixin `ShootDataTable` with access to the mixins.

frontend/src/utils/shootList.js
grolu: done

frontend/src/store/modules/shoots/getters.js
```diff
@@ -0,0 +1,247 @@
...
+      if (startsWith(column, 'Z_')) {
+        const path = get(rootGetters.shootCustomFields, [column, 'path'])
+        return get(item, path)
+      }
+      return metadata[column]
+    }
+  }
+}
+
+const getSortVal = (rootGetters, item, sortBy) => {
```
holgerkoser, line 64:
```suggestion
function getSortVal (rootGetters, item, sortBy) {
```

frontend/src/store/modules/shoots/getters.js
```diff
@@ -0,0 +1,247 @@
...
+import {
+  getCreatedBy,
+  getProjectName,
+  isShootStatusHibernated,
+  isReconciliationDeactivated
+} from '@/utils'
+import { findItem } from './helper'
+import { isUserError, errorCodesFromArray } from '@/utils/errorCodes'
+
+const getRawVal = (rootGetters, item, column) => {
```
holgerkoser, line 30:
```suggestion
function getRawVal (rootGetters, item, column) {
```

frontend/src/store/modules/shoots/helper.js
```diff
@@ -0,0 +1,18 @@
+//
+// SPDX-FileCopyrightText: 2020 SAP SE or an SAP affiliate company and Gardener contributors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+const keyForShoot = ({ name, namespace }) => {
```
holgerkoser, line 7:
```suggestion
export function keyForShoot ({ name, namespace }) {
```

frontend/src/store/modules/shoots/helper.js
```diff
@@ -0,0 +1,18 @@
...
+  return `${name}_${namespace}`
+}
+
+const findItem = state => {
```
holgerkoser, line 11:
```suggestion
export function findItem (state) {
```

frontend/src/store/modules/shoots/helper.js
```diff
@@ -0,0 +1,18 @@
...
+  return ({ name, namespace }) => state.shoots[keyForShoot({ name, namespace })]
+}
+
+module.exports = {
+  keyForShoot,
+  findItem
+}
```
holgerkoser, line 18:
```suggestion
```

frontend/src/store/modules/shoots/index.js
```diff
@@ -12,33 +12,23 @@ import omit from 'lodash/omit'
...
-import split from 'lodash/split'
-import join from 'lodash/join'
 import set from 'lodash/set'
 import head from 'lodash/head'
 import sample from 'lodash/sample'
 import isEmpty from 'lodash/isEmpty'
 import cloneDeep from 'lodash/cloneDeep'
-import semver from 'semver'
-import store from '../'
+import store from '../../'
```
holgerkoser, line 23: brauchen wir den store hier noch? Können wir das nicht auch gleich los werden?

frontend/src/store/modules/shoots/getters.js
```diff
@@ -0,0 +1,247 @@
...
+  onlyShootsWithIssues (state, getters) {
+    return get(getters.getShootListFilters, 'onlyShootsWithIssues', true)
+  },
+  newShootResource (state) {
+    return state.newShootResource
+  },
+  initialNewShootResource (state) {
+    return state.initialNewShootResource
+  },
+  searchShoots (state, getters, rootState, rootGetters) {
```
holgerkoser, line 154: I think we should use a generic naming convention inside a module or class instead of repeating the name of the module or class in each and every method or property. We already do this e.g in `filteredItems`, `itemByNameAndNamespace` and `selectedItem`, `ITEM_PUT`, `HANDLE_EVENTS`, ... Each store module should have a similar structure and a generic naming convention. You can see this concept in the structure of a kubernetes resource with `kind`, `metadata`, `spec`, `items`, `status` properties. My proposal has been to name these new getter functions `searchItems` and `sortItems`.

frontend/src/store/modules/shoots/index.js
holgerkoser: Ok. We still need the store in the `HANDLE_EVENTS` mutation. We should not remove it in this PR.

frontend/src/store/modules/shoots/getters.js
```diff
@@ -0,0 +1,247 @@
...
+  itemByNameAndNamespace (state) {
+    return ({ namespace, name }) => {
+      return findItem(state)({ name, namespace })
+    }
+  },
+  selectedItem (state) {
+    if (state.selection) {
+      return findItem(state)(state.selection)
+    }
+  },
```
holgerkoser, line 141:
```suggestion
  selectedItem (state, getters) {
    if (state.selection) {
      return getters.itemByNameAndNamespace(state.selection)
    }
  },
```

frontend/src/store/modules/shoots/getters.js
```diff
@@ -0,0 +1,247 @@
...
+}
+
+export default {
+  filteredItems (state) {
+    return state.filteredShoots
+  },
+  itemByNameAndNamespace (state) {
+    return ({ namespace, name }) => {
+      return findItem(state)({ name, namespace })
+    }
```
holgerkoser, line 135:
```suggestion
    return findItem(state)
```

frontend/src/store/modules/shoots/getters.js
```diff
@@ -0,0 +1,247 @@
...
+    case 'ticket': {
+      const { namespace, name } = item.metadata
+      return rootGetters.latestUpdatedTicketByNameAndNamespace({ namespace, name })
+    }
+    default:
+      return toLower(value)
+  }
+}
+
+export default {
```
holgerkoser, line 128: I would use both [named export](https://developer.mozilla.org/en-us/docs/Web/JavaScript/Reference/Statements/export#using_named_exports) and [default export](https://developer.mozilla.org/en-us/docs/Web/JavaScript/Reference/Statements/export#using_the_default_export).

```js
export function filteredItems (state) {
    return state.filteredShoots
}
export function itemByNameAndNamespace (state) {
   return findItem(state)
}
...
export default {
  filteredItems,
  itemByNameAndNamespace,
  ... 
}

or

function filteredItems (state) {
    return state.filteredShoots
}
function itemByNameAndNamespace (state) {
   return findItem(state)
}
export {
  getRawVal,
  getSortVal,
  filteredItems,
  itemByNameAndNamespace,
  ... 
}
export default {
  filteredItems,
  itemByNameAndNamespace,
  ... 
}
```

In the named exports we could export all functions also `getRawVal` and `getSortVal` e.g. makes it easier to test them. But in the default export only those which follow the getter signature and could / should be added to the store.

frontend/src/store/modules/shoots/getters.js
grolu: ok

holgerkoser (approved):
/lgtm

PR #947, opened 2021-02-11
Removed moment & moment timezone libraries

### LINKED ISSUE #944: Timezone selection in Maintenance schedule is ignored - we are all living in Africa/Abidjan   now
Changing the Timezone of the maintenance schedule of a cluster in the gardener webUI  has  no effect. 
Lifecycle
Hibernation
Waking up Cluster...
**Maintenance**
Start time: 01:00 **Africa/Abidjan**

Maybe also it would be meaningful to clarify how the maintenance happens when the cluster is hibernating?!?
Or if that is not working maybe the UI should try to limit the maintenance schedule to the wake times?

grolu:
We removed moment and moment timezone libraries from our frontend. The main reason is to reduce (frontend) build size as we use only a very small subset of the library. We implemented these functions on our own and now have full control. Main changes are:
- Added replacements for moment time / date functions (humanize)
- Added list of locations (moment.tz.names)
- Maintenance: removed local timezone conversions  / do not use location / always use timezones
- Hibernation: Do not try to convert to local timezone if no location configured, use UTC instead

Fixes #944

**Release note**:

```feature user
We do no longer use locations to show the maintenance start time. We now use timezones to show and configure the maintenance start time as this reflects the yaml representation. We also do no longer convert the time into local browser time but rather show and store the timezone offset to UTC
```


frontend/src/utils/time.js
```diff
@@ -0,0 +1,867 @@
...
+//
+
+import includes from 'lodash/includes'
+import sample from 'lodash/sample'
+import join from 'lodash/join'
+import padStart from 'lodash/padStart'
+import map from 'lodash/map'
+import takeWhile from 'lodash/takeWhile'
+
+export const locations = [
```
holgerkoser, line 14: move locations and timezones in separate files. Create a folder for time and move this into index.

grolu: moved to dedicated files

grolu:
I checked the dayjs library. As far as I can tell, the library does not offer options to retrieve the list of locations and / or timezones. So I guess we need to keep these files in our code. Regarding the other features that dayjs offers, the most interesting thing is the duration plugin which offers the features I implemented with the `humanizeDuration`method. However I don't know if we want to incorporate the library and plugin only for 100 lines of code.

grolu:
Closed this PR in favour for https://github.com/gardener/dashboard/pull/950
We will replace the moment library in a separate request

PR #959, opened 2021-03-01
Increase terminal  timeout

petersutter:
In some cases the terminal-controller-manager takes more time than usual to setup all the required resources and update the terminal resource, e.g. due to heavy load on the host or target cluster or the terminal-controller-manager itself. That's why often times the 10 second timeout is reached and the terminal creation is aborted with an error.
This PR increases the 10 second timeout to wait for the terminal resource to be in a "ready" state to 60 seconds.

Fixes #

**Release note**:

```other user

```


grolu (approved):
/lgtm

holgerkoser (approved):
/lgtm

PR #966, opened 2021-03-04
Use raw value if html color code

petersutter:
Fixes connected icon color of terminal

Issue:
Screenshot 2021-03-04 at 11 20 09

Fixes #

**Release note**:

```bugfix user

```

### DIFF
--- a/frontend/src/components/icons/IconBase.vue
+++ b/frontend/src/components/icons/IconBase.vue
@@ -24,6 +24,7 @@ SPDX-License-Identifier: Apache-2.0
 </template>
 
 <script>
+import { isHtmlColorCode } from '@/utils'
 
 export default {
   props: {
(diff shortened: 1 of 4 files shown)


grolu (approved):
/lgtm

frontend/src/components/icons/IconBase.vue
```diff
@@ -51,6 +51,9 @@ export default {
   computed: {
     iconColorCode () {
       const iconColor = this.iconColor
+      if (/^#([a-fA-F0-9]{6}|[a-fA-F0-9]{3})$/.test(iconColor)) {
```
holgerkoser, line 54:
```suggestion
      if (/^#([a-f0-9]{6}|[a-f0-9]{3})$/i.test(iconColor)) {
```

holgerkoser: case insensitive regex

holgerkoser (approved):
/lgtm

frontend/src/components/icons/IconBase.vue
grolu: then we should do it in all places or even better move the regex to utils

petersutter: moving it to utils makes sense

holgerkoser (approved):
/lgtm

PR #970, opened 2021-03-08
Removed unused color properties

grolu:
Removed unused color properties from dialogs. We do not need them anymore because we now style all dialogs with toolbar-background color.

Fixes #

**Release note**:

```bugfix user

```


petersutter (approved):
/lgtm

PR #978, opened 2021-03-18
Detect Dark Mode Settings

grolu:
With this PR the Dashboard can detect the system settings for dark mode

Screenshot 2021-03-22 at 13 26 03

Fixes #

**Release note**:

```feature user
Dark Mode: The Dashboard now applies system settings by default
```


frontend/src/components/MainToolbar.vue
```diff
@@ -104,8 +104,17 @@ SPDX-License-Identifier: Apache-2.0
           <v-divider></v-divider>
           <v-card-actions class="px-3">
             <v-icon color="primary" class="ml-2 mr-3">mdi-brightness-6</v-icon>
-            <v-switch v-model="darkMode" dense hide-details class="ma-0" color="primary" @click.native.stop></v-switch>
-            <span class="primary--text text-button">Dark</span>
+            <v-btn-toggle v-model="darkMode" borderless mandatory @click.native.stop>
```
petersutter, line 107: add tooltips

grolu: ah yes I forgot that actually I wanted to do this

frontend/src/store/index.js
```diff
@@ -1498,9 +1498,25 @@ const mutations = {
     state.splitpaneResize = value
   },
   SET_DARK_MODE (state, value) {
-    state.darkMode = value
-    localStorage.setItem('global/dark-mode', value)
-    Vue.vuetify.framework.theme.dark = value
+    if (value !== undefined) {
```
holgerkoser, line 1501: From my point of view this is an antipattern. A mutation should only mutate the store state and not run mediaqueries.  The new value must come from outside the mutation. Why do we need two properties `darkMode` and `darkTheme`.

holgerkoser: We should start using store plugins for this kind of problems.
```js
export default function createMediaPlugin (window) {
  return store => {
    const mq = window.matchMedia('(prefers-color-scheme: dark)')
    store.commit('SET_PREFERS_COLOR_SCHEME', mq.matches ? 'dark' : 'light')
    mq.addListener(e => {
      store.commit('SET_PREFERS_COLOR_SCHEME', e.matches ? 'dark' : 'light')
    })
  }
}
```

frontend/src/app.js
```diff
@@ -21,8 +21,13 @@ const App = Vue.extend({
       }
     })
 
-    const darkMode = this.$localStorage.getItem('global/dark-mode') === 'true'
+    const darkMode = this.$localStorage.getItem('global/setting-dark-mode') || 2
```
holgerkoser, line 24: Why `global/seeting-dark-mode` and not `global/dark-mode` or `global/settings/dark-mode`?

holgerkoser: I would propose to have a localStorage porperty 'global/color-scheme' or  'global/settings/color-scheme' with possible values `dark`, `light` and `auto` or `no-preference`

frontend/src/components/MainToolbar.vue
grolu: done

frontend/src/components/MainToolbar.vue
```diff
@@ -92,6 +92,32 @@ SPDX-License-Identifier: Apache-2.0
...
+                  </template>
+                  <span>Dark Mode</span>
+                </v-tooltip>
+                <v-tooltip top>
+                  <template v-slot:activator="{ on }">
+                    <v-btn small v-on="on">
+                      <v-icon color="primary">mdi-brightness-auto</v-icon>
+                    </v-btn>
+                  </template>
+                  <span>Auto</span>
```
petersutter, line 118:
```suggestion
                  <span>Automatically choose theme based on your system settings</span>
```

frontend/src/components/MainToolbar.vue
```diff
@@ -199,12 +220,29 @@ export default {
...
+          case 0:
+            this.setColorScheme('light')
+            break
+          case 1:
+            this.setColorScheme('dark')
+            break
+          default:
+            this.setColorScheme('auto')
+            break
+        }
```
petersutter, line 245: use constants instead of magic/index numbers

grolu: done

petersutter (approved):
/lgtm

holgerkoser (approved):
/lgtm

PR #979, opened 2021-03-22
Fixed autoscaler min / max value not showing up in worker description

grolu:
Fixed autoscaler min / max  or surge value not showing up in worker description if value is zero

Fixes #

**Release note**:

```bugfix user
Fixed: Autoscaler min / max  or surge value not showing up in worker description if one of the values is zero
```


holgerkoser (approved):
/lgtm

PR #980, opened 2021-03-24
Chart assets

morremeyer:

This PR adds asset configuration to the helm chart

**Release note**:

```feature operator
Enable asset configuration in the helm chart
```

### DIFF
--- a/docs/Theming.md
+++ b/docs/Theming.md
@@ -22,28 +22,34 @@ The following colors can be configured:
 | `warning`               | Snotify warning popups, warning texts |
 | `error`                 | Snotify error popups, error texts |
 
+If you use the helm chart, you can configure those with `frontendConfig.themes.light` for the light theme and `frontendConfig.themes.dark` for the dark theme.
+
 ## Logos and Icons
-It is also possible to exchange the Dashboard logo and icons. You can replace the [assets](https://github.com/gardener/dashboard/tree/master/frontend/public/static/assets) folder via a volume mount in your deployment.
 
-The simplest solution is to create a configmap and add the assets to it:
-```
-apiVersion: v1
-kind: ConfigMap
-metadata:
-  name: dashboard-assets
-data:
-  logo.svg: |
-    ...
+It is also possible to exchange the Dashboard logo and icons. You can replace the [assets](https://github.com/gardener/dashboard/tree/master/frontend/public/static/assets) folder when using the [helm chart](https://github.com/gardener/dashboard/blob/master/charts/gardener-dashboard) in the `frontendConfig.assets` map.
+
+Attention: You need to set values for all files as mapping the volume will overwrite all files. It is not possible to exchange single files.
+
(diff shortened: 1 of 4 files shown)


gardener-robot-ci-1:
Thank you @morremeyer for your contribution. Before I can start building your PR, a member of the organization must set the required label(s) {'reviewed/ok-to-test'}. Once started, you can check the build status in the PR checks section below.

grolu:
Hi @morremeyer,
thank you for your pull request. It seems that investing in this topic is appreciated by the community. Integrating the asset volume mount into the helm charts makes absolutely sense from our point of view.
Just some small remarks:
I think it would be better to not include the full base64 examples for the images, just include the filenames and maybe some information about how to create the data for the images in the comment.
Moreover, it is great that you adapted the Theming documentation. I think you can also remove the section about how to create the config map and mount it completely from the documentation and just refer to the helm chart.

morremeyer:
@grolu I was thinking about how to avoid putting the data in the file already, thanks for the suggestion.

I’ll incorporate your suggestions :)

And yes, investment in this topic is definitely appreciated, at least by us.

morremeyer:
@grolu I created a snippet that can be copy-pasted to generate the default values and liked that in the documentaiton. I’m much more happy with that solution compared to the first one.

docs/Theming.md
```diff
@@ -22,28 +22,34 @@ The following colors can be configured:
...
+
+```bash
+cat << EOF
+  ###
+  ### COPY EVERYTHING BELOW THIS LINE
+  ###
+
+  assets:
+    favicon-16x16.png: |
+      $(cat frontend/public/static/assets/favicon-16x16.png | base64 -w0)
```
grolu, line 43: -w0 does not work on macOS. Actually, it is the default mode in case of macOS. We should document this here

holgerkoser: For me it works because I am `base64` from [`coreutils`](https://formulae.brew.sh/formula/coreutils) on macOS
```bash
$ brew install coreutils
```

petersutter: In any case it should be documented how it can be run on MacOS :) I would vote for recommending to install `coreutils`

morremeyer: I agree with @petersutter, will add this tomorrow.

morremeyer: I added 

> If you use macOS, install coreutils with brew (`brew install coreutils`) or remove the `-w0` parameter.

petersutter (approved):
/lgtm

however I haven't tested the chart deployment and only rendered the template with and without assets, which looks good

grolu (approved):
I also didn't test the deployment. @morremeyer Did you test this end-to-end? If not, I will do a quick test tomorrow. Otherwise, /lgtm from my side

morremeyer:
@grolu Yes, I did test this E2E in one of our test setups.

PR #982, opened 2021-03-26
Update charts

poelzi:

fixes the ingress helm chart object to work with non ngix ingress and no tls

Fixes #

**Release note**:

```bugfix user

```


gardener-robot-ci-1:

charts/gardener-dashboard/templates/ingress.yaml
```diff
@@ -31,5 +33,5 @@ spec:
       - backend:
           serviceName: gardener-dashboard-service
           servicePort: {{ $.Values.servicePort }}
-        path: /
+        path: /*
```
petersutter, line 36: I guess it makes sense to have the path configurable in the `values.yaml` instead of hardcoding it here, WDYT?

poelzi: not really. According to official ingress documentation the previous path only matches the root page and no subpages, a phenomena I see with isio-ingress. kubernetes-nginx seems more relaxed then other ingress providers.

petersutter: I guess the problem is rather, that we do not define the `pathType` https://kubernetes.io/docs/concepts/services-networking/ingress/#path-types

By default, `ImplementationSpecific` is used
> `ImplementationSpecific`: With this path type, matching is up to the IngressClass. Implementations can treat this as a separate pathType or treat it identically to Prefix or Exact path types.

However instead we should rather use `pathType: Prefix` and keep the path with `/` as is

> `Prefix`: Matches based on a URL path prefix split by /. Matching is case sensitive and done on a path element by element basis. A path element refers to the list of labels in the path split by the / separator. A request is a match for path p if every p is an element-wise prefix of p of the request path.

Would that solve your issue?

poelzi: good catch, i changed the type to prefix

petersutter (approved):
/lgtm

holgerkoser (approved):
/lgtm

grolu (approved):

PR #989, opened 2021-04-08
TLS settings update. Support CA environment / cert-manager

poelzi:

Allows CA to be passed by a secret (environment variable).
This is very useful for test environments where the certificates are managed by cert-manager.
Smaller cleanup: rename tlsSecretName to tls.secretName (I guess it is not used, at least was not documented, but in theory a breaking change in helm charts.)
Add a proper tls.enabled flag to enable/disable tls
Allows to enable tls but not provide a key/cert (using cert-manager)

Fixes #

**Release note**:

```bugfix user
- category:       breaking
- category: feature
```

### DIFF
--- a/backend/test/config.spec.js
+++ b/backend/test/config.spec.js
@@ -74,8 +74,7 @@ describe('config', function () {
         OIDC_ISSUER: 'issuer',
         OIDC_CLIENT_ID: 'client_id',
         OIDC_CLIENT_SECRET: 'client_secret', // pragma: whitelist secret
-        OIDC_REDIRECT_URI: 'redirect_uri',
-        OIDC_CA: 'my_ca'
+        OIDC_REDIRECT_URI: 'redirect_uri'
       }
 
       beforeEach(() => {
@@ -84,12 +83,22 @@ describe('config', function () {
 
       it('should return the config in test environment', function () {
         const env = Object.assign({
-          NODE_ENV: 'test'
+          NODE_ENV: 'test',
+          OIDC_CA: 'ca'
         }, requiredEnvironmentVariables)
 
         const config = gardener.loadConfig(undefined, { env })
         const defaults = gardener.getDefaults({ env })
         expect(config).toMatchObject(defaults)
+      })
+
+      it('should return optional environment variables', function () {
+        const env = Object.assign({
+          OIDC_CA: 'ca'
+        }, requiredEnvironmentVariables)
+
+        const config = gardener.loadConfig(undefined, { env })
+        expect(config.oidc.ca).toBe('ca')
       })
 
       it('should return the config in production environment', function () {
@@ -103,7 +112,6 @@ describe('config', function () {
         expect(gardener.readConfig.mock.calls[0]).toEqual([filename])
         expect(config.port).toBe(1234)
         expect(config.logLevel).toBe('warn')
-        expect(config.oidc.ca).toBe('my_ca')
       })
 
       it('should return the config with port and logLevel overridden by environment variables', function () {
--- a/backend/test/config.spec.js
+++ b/backend/test/config.spec.js
@@ -74,7 +74,8 @@ describe('config', function () {
         OIDC_ISSUER: 'issuer',
         OIDC_CLIENT_ID: 'client_id',
         OIDC_CLIENT_SECRET: 'client_secret', // pragma: whitelist secret
-        OIDC_REDIRECT_URI: 'redirect_uri'
+        OIDC_REDIRECT_URI: 'redirect_uri',
+        OIDC_CA: 'my_ca'
       }
 
       beforeEach(() => {
@@ -102,6 +103,7 @@ describe('config', function () {
         expect(gardener.readConfig.mock.calls[0]).toEqual([filename])
         expect(config.port).toBe(1234)
         expect(config.logLevel).toBe('warn')
+        expect(config.oidc.ca).toBe('my_ca')
       })
 
       it('should return the config with port and logLevel overridden by environment variables', function () {
--- a/charts/gardener-dashboard/templates/deployment.yaml
+++ b/charts/gardener-dashboard/templates/deployment.yaml
@@ -105,6 +105,12 @@ spec:
               secretKeyRef:
                 name: gardener-dashboard-oidc
                 key: client_secret
+          {{- if .Values.oidc.caSecret -}}
+          - name: OIDC_CA
+            valueFrom:
+              secretKeyRef:
+                name: {{ .Values.oidc.caSecret }}
+                key: {{ .Values.oidc.caSecretKey | default "ca.crt" }}
           {{- end }}
           {{- if .Values.gitHub }}
           - name: GITHUB_AUTHENTICATION_USERNAME
--- a/charts/gardener-dashboard/templates/ingress.yaml
+++ b/charts/gardener-dashboard/templates/ingress.yaml
@@ -19,7 +19,7 @@ metadata:
 spec:
 {{- if .Values.tls.enabled }}
   tls:
-  - secretName: {{ .Values.tlsSecretName | default "gardener-dashboard-tls" }}
+  - secretName: {{ .Values.tls.secretName | default "gardener-dashboard-tls" }}
     hosts:
     {{- range .Values.hosts }}
     - {{ . }}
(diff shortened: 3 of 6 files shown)


gardener-robot-ci-3:

backend/test/config.spec.js
```diff
@@ -74,7 +74,8 @@ describe('config', function () {
         OIDC_ISSUER: 'issuer',
         OIDC_CLIENT_ID: 'client_id',
         OIDC_CLIENT_SECRET: 'client_secret', // pragma: whitelist secret
-        OIDC_REDIRECT_URI: 'redirect_uri'
+        OIDC_REDIRECT_URI: 'redirect_uri',
+        OIDC_CA: 'my_ca'
```
holgerkoser, line 78:
```suggestion
        OIDC_CA: 'ca'
```

holgerkoser, line 78:
```suggestion
        OIDC_REDIRECT_URI: 'redirect_uri'
```
OIDC_CA is not a required environment variable!

backend/test/config.spec.js
```diff
@@ -102,6 +103,7 @@ describe('config', function () {
         expect(gardener.readConfig.mock.calls[0]).toEqual([filename])
         expect(config.port).toBe(1234)
         expect(config.logLevel).toBe('warn')
+        expect(config.oidc.ca).toBe('my_ca')
```
holgerkoser, line 106:
```suggestion
```
See comment above!

poelzi: done

holgerkoser:
You already committed the changes regarding my comment 😄. I am still trying to understand the motivation behind the changes you are proposing. I think the PR contains two independent changes:

1. The change which should allow to read the OIDC certificate authority from a secret which is not part of the dashboard helm chart.
2. The change which should allow to create an ingress with a tls configuration but does not create the tls secret. The tls secret could be created by the cert manager.

I understand the second use case and I think we should support it. 

But you need to explain why it is necessary to read the oidc from a secret. The ca of the oidc provider is not secret so we can have this information written into to configmap. I don't see a reason why we would need an indirection here. When the dashboard is deployed the oidc ca should be known. It cannot be injected at runtime or do I miss something? 
 
Would it be possible to split these two use cases into two separate PRs?

holgerkoser:
If the oidc certificate is a Let's Encrypt production environment certificate we do not need a ca I think. If it is a Let's Encrypt staging environment certificate we need the full chain I think. As far as I know the `ca.cert` in the secret does only contain the intermediate certificate. Have you tested it with staging or production environment?

poelzi:
we have a automatic test environment which uses cert-manager to create a local ca that signes all certificates. It does not use lets encrypt as the test environment is local only. With this simple change, the ca cert can be copied into a secret and is checked against. otherwise we would need to extract the ca, create a custom config map and merge this with helm somehow which is rather inconvenient.

charts/gardener-dashboard/templates/deployment.yaml
```diff
@@ -105,6 +105,12 @@ spec:
               secretKeyRef:
                 name: gardener-dashboard-oidc
                 key: client_secret
+          {{- if .Values.oidc.caSecret -}}
+          - name: OIDC_CA
+            valueFrom:
+              secretKeyRef:
+                name: {{ .Values.oidc.caSecret }}
+                key: {{ .Values.oidc.caSecretKey | default "ca.crt" }}
```
holgerkoser, line 113:
```suggestion
          {{- if .Values.oidc.caSecretKeyRef }}
          - name: OIDC_CA
            valueFrom:
              secretKeyRef:
                name: {{ .Values.oidc.caSecretKeyRef.name }}
                key: {{ .Values.oidc.caSecretKeyRef.key | default "ca.crt" }}
```

charts/gardener-dashboard/values.yaml
```diff
@@ -63,6 +65,11 @@ oidc:
   clientSecret: ~
   # redirectUri is the endpoint receiving the OTAC from the OpenID provider after successful login
   redirectUri: https://dashboard.ingress.example.org/auth/callback
+  # ca as a string to use
+  # ca:
+  # ca as a secret to use
+  # caSecret:
+  # caSecretKey: ca.crt
```
holgerkoser, line 72:
```suggestion
  # ca of the OpenID provider
  # ca: |
  #   -----BEGIN CERTIFICATE-----
  #   Li4u
  #   -----END CERTIFICATE-----
  # reference to the secretKey of the OpenID provider
  # caSecretKeyRef:
  #   name: oidc-ca-secret-name
  #   key: ca.crt
```

charts/gardener-dashboard/values.yaml
```diff
@@ -41,6 +41,8 @@ ingress:
     nginx.ingress.kubernetes.io/use-port-in-redirects: "true"
     kubernetes.io/ingress.class: nginx
 tls:
+  enabled: true
+  # secretName: gardener-dashboard-tls
```
holgerkoser, line 45:
```suggestion
```

I think it is better to keep the value `tlsSecretName` on top level and do not introduce the enabled flag as part of the tls section. The tls section should only contain information for the gardener-dashboard tls-secret and not for the  ingress.

```
ingress:
  ...
tlsSecretName: gardener-dashboard-tls
tls:
  ...
```

charts/gardener-dashboard/templates/ingress.yaml
```diff
@@ -17,9 +17,9 @@ metadata:
 {{ toYaml . | indent 4 }}
   {{- end }}
 spec:
-{{- if .Values.tls }}
+{{- if .Values.tls.enabled }}
```
holgerkoser, line 20:
```suggestion
{{- if or (eq .Values.tls true) (eq .Values.tlsSecretName true) }}
```

charts/gardener-dashboard/templates/ingress.yaml
```diff
@@ -17,9 +17,9 @@ metadata:
...
   tls:
-  - secretName: {{ .Values.tlsSecretName | default "gardener-dashboard-tls" }}
+  - secretName: {{ .Values.tls.secretName | default "gardener-dashboard-tls" }}
```
holgerkoser, line 22:
```suggestion
  - secretName: {{ .Values.tlsSecretName | default "gardener-dashboard-tls" }}
```

charts/gardener-dashboard/templates/secret-tls.yaml
```diff
@@ -2,11 +2,12 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-{{- if .Values.tls }}
+{{- if .Values.tls.enabled }}
+{{- if and .Values.tls.crt .Values.tls.key -}}
```
holgerkoser, line 6:
```suggestion
{{- if .Values.tls }}
```

charts/gardener-dashboard/templates/secret-tls.yaml
```diff
@@ -2,11 +2,12 @@
...
 apiVersion: v1
 kind: Secret
 metadata:
-  name: {{ .Values.tlsSecretName | default "gardener-dashboard-tls" }}
+  name: {{ .Values.tls.secretName | default "gardener-dashboard-tls" }}
```
holgerkoser, line 10:
```suggestion
  name: {{ .Values.tlsSecretName | default "gardener-dashboard-tls" }}
```

charts/gardener-dashboard/templates/secret-tls.yaml
```diff
@@ -18,6 +19,7 @@ data:
   tls.crt: {{ required ".Values.tls.crt is required" (b64enc .Values.tls.crt) }}
   tls.key: {{ required ".Values.tls.key is required" (b64enc .Values.tls.key) }}
 {{- end }}
+{{- end }}
```
holgerkoser, line 22:
```suggestion
```

holgerkoser:
1. I would create inside the oidc config a property named `caSecretKeyRef` to refer to a key of a secret. 
2. As I have written in the comments the tls section should only have the information for the tls secret and not for the ingress. We should keep the `tlsSecretName` on top level and if this property is defined we create the tls section inside the ingress. The property is than used for the tlsSecret and the ingress.

holgerkoser:
The requirements of this PR will be covered by #1001.

PR #992, opened 2021-04-13
Replace moment-timezone with dayjs

### LINKED ISSUE #956: Migrate from moment.js to dayjs
Replace [moment](https://momentjs.com/) and [moment-timezone](https://momentjs.com/timezone/)with [day.js](https://day.js.org/). 

**Why is this needed**:
Optimize frontend bundle size.

| Library        | Size        |
| ------------- |:-------------:| 
| [moment@2.29.1](https://bundlephobia.com/result?p=moment@2.29.1)      | 71.2 kB |
| [moment-timezone@0.5.33](https://bundlephobia.com/result?p=moment-timezone@0.5.33)      | 100.9 kB |
| [dayjs@1.10.4](https://bundlephobia.com/result?p=dayjs@1.10.4)      | 2.8 kB |

holgerkoser:
This PR replaces the [`moment`](https://momentjs.com/) and [`moment-timezone`](https://momentjs.com/timezone/) with [`dayjs`](https://day.js.org/). The reason are described in the following articles:
* [You don't (may not) need Moment.js](https://github.com/you-dont-need/You-Dont-Need-Momentjs)
* [Google’s Lighthouse is now recommending JavaScript library alternates](https://searchengineland.com/googles-lighthouse-is-now-recommending-javascript-library-alternates-344940)

day.js with plugins

Fixes #956

**Release note**:

```bugfix user

```


frontend/src/components/ShootHibernation/HibernationScheduleEvent.vue
```diff
@@ -77,16 +77,17 @@ SPDX-License-Identifier: Apache-2.0
...
 import split from 'lodash/split'
 import get from 'lodash/get'
 import map from 'lodash/map'
 import find from 'lodash/find'
 import isEqual from 'lodash/isEqual'
 import sortBy from 'lodash/sortBy'
+import { required, requiredIf } from 'vuelidate/lib/validators'
+
+import { getValidationErrors } from '@/utils'
+import moment from '@/utils/moment'
```
petersutter, line 90: When editing an existing hibernation schedule, it always shows the current time when the dialog was opened instead of the configured `wake up` and `hibernate at` values 
Screenshot 2021-04-13 at 23 07 06

petersutter: and another bug on the shoot details page. The creation date is showing the value `lll`. Only the tooltip is correct.

Screenshot 2021-04-13 at 23 14 16

frontend/src/store/index.js
```diff
@@ -55,8 +55,8 @@ import fromPairs from 'lodash/fromPairs'
 import isEqual from 'lodash/isEqual'
 import assign from 'lodash/assign'
 import forOwn from 'lodash/forOwn'
-import moment from 'moment-timezone'
 
+import moment from '@/utils/moment'
```
petersutter, line 59: is the comment in this file still valid or can it be removed?

https://github.com/gardener/dashboard/blob/8f997118d9b5608d6ea9af71b091e31d251c90d5/frontend/src/store/index.js#L86-L88

petersutter (changes requested):
I found two bugs that sneaked in with this PR, please see my comments

frontend/src/store/index.js
holgerkoser: I think we should remove this method completely and use `dayjs.tz.guess() || 'UTC'` directly. We do we check against the list of possible timezones. If it return something it should be valid location. Or do I miss something?
https://github.com/iamkun/dayjs/blob/dev/src/plugin/timezone/index.js#L143-L145

frontend/src/components/ShootHibernation/HibernationScheduleEvent.vue
holgerkoser: Added `localizedFormat` plugin to support `lll` format.

holgerkoser: dayjs does not change the date object if you call `hour` or `minute` but returns the new modified object instead. We have to chain all setters because of this.

```js
// old code
const momentObj = moment().utc()
momentObj.hour(13)
momentObj.minute(27)
console.log(momentObj.format())
// new code
const dayjsObj = moment().utc()
  .hour(13)
  .minute(27)
console.log(dayjsObj.format())
```

frontend/src/store/index.js
holgerkoser: I have move everything into the `@/utils/moment.js` file.

frontend/src/utils/moment.js
```diff
@@ -0,0 +1,622 @@
...
+
+dayjs.extend(relativeTime)
+dayjs.extend(duration)
+dayjs.extend(customParseFormat)
+dayjs.extend(utc)
+dayjs.extend(timezone)
+dayjs.extend(timezoneNames)
+
+function timezoneNames (option, dayjsClass, dayjsFactory) {
+  const tzNames = [
```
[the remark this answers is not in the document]
holgerkoser: This is list what `moment.tz.names()` returns. Dayjs does not have this method yet because the Intl API does not support this yet (see https://github.com/tc39/proposal-intl-enumeration). I think there is no other option than to bundle list list in our code. There is no reason to update this list very often.

grolu (approved):
/lgtm

petersutter (approved):
/lgtm

PR #997, opened 2021-04-16
Show copy button only if mouse over parent element

### LINKED ISSUE #996: Show copy button on mouse-over
The copy button should only be visible if the mouse hovers over the respective text or cell

**Why is this needed**:
To have a cleaner UI. Especially with https://github.com/gardener/dashboard/pull/986 on the cluster list the UI does not look clean with all the non-aligned copy buttons

grolu:
This PR adds the posibility to auto hide copy buttons.
This is a first attempt, maybe someone has a better solution?
Note: With the current implementation, `auto-hide` prop must not be dynamic. This could be changed easily but I don't think there is a use case for it right now.

Fixes #996 

**Release note**:

```other user
Cluster List: Show copy button on hover only
```

### DIFF
--- a/frontend/src/components/CopyBtn.vue
+++ b/frontend/src/components/CopyBtn.vue
@@ -127,8 +127,8 @@ export default {
   mounted () {
     this.enableCopy()
     if (this.autoHide) {
-      this.$parent.$el.addEventListener('mouseover', e => { this.hidden = true })
-      this.$parent.$el.addEventListener('mouseout', e => { this.hidden = false })
+      this.$parent.$el.addEventListener('mouseover', () => this.hidden = true)
+      this.$parent.$el.addEventListener('mouseout', () => this.hidden = false)
     }
   }
 }
--- a/frontend/src/components/CopyBtn.vue
+++ b/frontend/src/components/CopyBtn.vue
@@ -58,8 +58,7 @@ export default {
       snackbar: false,
       clipboard: undefined,
       copySucceeded: false,
-      timeoutId: undefined,
-      hidden: false
+      timeoutId: undefined
     }
   },
   computed: {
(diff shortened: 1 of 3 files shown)


frontend/src/components/CopyBtn.vue
```diff
@@ -122,6 +126,16 @@ export default {
   },
   mounted () {
     this.enableCopy()
+    if (this.autoHide) {
+      this.$parent.$el.addEventListener('mouseover', e => { this.hidden = true })
+      this.$parent.$el.addEventListener('mouseout', e => { this.hidden = false })
```
petersutter, line 131:
```suggestion
      this.$parent.$el.addEventListener('mouseover', () => this.hidden = true)
      this.$parent.$el.addEventListener('mouseout', () => this.hidden = false)
```
(not tested)

grolu: Good catch, leftover from debugging :)

frontend/src/components/CopyBtn.vue
```diff
@@ -51,14 +51,18 @@ export default {
     userFeedback: {
       type: Boolean,
       default: true
+    },
+    autoHide: {
+      type: Boolean
```
holgerkoser, line 56:
```suggestion
      type: Boolean,
      default: false
```

holgerkoser: Just to make it clear

holgerkoser:
Wouldn't it be better to implement this feature explicitly by creating a new wrapper component with a default slot and the copy button. The copy button has a v-show and the DOM events are handled in this new wrapper component?

petersutter (approved):
/lgtm

holgerkoser (approved):
/lgtm

PR #1001, opened 2021-04-19
Helm chart enhancements

### LINKED ISSUE #985: VPA configuration for the Helm charts
The Helm charts lack a `VerticalPodAutoscaler` configuration. We should add it to allow the components to be properly scaled automatically by VPA.

**Why is this needed**:
Less manual ops, improved resource utilization.

### LINKED ISSUE #990: Using TLS in ingress needs certificates to be specified
**What happened**: With b9f6822b28ae16d5784494a58063698a4bf19d90, when one wants the Ingress to have a `spec.tls` section, `tls` needs to be specified in the values. This disables the possibility to have a `spec.tls` section in the ingress without specifying certificates, which is needed when using e.g. cert-manager.

**What you expected to happen**: Having a `spec.tls` section without explicitly specifying certificates should be possible

**How to reproduce it (as minimally and precisely as possible)**: Use any version newer than or equal to b9f6822b28ae16d5784494a58063698a4bf19d90 and not specify `tls`.

**Proposal**:

I propose to move the ingress TLS configuration to the one that is generated by `helm create`. It looks like this:

``` 
spec:
  {{- if .Values.ingress.tls }}
  tls:
    {{- range .Values.ingress.tls }}
    - hosts:
        {{- range .hosts }}
        - {{ . | quote }}
        {{- end }}
      secretName: {{ .secretName }}
    {{- end }}
  {{- end }}
```

I would then add a `.crt` and `.key` field to the `.hosts` range so that certificates can be explicitly specfied.

With this, we could then specify:

```
ingress:
  annotations:
    cert-manager.io/cluster-issuer: letsencrypt-production
  tls:
    - secretName: dashboard-secret
      crt: "optional cert" # if set, requires 'key' to be specified too and will create secret with name 'secretName'
      key: "optional key" # if set, requires 'crt' to be specified too and will create secret with name 'secretName'
      hosts:
        - dashboard.example.com
```

I’d like some feedback for that - if it is deemed a good idea, I’ll submit a PR.

holgerkoser:
The following improvements in the `gardener-dashboard` helm charts are part of this PR:

#####  All non-sensible defaults in `values.yaml` have been removed
- removed default value for `.Values.apiServerUrl` (value is required)
- removed default values for `.Values.tls.crt` and `.Values.tls.key` 
- removed default values for `.Values.hosts`
- removed default value for `.Values.oidc.issuerUrl`
- removed default and fallback value for `.Values.oidc.redirectUri`

##### More flexible and consistent ingress template configuration
The following ingress related configuration values have been moved from top level into the ingress section:
- moved value `.Values.tls` to `.Values.ingress.tls` and removed default values for `key` and `crt`
- moved value `.Values.hosts` to `.Values.ingress.hosts` and removed default values
- moved value `.Values.tlsSecretName` to `.Values.ingress.tls.secretName`

###### Examples:
1. Example configuration with tls enabled and cert and key specified: 
```yaml
ingress:
  annotations:
    nginx.ingress.kubernetes.io/ssl-redirect: "true"
    nginx.ingress.kubernetes.io/use-port-in-redirects: "true"
    kubernetes.io/ingress.class: nginx
  hosts:
  - dashboard.example.org
  tls:
    secretName: gardener-dashboard-tls
    crt: |
    -----BEGIN CERTIFICATE-----
    ...
    -----END CERTIFICATE-----
    key: |
    -----BEGIN RSA PRIVATE KEY-----
    ...
    -----END RSA PRIVATE KEY-----
```

2. Example configuration with tls enabled and cert-manager:
```yaml
ingress:
  annotations:
    ...
    cert.gardener.cloud/purpose: managed
  hosts:
  - dashboard.example.org
  tls:
    secretName: gardener-dashboard-tls
```

3. Example configuration with tls disabled:
```yaml
ingress:
  annotations:
    kubernetes.io/ingress.class: nginx
  hosts:
  - dashboard.example.org
  tls: ~
```

###### Examples:
1. Example configuration with vpa enabled:
```yaml
vpa:
  updateMode: Auto
  controlledValues: RequestsOnly
  minAllowedCpu: 30m
  minAllowedMemory: 80Mi
  maxAllowedCpu: 250m
  maxAllowedMemory: 250Mi
```

###### Examples:
1. Example oidc configuration oidc with ca secret reference:
```yaml
oidc:
  issuerUrl: https://identity.example.org
  clientId: dashboard
  clientSecret: my-super-secure-secret
  redirectUri: https://dashboard.example.org/auth/callback
  caSecretKeyRef:
    name: oidc-ca-secret-name
    key: ca.crt
```

##### Miscellaneous
- Adds unit tests for all helm templates of the `gardener-dashboard` chart.
- Adds support for the vscode [vetur](https://marketplace.visualstudio.com/items?itemName=octref.vetur) extension.

Fixes #985, Fixes #990, Fixes #989 

**Release note**:

```breaking operator
Please note the following breaking changes in the `values.yaml` file of the `gardener-dashboard` helm chart:
- The configuration properties `tlsSecretName`, `tls` and `hosts` in the `values.yaml` have been moved to `ingress.tls.secretName`, `ingress.tls` and `ingress.hosts`.
- The configuration property `apiServerUrl` is now a required property. The dummy default value has been removed.
- The configuration property `oidc.redirectUri` is now a required property. The fallback value based on the first `hosts` entry has been removed.
```

```feature operator
`gardener-dashboard` helm chart:
- Vertical Pod Autoscaler can be enabled via the configuration property `vpa`  in the `values.yaml` file of the `gardener-dashboard` helm chart.
- OpenID Provider certificate authority can be passed via secret reference `oidc.caSecretRef` in the `values.yaml` file of the `gardener-dashboard` helm chart.
```


charts/gardener-dashboard/templates/vpa.yaml
```diff
@@ -0,0 +1,24 @@
...
+  targetRef:
+    apiVersion: {{ include "deploymentversion" . }}
+    kind: Deployment
+    name: gardener-dashboard
+  updatePolicy:
+    updateMode: Auto
+  resourcePolicy:
+    containerPolicies:
+      - containerName: gardener-dashboard
+        minAllowed:
```
petersutter, line 21: don't we want to set sane `maxAllowed` values?

charts/gardener-dashboard/values.yaml
```diff
@@ -32,38 +36,51 @@ resources:
...
-  key: |
-    -----BEGIN RSA PRIVATE KEY-----
-    Li4u
-    -----END RSA PRIVATE KEY-----
+  # # configuration of hosts used for rules and tls
+  # hosts:
+  # - dashboard.ingress.example.org
+  # - dashboard.example.org
+  tls:
+    secretName: gardener-dashboard-tls
```
petersutter, line 50: shouldn't this also be commented out by default?

charts/gardener-dashboard/templates/vpa.yaml
holgerkoser: The Gardener also only specifies `minAllowed` values. Why do you think we should specify `maxAllowed` values?

charts/gardener-dashboard/values.yaml
holgerkoser: This is the default name for the secret. If I remove this default value the `.Values.ingress.tls` property will be falsy. It would need to look like this:
```
  tls: {}
    # secretName: gardener-dashboard-tls
```
I think it would be anyway better to specify all default values in `value.yaml`. What do you think.

charts/gardener-dashboard/templates/vpa.yaml
holgerkoser: added maxAllowed

charts/gardener-dashboard/values.yaml
holgerkoser: It is required now. If not changed the default value from `values.yaml` will be used.

petersutter:
cc @Diaphteiros

charts/vpa.yaml
```diff
@@ -0,0 +1,12 @@
...
+metadata:
+  name: gardener-dashboard-vpa
+  namespace: garden
+spec:
+  targetRef:
+    apiVersion: "apps/v1"
+    kind:       Deployment
+    name:       gardener-dashboard
+  updatePolicy:
+    updateMode: "Off"
```
petersutter, line 12:
```suggestion
    updateMode: "Off"

```
add newline at the end?

charts/gardener-dashboard/values.yaml
```diff
@@ -12,58 +12,81 @@ image:
...
+#   minAllowedCpu: 50m
+#   minAllowedMemory: 100Mi
+#   maxAllowedCpu: 500m
+#   maxAllowedMemory: 500Mi
+
 logLevel: debug
-apiServerUrl: https://api.example.org
-# the certificateAuthorityData of the kube apiserver
+# # the external url of the apiServer
+# apiServerUrl: https://api.example.org
```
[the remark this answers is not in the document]
holgerkoser: ✔️

petersutter (approved):
/lgtm

petersutter:
@holgerkoser I have updated the release notes, PTAL

PR #1002, opened 2021-04-20
Use the AWS IAM policy from g/provider-aws

ialidzhikov:
This is basically the diff between the policy in g/dashboard and g/provider-aws (ref https://github.com/gardener/gardener-extension-provider-aws/blob/v1.22.2/docs/usage-as-end-user.md#permissions).
I think we can remove these permission as they are route53 specific. In general the dns provider can be of any type and these permissions should be rather documented under g/external-dns-management.

**Release note**:

```bugfix user
NONE
```

### DIFF
--- a/frontend/src/components/dialogs/SecretDialogAws.vue
+++ b/frontend/src/components/dialogs/SecretDialogAws.vue
@@ -155,17 +155,6 @@ export default {
             ],
             Effect: 'Allow',
             Resource: '*'
-          },
-          {
-            Effect: 'Allow',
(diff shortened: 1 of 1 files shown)


petersutter (approved):
/lgtm

grolu (approved):
/lgtm
Regarding required permissions for DNS management, we need to add them in the manage secrets dns secrets help. I'm already working on it.

holgerkoser (approved):
/lgtm

PR #1009, opened 2021-05-04
Do not set .spec.provider.controlPlaneConfig.zone for alicloud

ialidzhikov:
/kind bug

Currently creating a new alicloud Shoot from the dashboard, the dashboard wrongly sets the `.spec.provider.controlPlaneConfig.zone` for alicloud Shoots. The corresponding zone field was removed ~1y ago with https://github.com/gardener/gardener-extension-provider-alicloud/pull/64.
Setting unknown fields causes problem with the strict decoding that provider extensions now enforce.

**Release note**:

```bugfix user
An issue causing the dashboard to set the `.spec.provider.controlPlaneConfig.zone` field for newly created alicloud Shoots is now resolved. The corresponding ControlPlaneConfig field is invalid as it was removed with https://github.com/gardener/gardener-extension-provider-alicloud/pull/64.
```

### DIFF
--- a/frontend/src/utils/createShoot.js
+++ b/frontend/src/utils/createShoot.js
@@ -299,7 +299,6 @@ export function getControlPlaneZone (workers, infrastructureKind, oldControlPlan
   switch (infrastructureKind) {
     case 'gcp':
     case 'openstack':
-    case 'alicloud':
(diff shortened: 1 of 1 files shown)


grolu (approved):
/lgtm

petersutter (approved):
/lgtm

PR #1012, opened 2021-05-06
Do not auto fix frontend lint errors in docker build

holgerkoser:
The Vue CLI plugin `@vue/cli-plugin-eslint` fixes lint errors per default. This means that currently the build step is successful if we have lint errors in the frontend. This kind of errors should lead to build errors in case of e.g. external PRs. With this PR the script `yarn lint` does not auto fix lint error in the frontend. This is also the default in all other workspaces.

Fixes #

**Release note**:

```bugfix user

```


petersutter (approved):
/lgtm, thanks

grolu (approved):
/lgtm good catch!

PR #1021, opened 2021-05-19
Improved search capabilities

### LINKED ISSUE #960: Support negative search terms in the search field of the "all projects" cluster view
You can specify search terms for the "all projects" cluster view.
It would be very helpful to have a way to specify negative filter terms or to invert the filter (similar to `grep -v`)

**Why is this needed**:
If you have lots a failing shoots in a single project (> 20), it is hard no to loose the overview of all the other projects.
The only work around currently is to order the list by project and then use paging. This is hard to use, as the list may be updated frequently and the shoots move between the pages.

### LINKED ISSUE #994: Add support for filtering per infrastructure provider

While viewing a project, allow filtering of clusters based on the infrastructure provider.

**Why is this needed**:

In large projects with more than a few hundred clusters spread among all infrastructure providers it is hard to monitor clusters of only a particular infra. In this scenario simply sorting isn't very much helpful either. Hence it would be nice if I could apply one additional filter to reduce the results. This can be generalized as a feature for other columns with finite values e.g. `purpose` too.

### LINKED ISSUE #1004: Improving the search / filtering capability of the dashboard
The current filtering capabilities in our tables is quite limited. While it may be sufficient for member or secrets lists, the limitations become obvious when trying to filter the shoot list. This is especially painful when operators filter / search on the *all projects* page.

**Why is this needed**:
The demand is clearly there, we already have some related issues. Let's collect issues and ideas here.

https://github.com/gardener/dashboard/issues/683
https://github.com/gardener/dashboard/issues/994
https://github.com/gardener/dashboard/issues/960
https://github.com/gardener/dashboard/issues/1031

Another issue reported by some users is responsiveness when searching in very large projects / all projects. We need to make the search perform the request asynchronously to detach it from the UI loop.

Additional requirement: Use filters also in project scope, not only on all projects page, e.g. filter for clusters with issues for one project only.

grolu:
Added extended search capabilities

Adds more search capabilities (see below). This should solve several raised requirements (see linked issues).
Moreover, this PR introduces a search delay to prevent multiple searches to block the UI while typing.

Screenshot 2021-06-25 at 15 55 48

Fixes #1004 Fixes #994 Fixes #960

**Release note**:

```feature user
Added extended search capabilities to cluster search:
- Search params are now ANDed, allowing one to refine the search
- Use quotes for exact words or phrases
- Use minus sign to exclude words that you don't want
```


holgerkoser:
I would propose to go with the gmail like advanced search query syntax and use something like [search-query-parser](https://www.npmjs.com/package/search-query-parser).
```js
const searchQueryParser = require("search-query-parser")

function parseSearchQuery(query) {
  const options = {
    tokenize: true,
    alwaysArray: true,
    offsets: false,
    keywords: ["name", "infra", "seed", "from", "purpose", "version", "ticket"],
    ranges: ["date"],
  }
  return searchQueryParser.parse(query, options)
}
const query =
  'from:foo@sap.com,bar@sap.com name:delete purpose:eval date:20200503-20210501 "free text" foo -hugo';

console.log(parseSearchQuery(query))
/*
{
  text: [ 'free text', 'foo' ],
  from: [ 'foo@sap.com', 'bar@sap.com' ],
  name: [ 'delete' ],
  purpose: [ 'eval' ],
  date: { from: '20200503', to: '20210501' },
  exclude: { text: 'hugo' }
}
*/
```

frontend/src/store/modules/shoots/getters.js
```diff
@@ -150,38 +151,51 @@ export default {
     return state.initialNewShootResource
   },
   searchItems (state, getters, rootState, rootGetters) {
+    const searchPattern = /(-?"[^"]+")*([\S]+)*/g
+    const termPattern = /^(-)?("?)([^"]+)("?)$/g
```
petersutter, line 155: add tests for the regex so that we can be sure in the future that it still works as expected in case we need to adapt it

petersutter (approved):
/lgtm

holgerkoser (approved):
\lgtm

PR #1043, opened 2021-06-24
Hetzner support

poelzi:

This PR adds support for the hetzner provider extension

[hcloud provider extension](https://github.com/23technologies/gardener-extension-provider-hcloud)
[hcloud machine controller](https://github.com/23technologies/machine-controller-manager-provider-hcloud)

**Release note**:

```feature user
* Support for the hetzner cloud extension (hcloud)
```


gardener-robot-ci-1:

frontend/src/assets/hcloud.svg
```diff
@@ -0,0 +1,81 @@
...
+  -webkit-filter: invert(100%) hue-rotate(180deg) !important;
+  filter: invert(100%) hue-rotate(180deg) !important;
+}
+[style*="background:url"] *,
+[style*="background-image:url"] *,
+[style*="background: url"] *,
+[style*="background-image: url"] *,
+input,
+[background] *,
+img[src^="https://s0.wp.com/latex.php"],
```
petersutter, line 32: for what is this external page used? Can this be removed? Could you also double check the other css rules etc if they are really required?

frontend/src/views/Secrets.vue
```diff
@@ -313,6 +317,13 @@ export default {
               value: secretData.accessKeyID
             }
           ]
+        case 'hcloud':
+          return [
+            {
+              label: 'HCloud Token',
```
holgerkoser, line 323:
```suggestion
              label: 'Hetzner Cloud Token',
```

frontend/src/components/dialogs/SecretDialogHCloud.vue
```diff
@@ -0,0 +1,135 @@
...
+        @blur="$v.hcloudToken.$touch()"
+        ></v-text-field>
+      </div>
+    </template>
+
+    <template v-slot:help-slot>
+      <div>
+        <p>
+          Before you can provision and access a Kubernetes cluster on Hetzner Cloud, you need to add a HCloud Token.
+          The Gardener needs these credentials to provision and operate HCloud infrastructure for your Kubernetes cluster.
```
holgerkoser, line 36:
```suggestion
          Before you can provision and access a Kubernetes cluster on Hetzner Cloud, you need to add a Hetzner Cloud Token.
          The Gardener needs these credentials to provision and operate Hetzner Cloud infrastructure for your Kubernetes cluster.
```

holgerkoser: The Hetzner Cloud documentation always uses the term `Hetzner Cloud`. Only the CLI is called `hcloud`. I think we should also use the full name  `Hetzner Cloud` in the text.

frontend/src/assets/hcloud.svg
holgerkoser: Let's use the svg from [here](view-source:https://www.hetzner.com/assets/Uploads/icon-circle-cloud.svg)

frontend/src/components/dialogs/SecretDialogHCloud.vue
```diff
@@ -0,0 +1,135 @@
...
+
+<template>
+  <secret-dialog
+    :value=value
+    :data="secretData"
+    :data-valid="valid"
+    :secret="secret"
+    cloud-provider-kind="hcloud"
+    create-title="Add new HCloud Secret"
+    replace-title="Replace HCloud Secret"
```
petersutter, line 15: this should probably be adapted as well

grolu:
Hi @poelzi,
thank you for your contribution. Could you please adapt the header in your files to our standard header

```
<!--
SPDX-FileCopyrightText: 2021 SAP SE or an SAP affiliate company and Gardener contributors

SPDX-License-Identifier: Apache-2.0
 -->
 ```
 
 By merging your request we take over ownership and responsibility for your code, therefore the headers have to be adapted.
 
 Thank you

File: frontend/src/components/ShootWorkers/ManageWorkers.vue [hunk 5]
```diff
@@ -120,6 +120,9 @@ export default {
       return findFreeNetworks(this.currentZonesNetworkConfiguration, this.existingWorkerCIDR, this.cloudProviderKind, this.allZones.length)
     },
     availableZones () {
+      if (!this.zonedCluster) {
```
grolu, line 123: One last question before we merge your PR: Why did you add this code here? While I don't think that it will cause any problem with other infrastructures, I'm curious why you added it in the first place. I briefly tested your implementation and for me it worked without errors even without this code (at least not on the create cluster screen). Did you experience an error when changing an existing worker or did you just add this to be on the safe side?

poelzi: Because the hetzner provider implemenation was first without zones and the frontend assumes zones so i fixed some issues in this regard. We later switched to zones and the fixes became obsolete. I think they are still useful so I let them in this PR.

grolu: Just for clarification: When did the issue occur? After shoot creation on the configure workers popup? Or somewhere else?

poelzi: When creating a now shoot, it complained about missing zones even the api returned no zones. I added the zonedCluster to return false in hcloud case and with this fix the problem was gone.

grolu (approved):
/lgtm

petersutter (approved):
/lgtm

PR #1051, opened 2021-06-25
Fix privileged terminal container bug

### LINKED ISSUE #1050: Ops-toolbelt container in terminal pod is not started with privileged security context when specified
**What happened**:
I used the terminal functionality to get a console connection to a cluster's node and I enabled the `privileged` option from the `Terminal Configuration` UI. However, after inspecting the terminal pod I noticed that the ops pod container is not started with a privileged security context.
```
containers:
  - env:
    - name: KUBECONFIG
      value: /mnt/.kube/config
    - name: KUBERNETES_SERVICE_HOST
      value: api.example.com
    image: eu.gcr.io/sap-se-gcr-k8s-public/eu_gcr_io/gardener-project/gardener/ops-toolbelt:0.11.0-mod1
    imagePullPolicy: IfNotPresent
    name: terminal
    resources: {}
    securityContext:
      privileged: false
...
```

I expected that the ops-pod container would be started with a privileged security context and to be able to execute commands like listing iptable rules.

Open a terminal, chose cluster as target and from terminal options chose Cluster Terminal Runtime and enable `Privileged`.
Inspect the created pod with kubectl.

**Environment**:
Dashboard version 1.50.2

petersutter:
This PR fixes an issue where the field `spec.host.pod.container.privileged` was not set on the `Terminal` resource and hence the `terminal-controller-manager` did not create the terminal container with 

```yaml
securityContext:
  privileged: true
```

Fixes #1050 

**Release note**:

```bugfix user
Fixed an issue where the terminal container was not created with `privileged` set to `true` of the `container`s `securityContext` when enabling the `Privileged` flag on the terminal settings UI
```

### DIFF
--- a/backend/lib/services/terminals/index.js
+++ b/backend/lib/services/terminals/index.js
@@ -375,8 +375,8 @@ function getContainerConfigFromBody ({ container }) {
 }
 
 function getConfigFromBody (body) {
-  const config = _.pick(body, ['node', 'privileged', 'hostPID', 'hostNetwork', 'container'])
-  config.container = _.pick(config.container, ['image', 'command', 'args'])
+  const config = _.pick(body, ['node', 'hostPID', 'hostNetwork', 'container'])
+  config.container = _.pick(config.container, ['image', 'command', 'args', 'privileged'])
   return config
 }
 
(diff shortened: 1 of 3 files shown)


petersutter:
cc @plkokanov, thanks for reporting the issue

holgerkoser (approved):
\lgtm

grolu (approved):
/lgtm

PR #1052, opened 2021-06-28
Remove circular dependencies

### LINKED ISSUE #718: Circular dependency between Store and Emitter
**What happened**:
We have a circular dependency between Store and Emitter in the frontend.

**Environment**:

holgerkoser:
This PR removes the circular dependency between `store` and `emitter` via a Vuex plugin and between `store` and `utils`. I addition it ensures that the store is only used correctly via the component API and not via direct import. I also added the [circular-dependency-plugin](https://github.com/aackerman/circular-dependency-plugin) to the webpack config.

Fixes #718

**Release note**:

```bugfix user

```


frontend/src/store/index.js
```diff
@@ -1010,6 +1020,198 @@ const getters = {
   },
   onlyShootsWithIssues (state, getters) {
     return getters['shoots/onlyShootsWithIssues']
+  },
+  projectNameByNamespace (state, getters) {
+    return ({ namespace } = {}) => {
+      const project = find(getters.projectList, ['metadata.namespace', namespace])
+      return get(project, 'metadata.name') || replace(namespace, /^garden-/, '')
```
petersutter, line 1027: Do we still need this fallback?

holgerkoser: I don't know. But I don't wanted to start changing any code in this PR.

petersutter: ok

grolu (approved):
/lgtm

PR #1068, opened 2021-07-22
Direct update of store for dns configuration

holgerkoser:
This PR removes event from dns configuration components and uses the store to sync state between different components. 

Fixes #

**Release note**:

```bugfix user

```

### DIFF
--- a/frontend/src/store/index.js
+++ b/frontend/src/store/index.js
@@ -61,6 +61,7 @@ import assign from 'lodash/assign'
 import forOwn from 'lodash/forOwn'
 import replace from 'lodash/replace'
 import sample from 'lodash/sample'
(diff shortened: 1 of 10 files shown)


frontend/src/store/index.js
```diff
@@ -74,15 +75,16 @@ import draggable from './modules/draggable'
 import members from './modules/members'
 import cloudProviderSecrets from './modules/cloudProviderSecrets'
 import tickets from './modules/tickets'
-import componentStates from './modules/componentStates'
+import shootSpec from './modules/shootSpec'
```
holgerkoser, line 78: stagingShootSpec

grolu (approved):
/lgtm

PR #1072, opened 2021-08-03
Do not modify theme objects returned from store

### LINKED ISSUE #1035: Do not mutate vuex store state outside mutation handlers
**What happened**:
Do not mutate the custom theme configuration.
https://github.com/gardener/dashboard/blob/master/frontend/src/store/index.js#L1311-L1315

**Environment**:

grolu:

Fixes #1035

**Release note**:

```bugfix user

```

### DIFF
--- a/frontend/src/store/index.js
+++ b/frontend/src/store/index.js
@@ -1546,21 +1546,18 @@ const actions = {
 
     const themes = get(Vue, 'vuetify.framework.theme.themes')
     if (themes) {
-      const applyCustomThemeConfiguration = (name) => {
-        const customTheme = get(state, ['cfg', 'themes', name])
-        if (customTheme) {
-          forOwn(customTheme, (value, key) => {
-            const color = get(colors, value)
-            if (color) {
-              themes[name][key] = color
-            } else if (isHtmlColorCode(value)) {
-              themes[name][key] = value
-            }
-          })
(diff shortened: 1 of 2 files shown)


frontend/src/store/index.js
```diff
@@ -1546,7 +1546,7 @@ const actions = {
     const themes = get(Vue, 'vuetify.framework.theme.themes')
     if (themes) {
       const applyCustomThemeConfiguration = (name) => {
-        const customTheme = get(state, ['cfg', 'themes', name])
+        const customTheme = cloneDeep(get(state, ['cfg', 'themes', name]))
```
holgerkoser, line 1549: I think it would be better to directly update the vuetify theme without cloning and changing the configuration.
```js
  for (const name of ['light', 'dark']) {
    const theme = get(Vue.vuetify.framework.theme.themes, name)
    const themeConfiguration = get(state.cfg.themes, name)
    forOwn(themeConfiguration, (value, key) => {
      const color = get(colors, value)
      if (color) {
        theme[key] = color
      } else if (isHtmlColorCode(value)) {
        theme[key] = value
      }
    })
  }
```

grolu: ok

petersutter (approved):
/lgtm

PR #1073, opened 2021-08-06
Fix controllerregistrations permission issue

grolu:

Fixes #

**Release note**:

```bugfix user

```

### DIFF
--- a/backend/lib/services/controllerregistrations.js
+++ b/backend/lib/services/controllerregistrations.js
@@ -6,23 +6,36 @@
 
 'use strict'
 
-const { Forbidden } = require('http-errors')
 const { getControllerRegistrations } = require('../cache')
 const authorization = require('./authorization')
 const _ = require('lodash')
 
 exports.listExtensions = async function ({ user }) {
-  const allowed = await authorization.canListControllerRegistrations(user)
-  if (!allowed) {
-    throw new Forbidden('You are not allowed to list controllerregistrations')
-  }
-
   const controllerregistrations = getControllerRegistrations()
-  return _.map(controllerregistrations, ({ metadata, spec }) => {
+  // required resoure kinds are essential for the frontend and need to be returned even if the user has not the right to read controllerregistrations
+  const requiredResourceKinds = ['Network', 'DNSProvider']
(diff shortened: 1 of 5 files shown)


backend/lib/services/controllerregistrations.js
```diff
@@ -6,23 +6,36 @@
...
 
 exports.listExtensions = async function ({ user }) {
-  const allowed = await authorization.canListControllerRegistrations(user)
-  if (!allowed) {
-    throw new Forbidden('You are not allowed to list controllerregistrations')
-  }
-
   const controllerregistrations = getControllerRegistrations()
-  return _.map(controllerregistrations, ({ metadata, spec }) => {
+  // required resoure kinds are essential for the frontend and need to be returned even if the user has not the right to read controllerregistrations
```
petersutter, line 15:
```suggestion
  // required resoure kinds are essential for the frontend and need to be returned even if the user has not the permission to read controllerregistrations
```

holgerkoser:
```js

const REQUIRED_RESOURCE_KINDS = ['Network', 'DNSProvider']

exports.listExtensions = async function ({ user }) {
  const controllerregistrations = getControllerRegistrations()
  const allowed = await authorization.canListControllerRegistrations(user)

  const extensions = []
  for (const { metadata, spec = {} } of controllerregistrations) {
    const name = metadata.name
    if (allowed) {
      const version = _.get(spec, 'deployment.providerConfig.values.image.tag')
      const resources = spec.resources
      extensions.push({ name, version, resources })
    } else {
      // required resoure kinds are essential for the frontend and need to be returned even if the user has not the permission to read controllerregistrations
      const resources = _.filter(spec.resources, ({ kind }) => REQUIRED_RESOURCE_KINDS.includes(kind))
      if (!_.isEmpty(resources)) {
        extensions.push({ name, resources })
      }
    }
  }
  return extensions
}
```

grolu: done

holgerkoser (approved):
/lgtm

petersutter (approved):
/lgtm

PR #1083, opened 2021-08-11
[master] DNS provider w/o secretBinding should be readonly

holgerkoser:
If no secret binding was found for a given secretName and the cluster is not new,
then we assume that the secret exists and was manually created.
The DNS provider row should not be changed in this case.

Fixes #

**Release note**:

```bugfix user
none
```

### DIFF
--- a/frontend/src/components/ShootDns/DnsProviderRow.vue
+++ b/frontend/src/components/ShootDns/DnsProviderRow.vue
@@ -5,13 +5,14 @@ SPDX-License-Identifier: Apache-2.0
 -->
 
 <template>
-  <v-tooltip top :disabled="!secretBindingMissing" open-delay="0">
+  <v-tooltip top :disabled="!readonly" open-delay="0">
     <template v-slot:activator="{ on }">
      <v-row align="center" class="ma-0" v-on="on">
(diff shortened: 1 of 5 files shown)


frontend/src/components/ShootDns/DnsProviderRow.vue
```diff
@@ -261,6 +270,9 @@ export default {
       return getValidationErrors(this, field)
     },
     onDelete () {
+      if (this.readonly) {
+        return
+      }
```
petersutter, line 275: is this check really needed as the button is anyhow disabled
and if it is needed, why not check for `readonly || primaryReadonly`?

holgerkoser: No it is not necessary. Yes it should be `this.readonly || this.primaryReadonly`. I have added it because I added the same to the `setData` method. Because I don't want to update the valid state if it is readonly. But we could move this to `updateValid` and also remove it from `setData`.

grolu (approved):
/lgtm
Rel note muss aber in den hotfix PR

petersutter (approved):
/lgtm

PR #1086, opened 2021-08-12
[hotfix-1.51] DNS provider w/o secretBinding should be readonly

holgerkoser:
If no secret binding was found for a given secretName and the cluster is not new,
then we assume that the secret exists and was manually created.
The DNS provider row should not be changed in this case.

Fixes #

**Release note**:

```bugfix user
Fixed a problem in the DNS provider configuration that caused a newly added DNS provider to always be disabled on an existing cluster.
```


grolu (approved):
/lgtm

PR #1087, opened 2021-08-12
[master] Do not show scrollbars when there is enough space.

### LINKED ISSUE #1085: Enhance Scrollbars in WebUI
Remove scroll bars from info fields, e.g. in cluster overview.

Bildschirmfoto 2021-08-12 um 10 10 45

This also affects the Status popup as well as the readiness popups. Might affect some more.

**Why is this needed**:
If your screen is big enough this should not be viewed.

holgerkoser:
Fixes the problem that scroll bars are displayed in the UI even if there is enough space available

Fixes #1085 

**Release note**:

```bugfix user
Fixes the problem that scroll bars are displayed in the UI even if there is enough space available.
```

### DIFF
--- a/frontend/src/components/GPopper.vue
+++ b/frontend/src/components/GPopper.vue
@@ -138,7 +138,7 @@ export default {
   .inner-card {
     max-width: 1000px;
     max-height: 85vh;
-    overflow-y: scroll;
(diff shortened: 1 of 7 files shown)


grolu (approved):
/lgtm

PR #1091, opened 2021-08-30
[master] Fixed memory leak in cluster list

holgerkoser:
Fixed a memory leak problem caused by registering popovers to the global esc key event. The popover components were registered to the global esc key event on creation but not deregistered on beforeDestroy. Therefore, the popover components could no longer be cleared by the garbage collector.

Fixes #

**Release note**:

```bugfix user
Fixed a memory leak in the cluster list. The memory consumption of the browser increased over time until finally the corresponding process crashed.
```


petersutter (approved):
/lgtm

PR #1095, opened 2021-09-02
[master] Handle responses without body correctly

### LINKED ISSUE #1093: Rotate Service Account Secret shows an error in the frontend
**What happened**:
If you trigger rotation of a service account secret from gardener dashboard the rotation is initiated correctly but when the frontend handles the backend response an "_Unexpected end of JSON input_" error occurs.

image

The  backend response should be handled correctly and no error should be displayed if the operation has been successful.

**Environment**:

holgerkoser:
With this PR http response without body are handled correctly in the frontend. This caused error messages in the dashboard although no error occurred at all.

Fixes #1093

**Release note**:

```bugfix user
Solves a problem when a user has rotated a service account secret. In the frontend an error message was erroneously displayed, which is fixed now.
```


frontend/src/utils/api.js
```diff
@@ -7,6 +7,33 @@
 import get from 'lodash/get'
 
 /* General Purpose */
+async function toPlainResponseObject (response) {
+  const { status, statusText } = response
+  const headers = {}
+  for (const [key, value] of response.headers.entries()) {
+    headers[key] = value
+  }
+  const contentLength = +get(headers, 'content-length', '0')
```
petersutter, line 16:
```suggestion
  const contentLength = get(headers, 'content-length', '0')
```

petersutter: as you just wrote me:
```js
typeof ('200')
"string"

typeof +('200')
"number"
```
-> ignore this suggestion

petersutter (approved):
/lgtm

PR #1102, opened 2021-09-20
[master] Fixed some issues

### LINKED ISSUE #1100: Dashboard is showing 410 Gone for existing clusters
**What happened**:
A user has opened the list of all clusters with issues and navigates into the details of a cluster which is in creation state but still has errors. On the cluster details page to creation process of this cluster is progressing and the cluster becomes ready and healthy. The is removed from the list of clusters with issues and on the cluster details page the user gets an 410 error page  with the message "The cluster you are looking for is no longer available" 

The cluster details page should show the cluster as healthy and ready and not as gone if the creation process is successful.

**Environment**:

holgerkoser:
* Fixed issue #1100
* shootItem mixin requires shootItem prop to be defined
* Update `caniuse-lite` to latest version.

Fixes #1100

**Release note**:

```bugfix user
NONE
```


grolu (approved):
/lgtm

PR #1104, opened 2021-09-21
Use pull policy "IfNotPresent" for dashboard image by default

vpnachev:
Use pull policy "IfNotPresent" for dashboard image by default. 
This way the start-up time for new replicas are much faster when the image is always available on the node. It also saves some network costs by avoiding to pull the image when not needed.

Fixes #

**Release note**:

```bugfix user

```


grolu (approved):
/lgtm

PR #1113, opened 2021-10-12
[hotfix-1.52] Fixed: Invalid (outdated) image breaks configure worker dialog

grolu:

Fixes #

**Release note**:

```bugfix user
Fixed an issue that caused `Configure Workers` dialog to break if a cluster has an outdated worker image configured that is no longer in the list of supported images (removed from cloud profile)
```


holgerkoser (approved):
/lgtm

PR #1117, opened 2021-10-14
[master] Set apiServer of the target cluster in case of garden terminals

petersutter:
The apiServer for the garden target cluster must always be set.

Fixes #

**Release note**:

```bugfix user

```


grolu (approved):
/lgtm

PR #1127, opened 2021-11-05
Clarify behaviour after Secret change in dialogue

BeckerMax:

Repeatedly Shoot owners were surprised by the behaviour that after they rotate a
cloudprovider secret in the garden cluster it is not instantly synced to the Seed cluster.

This commit adds an additional alert dialogue to help the user understand when a Secret
change will become active for their Shoots.

Fixes gardener/gardener#4630

cc @grolu 

**Release note**:

```doc user
Added an additional warning message when changing cloud provider secrets to clarify when the change will become active for Shoot owners.
```

### DIFF
--- a/frontend/src/components/dialogs/SecretDialog.vue
+++ b/frontend/src/components/dialogs/SecretDialog.vue
@@ -62,7 +62,7 @@ SPDX-License-Identifier: Apache-2.0
         This secret is used by {{relatedShootCount}} clusters. The new secret should be part of the same account as the one that gets replaced.
       </v-alert>
        <v-alert :value="!isCreateMode && relatedShootCount > 0" type="warning" tile>
-        Shoots will only start using the new secret after they got reconciled. Therefore, wait until all {{relatedShootCount}} Shoots using the secret are reconciled before you disable the old secret in your infrastructure account. Otherwise the Shoots will no longer function.
+        Clusters will only start using the new secret after they got reconciled. Therefore, wait until all clusters using the secret are reconciled before you disable the old secret in your infrastructure account. Otherwise the clusters will no longer function.
       </v-alert>
       <v-divider></v-divider>
       <v-card-actions>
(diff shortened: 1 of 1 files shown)


frontend/src/components/dialogs/SecretDialog.vue
```diff
@@ -60,6 +60,9 @@ SPDX-License-Identifier: Apache-2.0
       </v-card-text>
       <v-alert :value="!isCreateMode && relatedShootCount > 0" type="warning" tile>
         This secret is used by {{relatedShootCount}} clusters. The new secret should be part of the same account as the one that gets replaced.
+      </v-alert>
+       <v-alert :value="!isCreateMode && relatedShootCount > 0" type="warning" tile>
+        Shoots will only start using the new secret after they got reconciled. Therefore, wait until all  {{relatedShootCount}} Shoots using the secret are reconciled before you disable the old secret in your infrastructure account. Otherwise the Shoots will no longer function.
```
petersutter, line 65:
```suggestion
        Shoots will only start using the new secret after they got reconciled. Therefore, wait until all {{relatedShootCount}} Shoots using the secret are reconciled before you disable the old secret in your infrastructure account. Otherwise the Shoots will no longer function.
```

petersutter (approved):
/lgtm

frontend/src/components/dialogs/SecretDialog.vue
```diff
@@ -60,6 +60,9 @@ SPDX-License-Identifier: Apache-2.0
       </v-card-text>
       <v-alert :value="!isCreateMode && relatedShootCount > 0" type="warning" tile>
         This secret is used by {{relatedShootCount}} clusters. The new secret should be part of the same account as the one that gets replaced.
+      </v-alert>
+       <v-alert :value="!isCreateMode && relatedShootCount > 0" type="warning" tile>
+        Shoots will only start using the new secret after they got reconciled. Therefore, wait until all {{relatedShootCount}} Shoots using the secret are reconciled before you disable the old secret in your infrastructure account. Otherwise the Shoots will no longer function.
```
grolu, line 65:
```suggestion
        Clusters will only start using the new secret after they got reconciled. Therefore, wait until all clusters using the secret are reconciled before you disable the old secret in your infrastructure account. Otherwise the clusters will no longer function.
```

grolu:
We discussed this and we will approve the double alert for now, we can enhance this later on. However, I don't like the text.
- We do not use the term `Shoot` in the UI, we always refer to them as `Clusters`
- The information about affected clusters is duplicate. Moreover, if you want to have it in this alert, too it should be more like in the existing alert. `Therefore, wait until all 1 Shoots using the secret are reconciled` would sound a bit odd to be...
I made a proposal. WDYT?

BeckerMax:
/squash
Sounds good. Thanks for your feedback!

PR #1143, opened 2021-12-08
Add check-docforge step in the .ci/check

Kristian-ZH:
Add check-docforge step in the .ci/check

Fixes #

@holgerkoser 

**Release note**:

```feature operator
a new check-docforge step will be executed on each PR
```

### DIFF
--- a/.ci/check-docforge
+++ b/.ci/check-docforge
@@ -1,0 +1,39 @@
+#!/usr/bin/env bash
+
+# SPDX-FileCopyrightText: 2021 SAP SE or an SAP affiliate company and Gardener contributors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+echo "Check Docforge Manifest"
+
+docforgeVersion="v0.21.0"
+docCommitHash="12922abea4c5afa207cc9c09b18414cd2348890c"
+
+repoPath="$(readlink -f $(dirname ${0})/..)"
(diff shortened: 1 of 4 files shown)


scripts/check-docforge.sh
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env bash
```
holgerkoser, line 1: We use an alpine based nodejs image for the check step. There is no bash available. Is it possible to change this to 
```
#!/usr/bin/env sh
```
or do you need bash?

.ci/check
```diff
@@ -19,3 +19,6 @@ cd "${SOURCE_PATH}"
 # the cache folder was to be modified or
 # the checksums of the packages are not consistent
 yarn install --immutable --immutable-cache --check-cache
+
+# Run check-docforge script
+${SOURCE_PATH}/scripts/check-docforge.sh
```
holgerkoser, line 24:
```suggestion
./scripts/check-docforge.sh
```
We are already in the ${SOURCE_PATH} directroy.

scripts/check-docforge.sh
Kristian-ZH: I thon no, seems working with sh also

holgerkoser (approved):
/lgtm

PR #1145, opened 2021-12-13
[master] Improve socket connection and subscription handling

holgerkoser:

This PR improve socket connection and subscription handling.

* **Handling of `socket.io` connection error events** 
If a socket connection is denied by an authorization error a logout is forced and the user is redirected to the login page.

* **Handling of subscriptions error / done events**
Sometimes the shoots loading bar is spinning forever because the no subscription done event is received in the client or the namespaces do not match. The handling of subscriptions is a little bit simplified but the main logic is not not changed. Still an subscription error  does not stop the loading bar. 

Fixes #

**Release note**:

```bugfix user

```

### DIFF
--- a/backend/lib/io.js
+++ b/backend/lib/io.js
@@ -27,8 +27,11 @@ function socketAuthentication (nsp) {
       logger.debug('Socket %s authenticated (user %s)', socket.id, user.id)
       next()
     } catch (err) {
-      const level = /^ERR_JWT/.test(err.code) ? 'info' : 'error'
-      logger[level]('Socket %s authentication failed: "%s"', socket.id, err.message)
+      logger.error('Socket %s authentication failed: %s', socket.id, err)
+      if (isHttpError(err)) {
+        const { statusCode, code } = err
+        err.data = JSON.stringify({ statusCode, code })
+      }
       next(err)
     }
   })
@@ -48,7 +51,7 @@ function getUserFromSocket (socket) {
 
 function joinRoom (socket, room) {
   socket.join(room)
-  logger.debug('Socket %s subscribed to room "%s"', socket.id, room)
+  logger.debug('Socket %s subscribed to room %s', socket.id, room)
 }
 
 function leaveRooms (socket, predicate = () => true) {
--- a/backend/lib/io.js
+++ b/backend/lib/io.js
@@ -30,8 +30,7 @@ function socketAuthentication (nsp) {
     } catch (err) {
       logger.error('Socket %s authentication failed: %s', socket.id, err)
       if (isHttpError(err)) {
-        const { statusCode, code } = err
-        err.data = JSON.stringify({ statusCode, code })
+        err.data = _.pick(err, ['statusCode', 'code'])
       }
       next(err)
     }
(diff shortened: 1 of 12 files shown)


backend/lib/io.js
```diff
@@ -27,8 +28,11 @@ function socketAuthentication (nsp) {
       logger.debug('Socket %s authenticated (user %s)', socket.id, user.id)
       next()
     } catch (err) {
-      const level = /^ERR_JWT/.test(err.code) ? 'info' : 'error'
-      logger[level]('Socket %s authentication failed: "%s"', socket.id, err.message)
+      logger.error('Socket %s authentication failed: %s', socket.id, err)
+      if (isHttpError(err)) {
+        const { statusCode, code } = err
+        err.data = JSON.stringify({ statusCode, code })
```
holgerkoser, line 34: Remove JSON.stringify

holgerkoser: ✔️

backend/lib/io.js
```diff
@@ -72,75 +75,72 @@ function leaveCommentsRooms (socket) {
...
-  /* join current rooms */
-  if (!_.isArray(namespacesAndFilters)) {
-    return
+  // subscribe for all namespaces or a single namespace
+  if (Array.isArray(namespaces)) {
+    namespace = '_all'
+  } else if (namespace) {
+    namespaces = [namespace]
+  } else {
+    assert.fail('Either property namespaces or namespace is required')
```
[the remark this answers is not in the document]
holgerkoser: All errors thrown by the `assert` module are instances of the [`AssertionError`](https://nodejs.org/api/assert.html#class-assertassertionerror) class. The reason why I have used an assertion here is that this assertion should never ever fail. Currently this method is called from the following two shoot different shoot subscriptions:

**Subscribe shoots for all namespaces the user is authorized**
```js
  socket.on('subscribeAllShoots', async ({ filter }) => {
    try {
      const user = getUserFromSocket(socket)
      const projectList = await projects.list({ user })
      const namespaces = _.map(projectList, 'metadata.namespace')
      
      if (await authorization.isAdmin(user)) {
        subscribeShootsAdmin(socket, { namespaces, filter, user })
      } else {
        subscribeShoots(socket, { namespaces, filter, user })
      }
    } catch (err) {
      ...
    }
  })
```
In this case the `namespaces` are passed to the `subscribeShoots` function.

**Subscribe shoots for a single namespace**
```js
  socket.on('subscribeShoots', async ({ namespace, filter }) => {
    ...
    try {
      const user = getUserFromSocket(socket)
      ...
      subscribeShoots(socket, { namespace, filter, user })
    } catch (err) {
      ...
    }
  })
```
In this case the `namespace` is passed to the `subscribeShoots` function.

It is important that we do not await the result of call to `subscribeShoots` in both cases here. This means the surrounding try catch block does not catch the `AssertionError`. But this is the intended behavior of this assertion. If for whatever reasons the assertion fails the developer should immediately be informed about the fact that the assertions is not met anymore. The unit tests will never be successful with a failing assertion here because these lines are covered by our tests.

petersutter (approved):
/lgtm

grolu (approved):
/lgtm

I think we should add a release note

PR #1148, opened 2021-12-21
gardenctl-v2 commands

petersutter:
Adds support for `gardenctl-v2` commands. 

The user has the option to choose to see the legacy `gardenctl` commands or the `gardenctl-v2` commands. 
The user can also choose the preferred `shell`. The commands are then adapted to the selected shell.

Screenshot 2021-12-21 at 15 58 29

Screenshot 2021-12-21 at 16 00 07
Screenshot 2021-12-21 at 15 58 38

Screenshot 2021-12-21 at 15 58 47

Screenshot 2021-12-21 at 15 59 38

The default selected `gardenctl` version (legacy vs. `gardenctl-v2`) and the default shell can be set using the chart's `values.yaml`.

Example `values.yaml`:
```yaml
frontendConfig:
  # gardenctl - configure the default settings for the gardenctl commands
  gardenctl:
    legacyCommands: false # false to show gardenctl-v2 commands by default, true to show the legacy gardenctl commands. Can be overwritten by the user.
    shell: bash # default shell. Should be one of bash, fish, powershell, zsh. Can be overwritten by the user.
```

Fixes #

**Release note**:

```feature user
You can now switch between legacy `gardenctl` and `gardenctl-v2` commands on the `My Account` page. You can also choose your preferred shell for which the commands will be adapted.
```

### DIFF
--- a/backend/lib/routes/info.js
+++ b/backend/lib/routes/info.js
@@ -51,3 +51,23 @@ async function fetchGardenerVersion () {
     logger.warn(`Could not fetch gardener version. Error: ${err.message}`)
   }
 }
+
+router.route('/identity')
+  .get(async (req, res, next) => {
+    try {
+      const clusterIdentity = await fetchIdentity()
+      res.send({ clusterIdentity })
+    } catch (err) {
+      next(err)
+    }
+  })
+
+async function fetchIdentity () {
+  const {
+    data: {
+      'cluster-identity': clusterIdentity
+    }
+  } = await dashboardClient.core.configmaps.get('kube-system', 'cluster-identity')
+
+  return clusterIdentity
+}
--- a/frontend/src/components/GardenctlSettings.vue
+++ b/frontend/src/components/GardenctlSettings.vue
@@ -1,0 +1,78 @@
+<!--
+SPDX-FileCopyrightText: 2021 SAP SE or an SAP affiliate company and Gardener contributors
+
+SPDX-License-Identifier: Apache-2.0
+ -->
+
+<template>
+  <v-card flat class="mx-2 mt-2">
+    <v-card-text class="pt-0">
+      <v-radio-group
+        v-model="legacyCommands"
+        label="Select Version"
+        hint="Choose for which gardenctl version the commands should be displayed"
+        persistent-hint
+        @change="onChangeLegacyCommands"
+        class="pb-4"
+      >
+        <v-radio
+          label="Legacy Gardenctl"
+          :value="true"
+          color="primary"
+        ></v-radio>
+        <v-radio
+          label="Gardenctl-v2"
+          :value="false"
+          color="primary"
+        ></v-radio>
+      </v-radio-group>
+      <v-select
+        v-show="!legacyCommands"
+        color="primary"
+        item-color="primary"
+        label="Shell"
+        :items="['bash', 'fish', 'powershell', 'zsh']"
+        hint="Choose for which shell the commands should be displayed"
+        persistent-hint
+        v-model="shell"
+        @input="onInputShell"
+      ></v-select>
+    </v-card-text>
+  </v-card>
+</template>
+
+<script>
+import { mapActions } from 'vuex'
+
+export default {
+  data () {
+    return {
+      legacyCommands: false,
+      shell: 'bash'
+    }
+  },
+  methods: {
+    ...mapActions([
+      'refreshGardenctlOptions'
+    ]),
+    onInputShell () {
+      this.store()
+    },
+    onChangeLegacyCommands () {
+      this.store()
+    },
+    store () {
+      const gardenctlOptions = {
+        legacyCommands: this.legacyCommands,
+        shell: this.shell
+      }
+      this.$store.commit('SET_GARDENCTL_OPTIONS', gardenctlOptions)
+    }
+  },
+  async mounted () {
+    const { legacyCommands, shell } = await this.refreshGardenctlOptions()
+    this.legacyCommands = legacyCommands
+    this.shell = shell
+  }
+}
+</script>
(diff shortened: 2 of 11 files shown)


frontend/src/components/ShootDetails/GardenctlCommands.vue
```diff
@@ -125,7 +222,21 @@ export default {
...
+        case 'fish':
+          return `eval (${command})`
+        default:
+          return `eval $(${command})`
+      }
     }
+  },
+  async mounted () {
+    await this.initializeClusterIdentity()
+    await this.refreshGardenctlOptions()
```
grolu, line 239: This method invokes `SET_GARDENCTL_OPTIONS` in the store. This method updates the local storage with the current value. This means, that when the user first opens an arbitrary  shoot details page, the local storage gets initialized / updated with the values, even if the user never actually modified the settings. While this is not a real problem, I just wanted to point out that if the admin somehow misses to set the config values and users did open a detail shoot page once, the commands are not updated to the admin provided values (which the admin may provide shortly after) even if the user did not actively change them.

grolu (approved):
/lgtm

Besides the drawback I mentioned in my comment, this PR lgtm.

backend/lib/routes/info.js
```diff
@@ -51,3 +51,23 @@ async function fetchGardenerVersion () {
     logger.warn(`Could not fetch gardener version. Error: ${err.message}`)
   }
 }
+
+router.route('/identity')
```
holgerkoser, line 55: Since this is static information we should merge it into the frontend `/config.json`.

frontend/src/components/GardenctlSettings.vue
```diff
@@ -0,0 +1,78 @@
...
+    },
+    onChangeLegacyCommands () {
+      this.store()
+    },
+    store () {
+      const gardenctlOptions = {
+        legacyCommands: this.legacyCommands,
+        shell: this.shell
+      }
+      this.$store.commit('SET_GARDENCTL_OPTIONS', gardenctlOptions)
```
holgerkoser, line 69:
```js
...mapMutations({
   setGardenctlOptions: SET_GARDENCTL_OPTIONS
})
```

frontend/src/components/GardenctlSettings.vue
```diff
@@ -0,0 +1,78 @@
...
+    ...mapActions([
+      'refreshGardenctlOptions'
+    ]),
+    onInputShell () {
+      this.store()
+    },
+    onChangeLegacyCommands () {
+      this.store()
+    },
+    store () {
```
holgerkoser, line 64: Bad naming. Property name `store` is very misleading for a method. Better would be `updateGardenctlOptions` I think.

frontend/src/components/GardenctlSettings.vue
```diff
@@ -0,0 +1,78 @@
...
+    return {
+      legacyCommands: false,
+      shell: 'bash'
+    }
+  },
+  methods: {
+    ...mapActions([
+      'refreshGardenctlOptions'
+    ]),
+    onInputShell () {
```
holgerkoser, line 58: Why don't we directly bind to `store` or better `updateGardenctlOptions`?

frontend/src/components/GardenctlSettings.vue
```diff
@@ -0,0 +1,78 @@
...
+      this.store()
+    },
+    onChangeLegacyCommands () {
```
holgerkoser, line 61: dito

frontend/src/store/index.js
```diff
@@ -1258,6 +1268,12 @@ const getters = {
       })
       return filter(workerGroups, 'expirationDate')
     }
+  },
+  clusterIdentity (state) {
+    return state.clusterIdentity
+  },
+  gardenctlOptions (state) {
+    return state.gardenctlOptions
```
holgerkoser, line 1276: These getters make no sense. Delete them.

frontend/src/store/index.js
```diff
@@ -1660,6 +1676,41 @@ const actions = {
...
+      const { data: { clusterIdentity } } = await getClusterIdentity()
+      commit('SET_CLUSTER_IDENTITY', clusterIdentity)
+
+      return clusterIdentity
+    } catch (err) {
+      // eslint-disable-next-line no-console
+      console.warn(`Failed to get cluster-identity ${err.message}`)
+    }
+  },
+  refreshGardenctlOptions ({ commit }) {
```
holgerkoser, line 1698: the `state` property is missing

frontend/src/store/index.js
```diff
@@ -1660,6 +1676,41 @@ const actions = {
   setColorScheme ({ commit }, colorScheme) {
     commit('SET_COLOR_SCHEME', colorScheme)
     return state.colorScheme
+  },
+  async initializeClusterIdentity ({ commit, dispatch, state }) {
```
holgerkoser, line 1680: add to frontend config

frontend/src/store/index.js
holgerkoser, line 1698: I don't get the idea behind this action. Is it a getter for the default `gardenctlOptions` or is it used to initialize the localStorage. Why don't we use the same approach as with `global/color-scheme` and initialize the store in the root app.

frontend/src/components/GardenctlSettings.vue
```diff
@@ -0,0 +1,78 @@
...
+  </v-card>
+</template>
+
+<script>
+import { mapActions } from 'vuex'
+
+export default {
+  data () {
+    return {
+      legacyCommands: false,
```
holgerkoser, line 50: I think it would be better to have computed properties for `legacyCommands` and `shell`. The store could provider getters with defaults for both and in the setter we could commit the new values.

holgerkoser (approved):
/lgtm

PR #1149, opened 2021-12-28
Proposal for changes to PR gardenctl-v2 commands

holgerkoser:
Proposal for changes to PR #1148.
* treat `clusterIdentity` like `apiServerUrl` and allow configuration. Fallback to configmap `kube-system/cluster-identity` if the value is not configured.
* tread `gardenctlOptions` like all other properties in localStorage and do not keep copies in component data in order to avoid refreshing/sync in lifecycle hooks.
* move `frontendConfig[Handler]` from middleware into an own file and pass dashboardClient for the fallback code.
* fix vetur config

Fixes #

**Release note**:

```bugfix user

```

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -50,5 +50,5 @@
   "jestrunner.enableYarnPnpSupport": true,
   "jestrunner.jestCommand": "scripts/jest",
   "vetur.useWorkspaceDependencies": true,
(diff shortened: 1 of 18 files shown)


petersutter (approved):
/lgtm, thanks

I only have one minor change request for a typo in a comment

PR #1152, opened 2022-01-25
Gardenctl commands: Update store only on settings change

petersutter:
The store should only be updated when the user changes the `gardenctl` settings on the `My Account` page. 
This way, when the defaults change, it will be applied for all users that haven't actively changed the settings.

Precondition:

**Release note**:

```bugfix user

```


grolu (approved):
/lgtm

holgerkoser (approved):
/lgtm

PR #1158, opened 2022-02-16
Remove shell

petersutter:
With https://github.com/gardener/gardenctl-v2/pull/86 it is now easier to target a (shoot) cluster. Running `eval $(gardenctl provider-env bash)` after running the target command is not required any more to set the `KUBECONFIG` env variable. Instead, it only has to be run once (e.g. by including it into the shell profile). See more details here https://github.com/gardener/gardenctl-v2/blob/master/docs/concepts/cli_interaction.md#behavior-with-symlink

Fixes #

**Release note**:

```bugfix user

```


grolu (approved):
/lgtm

PR #1170, opened 2022-03-08
Improve ShootList performance

grolu:
Do not register new shoots individually for Vue reactivity. For a batch of new shoots, register reactivity by re-initializing the shoot state store object with all new shoots at one.
This significantly improves performance when initially navigating to a huge shoot list, e.g. all projects.  

Fixes #

**Release note**:

```feature user
Improve initial cluster list rendering performance for large cluster lists
```

### DIFF
--- a/frontend/src/store/modules/shoots/index.js
+++ b/frontend/src/store/modules/shoots/index.js
@@ -324,15 +324,32 @@ function setFilteredItems (state, rootState, rootGetters) {
   state.filteredShoots = items
 }
 
-const putItem = (state, newItem) => {
-  const item = findItem(state)(newItem.metadata)
-  if (item !== undefined) {
-    if (item.metadata.resourceVersion !== newItem.metadata.resourceVersion) {
-      Vue.set(state.shoots, keyForShoot(item.metadata), assign(item, newItem))
+const putItem = (state, item) => {
+  const shoot = findItem(state)(item.metadata)
+  const key = keyForShoot(item.metadata)
+  if (shoot !== undefined) {
+    if (shoot.metadata.resourceVersion !== item.metadata.resourceVersion) {
+      Vue.set(state.shoots, key, assign(shoot, item))
     }
   } else {
+    item.info = undefined // register property to ensure reactivity
+    Vue.set(state.shoots, key, item)
+  }
+}
+
+const putItems = (state, newItems) => {
+  const newShoots = {}
+  forEach(newItems, newItem => {
     newItem.info = undefined // register property to ensure reactivity
-    Vue.set(state.shoots, keyForShoot(newItem.metadata), newItem)
+    const key = keyForShoot(newItem.metadata)
+    newShoots[key] = newItem
+  })
+
+  // Do not use Vue.set but assign all new items at once by ceating a whole new object
+  // This significally improves performance while ensuring reactivity for the new items
+  state.shoots = {
+    ...state.shoots,
+    ...newShoots
   }
 }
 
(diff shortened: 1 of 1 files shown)


frontend/src/store/modules/shoots/index.js
```diff
@@ -324,15 +324,32 @@ function setFilteredItems (state, rootState, rootGetters) {
...
+  const key = keyForShoot(item.metadata)
+  if (shoot !== undefined) {
+    if (shoot.metadata.resourceVersion !== item.metadata.resourceVersion) {
+      Vue.set(state.shoots, key, assign(shoot, item))
     }
   } else {
+    item.info = undefined // register property to ensure reactivity
+    Vue.set(state.shoots, key, item)
+  }
+}
```
holgerkoser, line 338: We should not mix a renaming which has nothing to do with the performance improvement in the same commit. It is very confusing when reviewing the changes.

frontend/src/store/modules/shoots/index.js
```diff
@@ -366,14 +383,23 @@ const mutations = {
...
+            objectsToAdd.push(event.object)
+            setFilteredItemsRequired = true
+          }
+          break
         case 'MODIFIED':
-          if (rootState.namespace !== '_all' ||
-            !onlyShootsWithIssues ||
-            onlyShootsWithIssues === shootHasIssue(event.object)) {
-            // Do not add healthy shoots when onlyShootsWithIssues=true, this can happen when toggeling flag
+          if (shouldAddShoot(event.object)) {
```
holgerkoser, line 402: The old shoot had an issue but the new shoot does not have an issue anymore. We do not update the shoot in this case.

frontend/src/store/modules/shoots/index.js
```diff
@@ -366,14 +383,23 @@ const mutations = {
...
+    const shouldAddShoot = (shoot) => {
+      // Do not add healthy shoots when onlyShootsWithIssues=true, this can happen when toggeling flag
+      return rootState.namespace !== '_all' ||
+      !onlyShootsWithIssues ||
+      onlyShootsWithIssues === shootHasIssue(shoot)
+    }
     forEach(events, event => {
       switch (event.type) {
         case 'ADDED':
+          if (shouldAddShoot(event.object)) {
```
holgerkoser, line 396: What is the performance improvement of this change compared to the improvement by replacing the dict of shoots at once. Why do we send events for new shoots if we registered only for shoots without issues? I am not sure if this is needed.

frontend/src/store/modules/shoots/index.js
```diff
@@ -366,14 +383,23 @@ const mutations = {
   HANDLE_EVENTS (state, { rootState, rootGetters, events }) {
     const onlyShootsWithIssues = get(state, 'shootListFilters.onlyShootsWithIssues', true)
     let setFilteredItemsRequired = false
+    const objectsToAdd = []
+    const shouldAddShoot = (shoot) => {
+      // Do not add healthy shoots when onlyShootsWithIssues=true, this can happen when toggeling flag
+      return rootState.namespace !== '_all' ||
+      !onlyShootsWithIssues ||
+      onlyShootsWithIssues === shootHasIssue(shoot)
+    }
```
holgerkoser, line 392: If the implementation would follow what is described in the comment the comment would be not necessary.
```suggestion
    const doNotAddShoot = shoot => rootState.namespace === '_all' && onlyShootsWithIssues && !shootHasIssue(shoot)
```

frontend/src/store/modules/shoots/index.js
```diff
@@ -324,15 +324,32 @@ function setFilteredItems (state, rootState, rootGetters) {
...
+    if (shoot.metadata.resourceVersion !== item.metadata.resourceVersion) {
+      Vue.set(state.shoots, key, assign(shoot, item))
     }
   } else {
+    item.info = undefined // register property to ensure reactivity
+    Vue.set(state.shoots, key, item)
+  }
+}
+
+const putItems = (state, newItems) => {
```
holgerkoser, line 340:
```js
function putItems (state, newItems) {
  // Do not use Vue.set but assign all new items at once by ceating a whole new object
  // This significally improves performance while ensuring reactivity for the new items
  state.shoots = {
    ...state.shoots,
    ...keyBy(newItems, item => keyForShoot(item.metadata))
  }
}
```

PR #1179, opened 2022-03-23
Allow configuration of node CLI options

holgerkoser:
Allow configuration of the [node CLI options](https://nodejs.org/dist/latest-v16.x/docs/api/cli.html) as command arguments in order to optimize garbage collector.

Node (V8) uses a lazy garbage collector. With a default limit of about 1.5 GB, it sometimes waits until it absolutely has to before reclaiming unused memory. If your memory usage is increasing, it may not be a leak, but the usual lazy behavior of Node. Therefore, it may be useful to optimize the garbage collector behavior using V8 flags.
This is especially important if your application is running in an environment with less than 1.5 GB of available memory (e.g  512MB docker container).

Starting from Node.js v8.x is is possible to increase memory limit via the `--max-old-space-size` option flag. In production we can now avoid “Heap out of memory” errors if we increase the limit of the container and the Node.js limit accordingly. The value of `--max-old-space-size` should not higher than 90% of the available memory.

These settings in `values.yaml` 
```yaml
...
resources:
  limits:
    cpu: 1000m
    memory: 2Gi
  requests:
    cpu: 500m
    memory: 512Mi
nodeOptions: ["--optimize-for-size", "--max-old-space-size=1792", "--gc-interval=100"]
...
```

will result in the following deployment:
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: gardener-dashboard
  namespace: garden
  labels:
    app: gardener-dashboard
spec:
  replicas: 1
  selector:
    matchLabels:
      app: gardener-dashboard
  template:
    metadata:
      labels:
        app: gardener-dashboard
        release: release-name
    spec:
      ...
      containers:
        - name: gardener-dashboard
          image: "eu.gcr.io/gardener-project/gardener/dashboard:1.26.0"
          args:
          - --optimize-for-size
          - --max-old-space-size=1792
          - --gc-interval=100
          - server.js
          resources:
            limits:
              cpu: 1000m
              memory: 2Gi
            requests:
              cpu: 500m
              memory: 512Mi

``` 

Fixes #

**Release note**:

```bugfix operator
Allow configuration of the node CLI options in order to optimize garbage collector.
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -55,7 +55,6 @@ RUN apk add --no-cache tini libstdc++
 WORKDIR /usr/src/app
 
 ENV NODE_ENV "production"
-ENV NODE_OPTIONS "--require /usr/src/app/.pnp.js"
 
 ARG PORT=8080
 ENV PORT $PORT
@@ -72,4 +71,4 @@ EXPOSE $PORT
 
 VOLUME ["/home/node"]
 
-ENTRYPOINT [ "/sbin/tini", "--", "node", "server" ]
+ENTRYPOINT [ "/sbin/tini", "--", "node", "--require", "/usr/src/app/.pnp.js", "server.js"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -71,4 +71,5 @@ EXPOSE $PORT
 
 VOLUME ["/home/node"]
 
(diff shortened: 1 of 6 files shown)


grolu (approved):
/lgtm

petersutter (approved):
/lgtm

PR #1181, opened 2022-03-24
Use apiServerCaData for service account kubeconfigs

### LINKED ISSUE #1163: Make ca field optional for service account kubeconfigs (robot)
For service account kubeconfigs (robot), do not write `ca` field in case the garden cluster api-server presents a browser trusted certificate, e.g. when configured using a let's encrypt tls certificate.

**Why is this needed**:
The CA of the garden cluster can then be rotated without that the user having to download the kubeconfig for the service account again.

grolu:
Do not use ca.crt from service accounts to provide `certificate-authority-data` for API server endpoint in generated service account kubeconfigs. The ca can differ, depending on configured `apiServerUrl` in dashboard config. Use configured `apiServerCaData` instead (like we already do for user config). If no `apiServerCaData` is configured, do not include a ca in the kubeconfig. This allows to secure the API server with trusted certificates that do not require a ca to be included in the kubeconfig. These certificates can be rotated without the need to regenerate and update downloaded kubeconfigs.

Fixes #1163

**Release note**:

```feature operator
Use configured `apiServerCaData` value for generated service account kubeconfigs
```


packages/kube-config/lib/index.js
```diff
@@ -152,11 +152,17 @@ function fromKubeconfig (input) {
 }
 
 function dumpKubeconfig ({ user, context = 'default', cluster = 'garden', namespace, token, server, caData }) {
+  const certificateAuthorityData = caData
+    ? {
+        'certificate-authority-data': caData
+      }
+    : {}
+
```
holgerkoser, line 160:
```suggestion
  const cluster = {}
  if (caData) {
    cluster['certificate-authority-data'] = caData
  }
```

packages/kube-config/lib/index.js
```diff
@@ -152,11 +152,17 @@ function fromKubeconfig (input) {
...
   return new Config({
     clusters: [{
       name: cluster,
       cluster: {
-        'certificate-authority-data': caData,
+        ...certificateAuthorityData,
```
holgerkoser, line 165:
```suggestion
        ...cluster,
```

packages/kube-config/lib/index.js
holgerkoser: And we should rename the function arguments to userName, contextName and clusterName to avoid missunderstandings.

packages/kube-config/__tests__/kube-config.test.js
```diff
@@ -93,6 +93,20 @@ describe('kube-config', () => {
...
+        const kubeconfig = dumpKubeconfig({
+          user,
+          namespace,
+          token,
+          server: server.origin
+        })
+        expect(kubeconfig).toBeDefined()
+      } catch (err) {
+        throw new Error('No exception expected')
+      }
```
holgerkoser, line 107:
```suggestion
        const kubeconfig = yaml.safeLoad(dumpKubeconfig({
          user,
          namespace,
          token,
          server: server.origin
        }))
        expect(kubeconfig.clusters).toHaveLength(1)
        const cluster = kubeconfig.clusters[0]
        expect(cluster).not.toHaveProperty('certificate-authority-data')
```

petersutter (approved):
/lgtm

could you also kindly fix 
```
    {{- if .Values.apiServerCaData }}
    apiServerCaData: {{ .Values.apiServerCaData }}
    {{- else if .Values.apiServerCa }}
    apiServerCaData: {{ b64enc .Values.apiServerCa }}
```
in the charts (nothing that was changed by you, but we could clean it up with this PR)

PR #1182, opened 2022-03-25
Enh/improve custom cidr handling

### LINKED ISSUE #1172: Random CIDRs when creating AWS shoot with non-default values
**What happened**:

I have tried to create an AWS shoot in the Dashboard with non default CIDR ranges. I have set `.spec.provider.infrastructureConfig.networks.vpc.cidr` to `10.180.0.0/16` and `spec.networking.nodes` to `10.180.0.0/16`. When selecting all three zones the following is generated (no matter when CIDRs are specified for the VPC):

```
      networks:
        vpc:
          cidr: 10.180.0.0/16
        zones:
          - name: eu-west-1a
            workers: 10.250.0.0/19
            public: 10.250.32.0/20
            internal: 10.250.48.0/20
          - name: eu-west-1b
            workers: 10.250.64.0/19
            public: 10.250.96.0/20
            internal: 10.250.112.0/20
          - name: eu-west-1c
            workers: 10.250.128.0/19
            public: 10.250.160.0/20
            internal: 10.250.176.0/20
```

I expect the following to be generated:
```
      networks:
        vpc:
          cidr: 10.180.0.0/16
        zones:
          - name: eu-west-1a
            workers: 10.180.0.0/19
            public: 10.180.32.0/20
            internal: 10.180.48.0/20
          - name: eu-west-1b
            workers: 10.180.64.0/19
            public: 10.180.96.0/20
            internal: 10.180.112.0/20
          - name: eu-west-1c
            workers: 10.180.128.0/19
            public: 110.180.160.0/20
            internal: 10.180.176.0/20
```

See above

This might be broken for the other cloud providers as well.

**Environment**:

any

grolu:
Added `defaultNodesCIDR` to dashboard configuration to allow to overwrite the default nodes CIDR.
Improved worker CIDR generation for new clusters when using non default node networks. The dashboard now tries to generate the networks so that they match the custom node network configuration. It also ensures that `spec.networking.nodes` is set to the CIDR configured in the cloud provider extension network configuration.

Fixes #1172

**Release note**:

```feature user
Improved worker CIDR generation for new clusters when using non default node networks
```
```feature operator
Default node CIDR can be overwritten by setting the new `frontend.defaultNodesCIDR` value in the Dashboard configuration
```


frontend/src/utils/createShoot.js
```diff
@@ -335,3 +343,18 @@ export function getWorkerProviderConfig (infrastructureKind) {
...
   }
 }
+
+export function alignAndReturnNodeCIDR (shootResource) {
+  const networks = get(shootResource, 'spec.provider.infrastructureConfig.networks')
+  const networkCIDR = get(networks, 'vpc.cidr', get(networks, 'vnet.cidr', get(networks, 'workers')))
+  const nodeCIDR = get(shootResource, 'spec.networking.nodes')
+  if (networkCIDR && networkCIDR !== nodeCIDR) {
+    // networkCIDR is the leading CIDR as it is more likely to be modified by the user
+    // nodeCIDR needs to be a subset or equal to networkCIDR. Usually they should be equal
```
petersutter, line 353: > nodeCIDR needs to be a subset or equal to networkCIDR

what if nodeCIDR is a subset of the networkCIDR, in this case you would just reset it to the networkCIDR

frontend/src/utils/createShoot.js
```diff
@@ -335,3 +343,18 @@ export function getWorkerProviderConfig (infrastructureKind) {
     }
   }
 }
+
+export function alignAndReturnNodeCIDR (shootResource) {
```
holgerkoser, line 347:
```suggestion
export function alignNodeCIDR (shootResource) {
```

holgerkoser (approved):
/lgtm

petersutter (approved):
/lgtm

PR #1185, opened 2022-03-31
Migrate to use the networking.k8s.io/v1 API

### LINKED ISSUE #1183: Migrate to use the networking.k8s.io/v1 API version
**What happened**:

> The v1.22 release will stop serving the following deprecated API versions in favor of newer and more stable API versions:
> Ingress in the extensions/v1beta1 API version will no longer be served
> Migrate to use the [networking.k8s.io/v1beta1](http://networking.k8s.io/v1beta1) API version, available since v1.14. Existing persisted data can be retrieved/updated via the new version.

-> Migrate to use the networking.k8s.io/v1 API version

https://github.com/gardener/dashboard/blob/master/packages/kube-client/lib/groups.js#L36-L40
https://github.com/gardener/dashboard/blob/master/packages/kube-client/lib/resources/Extensions.js#L14-L22

**Environment**:

grolu:

Fixes #1183

**Release note**:

```bugfix user

```

### DIFF
--- a/backend/lib/services/terminals/terminalBootstrap.js
+++ b/backend/lib/services/terminals/terminalBootstrap.js
@@ -217,10 +217,15 @@ function replaceIngressApiServer (client, { name = TERMINAL_KUBE_APISERVER, name
           paths: [
             {
               backend: {
-                serviceName,
-                servicePort: 443
+                service: {
+                  name: serviceName,
+                  port: {
+                    number: 443
+                  }
+                }
               },
-              path: '/'
+              path: '/',
+              pathType: 'ImplementationSpecific'
             }
           ]
         }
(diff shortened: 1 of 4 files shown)


petersutter (approved):
/lgtm

petersutter:
please also add a release note which k8s versions are not supported anymore for the terminal bootstrapping

backend/lib/services/terminals/terminalBootstrap.js
```diff
@@ -217,10 +217,15 @@ function replaceIngressApiServer (client, { name = TERMINAL_KUBE_APISERVER, name
...
+                service: {
+                  name: serviceName,
+                  port: {
+                    number: 443
+                  }
+                }
               },
-              path: '/'
+              path: '/',
+              pathType: 'ImplementationSpecific'
```
holgerkoser, line 228:
```suggestion
              pathType: 'Prefix'
```
We want a clearly defined pathType. It should not be implementation specific up to the IngressClass.

> - **ImplementationSpecific:** With this path type, matching is up to the IngressClass. Implementations can treat this as a separate pathType or treat it identically to Prefix or Exact path types.
> - **Exact:** Matches the URL path exactly and with case sensitivity.
> - **Prefix:** Matches based on a URL path prefix split by /. Matching is case sensitive and done on a path element by element basis. A path element refers to the list of labels in the path split by the / separator. A request is a match for path p if every p is an element-wise prefix of p of the request path.

PR #1190, opened 2022-04-05
Enh/readniness chips

### LINKED ISSUE #1115: Colorblind users cannot identify cluster readiness issues
The cluster readiness chips can have different states:

Screen Shot 2021-10-13 at 11 10 50

However, it is difficult / not possible for colorblind users to differentiate between ok (cyan), error (red), progressing (blue) or unknown (grey) state. Maybe an icon can help to identify issues here.
There may be more places that we should improve, we can discuss this here in this issue.

**Why is this needed**:
Improve accessibility for users with handicaps.

### LINKED ISSUE #1159: Improve success/error indication
E.g. in case a user has a red–green color blindness it is hard to distinguish the error / success state of the readiness chips.
E.g. the fine lines could be made thicker in case of an error or an ❗ icon could be added, similar to the status. 
It was reported that the green on the status is recognized better.

**Why is this needed**:

### LINKED ISSUE #1174: Sort readiness chips

The readiness chips should be sorted. Simply sorting ascending should be sufficient.

**Why is this needed**:

When comparing similar clusters it is confusing because chips are in different locations.
image

grolu:
This PR adds several improvements to the readiness chips
- Sort readiness chips by name
- Left align readiness tags, this makes not much difference for regular projects but improves overview if additional chips are present for only some of the clusters
- Fill chips that have an error state (to make them more prominent)
- Icons for readiness chips

Fixes #1174 Fixes #1159 Fixes #1115

Screen Shot 2022-04-05 at 12 24 29

**Release note**:

```feature user
Several improvements for the readiness chips
- Sort readiness chips by name
- Left align readiness tags, this improves overview if additional chips are present for only some of the clusters
- Fill chips that have an error state (to make them more prominent)
- Added icons for readiness chips that have a state other than success
```


grolu:
We could also show icons only for chips that show a status other than _success_

Screen Shot 2022-04-05 at 12 24 29

WDYT?

petersutter:
> We could also show icons only for chips that show a status other than _success_

yes I would prefer this

petersutter:
please also add a release note

frontend/src/components/StatusTag.vue
```diff
@@ -197,12 +193,15 @@ export default {
...
+      }
+      return this.color
+    },
     visible () {
       if (!this.isAdmin) {
-        const { type } = this.condition
-
-        const conditionMetadata = this.conditionMetadataFromType(type)
-        return !get(conditionMetadata, 'showAdminOnly', false)
+        return !get(this.condition, 'showAdminOnly', false)
```
petersutter, line 204: this does not work anymore. Now, as non-admin I can see the control plane chip

grolu: fixed

petersutter (approved):
/lgtm

frontend/src/components/StatusTag.vue
```diff
@@ -131,60 +118,69 @@ export default {
...
-        title: this.tag.name,
+        title: this.condition.name,
         status: this.chipStatus,
-        description: this.tag.description,
-        userErrorCodeObjects: filter(objectsFromErrorCodes(this.tag.codes), { userError: true })
+        description: this.condition.description,
+        userErrorCodeObjects: filter(objectsFromErrorCodes(this.condition.codes), { userError: true })
+      }
+    },
+    chipIcon () {
```
holgerkoser, line 127: The chip icon needs more margin. We could also use a `v-badge` for this use case. It will also safe some space.

frontend/src/components/StatusTags.vue
```diff
@@ -35,11 +46,55 @@ export default {
   },
   mixins: [shootItem],
   computed: {
-    filteredConditions () {
+    ...mapState([
+      'conditionCache'
```
holgerkoser, line 50: It is very strange to use the store as cache. Of course, I know not this PR. We should only keep known conditions in the store and create defaults on the fly. Since the getter already does caching.

frontend/src/components/StatusTags.vue
```diff
@@ -35,11 +46,55 @@ export default {
...
+      })
+
+      return sortBy(conditions, 'shortName')
+    }
+  },
+  methods: {
+    ...mapMutations([
+      'setCondition'
+    ]),
+    conditionMetadataFromType (type) {
```
holgerkoser, line 73: Conditions do not have metadata. Metadata are data that describe the data. The name pattern you are using here is a pattern that we have introduced for getters that return a function with parameters. But here we have a method. So the name should be `getCondition`.

grolu: renamed to `getCondition`

frontend/src/components/StatusTags.vue
grolu: we were both right - there are hard-coded conditions in the store but they can be extended by `known conditions` using our dashboard configuration. Therefore I suggest to leave it for now as it requires more effort and is not related to this PR.

PR #1192, opened 2022-04-11
Fix isShootStatusHibernated computed prop

grolu:
Fixed a bug that caused an error for regular users (access to non existent isShootStatusHibernated computed property)

Fixes #

**Release note**:

```bugfix user
NONE
```


petersutter (approved):
/lgtm

PR #1195, opened 2022-04-19
Support service accounts with more than one secret

### LINKED ISSUE #1196: Support Gardener service accounts with more than one secret (token)
[1] Start considering "consistent" Gardener project service accounts (robot users) that have more than one secret assigned. Use always the latest created secret when generating the kubeconfig file offered for display/download via the Dashboard (Members/Service Accounts tole) . 
[2] For backward compatibility, the user should still be able to execute "Rotate Service Account Secret" but because of this change Gardener Dashboard should ensure the deletion of all existing for the service account secrets, not just one.

**Why is this needed**:
Certificate rotation for a Gardener landscape requires also rotation of the KeyPair configured for [service-account-signing-key-file](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/) of the Garden virtual cluster. Direct KeyPair replacement can cause disruption in the consumer processes if such replacement is made by the Gardener service provider without giving a gratis period for the kubeconfig/token rotation in the consumer automation processes. 
In order to give a chance to the Gardener service provider to offer their service without disruption for the authentication to the API calls, Gardener open source project has to ensure a smooth rotation of the service-account-signing-key-file KeyPair. This will require the support of several secrets for one and the same service account (robot user).

grolu:
Adds support for service account with multiple service account tokens.
Always returns the newest token. Allows to rotate service accounts with multiple tokens (delete all secrets).

Fixes #1196

**Release note**:

```bugfix user
Service Accounts with more than one secret are now handled correctly
```

### DIFF
--- a/backend/__fixtures__/secrets.js
+++ b/backend/__fixtures__/secrets.js
@@ -193,11 +193,10 @@ const secrets = {
       }
     })
   },
-  getServiceAccountSecret (namespace, name, creationTimestamp = '2019-03-13T13:11:36Z') {
+  getServiceAccountSecret (namespace, name) {
     return getSecret({
       name,
       namespace,
-      creationTimestamp,
       data: {
         namespace,
         token: name
@@ -259,10 +258,6 @@ const mocks = {
         }
         if (endsWith(name, '.kubeconfig')) {
           const item = secrets.getShootSecret(namespace, name)
-          return Promise.resolve(item)
-        }
-        if (name === 'newest-secret') {
-          const item = secrets.getServiceAccountSecret(namespace, name, '2022-03-13T13:11:36Z')
           return Promise.resolve(item)
         }
         if (/-token-[a-f0-9]{5}$/.test(name)) {
--- a/backend/__fixtures__/secrets.js
+++ b/backend/__fixtures__/secrets.js
@@ -193,10 +193,11 @@ const secrets = {
       }
     })
   },
-  getServiceAccountSecret (namespace, name) {
+  getServiceAccountSecret (namespace, name, creationTimestamp = '2019-03-13T13:11:36Z') {
     return getSecret({
       name,
       namespace,
+      creationTimestamp,
       data: {
         namespace,
         token: name
@@ -258,6 +259,10 @@ const mocks = {
         }
         if (endsWith(name, '.kubeconfig')) {
           const item = secrets.getShootSecret(namespace, name)
+          return Promise.resolve(item)
+        }
+        if (name === 'newest-secret') {
+          const item = secrets.getServiceAccountSecret(namespace, name, '2022-03-13T13:11:36Z')
           return Promise.resolve(item)
         }
         if (/-token-[a-f0-9]{5}$/.test(name)) {
--- a/backend/__fixtures__/serviceaccounts.js
+++ b/backend/__fixtures__/serviceaccounts.js
@@ -14,10 +14,8 @@ const { getTokenPayload } = require('./auth')
 
 const serviceAccountList = [
   getServiceAccount({ namespace: 'garden-foo', name: 'robot', createdBy: 'foo@example.org' }),
-  getServiceAccount({ namespace: 'garden-foo', name: 'robot-two-secrets', createdBy: 'foo@example.org' }),
   getServiceAccount({ namespace: 'garden-foo', name: 'robot-nomember' }),
   getServiceAccount({ namespace: 'garden-bar', name: 'robot', createdBy: 'bar@example.org' }),
-  getServiceAccount({ namespace: 'garden-foo', name: 'robot-missing-secret', createdBy: 'foo@example.org' }),
   getServiceAccount({ namespace: 'term-host-1', name: 'term-attach-1' }),
   getServiceAccount({ namespace: 'term-host-2', name: 'term-attach-2' }),
   getServiceAccount({ namespace: 'term-host-3', name: 'term-attach-3' })
@@ -29,7 +27,7 @@ function getServiceAccount ({
   createdBy = 'admin@example.org',
   creationTimestamp = '2020-01-01T00:00:00Z'
 }) {
-  const serviceAccount = {
+  return {
     metadata: {
       name,
       namespace,
@@ -42,14 +40,6 @@ function getServiceAccount ({
       name: [name, 'token', toHex(name).substring(0, 5)].join('-')
     }]
   }
-  if (name.includes('missing-secret')) {
-    serviceAccount.secrets.push({ name: 'missing-secret' })
-  }
-  if (name.includes('two-secrets')) {
-    serviceAccount.secrets.push({ name: 'newest-secret' })
-  }
-
-  return serviceAccount
 }
 
 const serviceaccounts = {
--- a/backend/__fixtures__/serviceaccounts.js
+++ b/backend/__fixtures__/serviceaccounts.js
@@ -14,8 +14,10 @@ const { getTokenPayload } = require('./auth')
 
 const serviceAccountList = [
   getServiceAccount({ namespace: 'garden-foo', name: 'robot', createdBy: 'foo@example.org' }),
+  getServiceAccount({ namespace: 'garden-foo', name: 'robot-two-secrets', createdBy: 'foo@example.org' }),
   getServiceAccount({ namespace: 'garden-foo', name: 'robot-nomember' }),
   getServiceAccount({ namespace: 'garden-bar', name: 'robot', createdBy: 'bar@example.org' }),
+  getServiceAccount({ namespace: 'garden-foo', name: 'robot-missing-secret', createdBy: 'foo@example.org' }),
   getServiceAccount({ namespace: 'term-host-1', name: 'term-attach-1' }),
   getServiceAccount({ namespace: 'term-host-2', name: 'term-attach-2' }),
   getServiceAccount({ namespace: 'term-host-3', name: 'term-attach-3' })
@@ -27,7 +29,7 @@ function getServiceAccount ({
   createdBy = 'admin@example.org',
   creationTimestamp = '2020-01-01T00:00:00Z'
 }) {
-  return {
+  const serviceAccount = {
     metadata: {
       name,
       namespace,
@@ -40,6 +42,14 @@ function getServiceAccount ({
       name: [name, 'token', toHex(name).substring(0, 5)].join('-')
     }]
   }
+  if (name.includes('missing-secret')) {
+    serviceAccount.secrets.push({ name: 'missing-secret' })
+  }
+  if (name.includes('two-secrets')) {
+    serviceAccount.secrets.push({ name: 'newest-secret' })
+  }
+
+  return serviceAccount
 }
 
 const serviceaccounts = {
--- a/backend/lib/services/members/MemberManager.js
+++ b/backend/lib/services/members/MemberManager.js
@@ -200,7 +200,7 @@ class MemberManager {
       .thru(promises => Promise.allSettled(promises))
       .value()
 
-    this.handleRejectedPromises(results)
+    this.constructor.getFulfilledValues(results)
   }
 
   async getKubeconfig (item) {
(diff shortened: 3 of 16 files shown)


backend/lib/services/members/MemberManager.js
```diff
@@ -187,35 +187,38 @@ class MemberManager {
...
-        }
-      })
-      .head()
-      .get('name')
+      .map('name')
       .value()
-    await this.client.core.secrets.delete(namespace, name)
+
+    _.forEach(names, async name => {
+      await this.client.core.secrets.delete(namespace, name)
```
grolu, line 203: we did not have error handling but I think we should add some here - e.g. accept 404 errors as the user could add non existent secrets to the service account but return an error to the frontend in all other cases so that the user has an indication that the rotation did not succeed (completely). WDYT?

backend/lib/services/members/MemberManager.js
```diff
@@ -187,35 +187,38 @@ class MemberManager {
...
 
   async getKubeconfig (item) {
     const { namespace, name } = Member.parseUsername(item.id)
-    const secretName = _
+    const secretNames = _
       .chain(item)
       .get('extensions.secrets')
+      .map('name')
+      .value()
+    const secrets = await this.client.core.secrets.list(namespace)
```
holgerkoser, line 214:
```js
const secrets = await _
      .chain(item)
      .get('extensions.secrets')
      .map(name => this.client.core.secrets.get(namespace, name))
      .thru(Promise.all)
      .value()
const secret = _
      .chain(secrets)
      .orderBy(['metadata.creationTimestamp'], ['desc'])
      .head()
      .value()
```

holgerkoser, line 214: At least we should use a field selector to reduce the resultSet of the list call: 
```sh
k get secrets --field-selector="type=kubernetes.io/service-account-token"
```

backend/lib/services/members/MemberManager.js
```diff
@@ -187,35 +187,38 @@ class MemberManager {
...
-          throw new UnprocessableEntity(`ServiceAccount ${namespace} has more than one secret`)
-        }
-      })
-      .head()
-      .get('name')
+      .map('name')
       .value()
-    await this.client.core.secrets.delete(namespace, name)
+
+    _.forEach(names, async name => {
```
holgerkoser, line 202:
```js
const results = await _
      .chain(item)
      .get('extensions.secrets')
      .map(name => this.client.core.secrets.delete(namespace, name))
      .thru(Promise.allSettled)
      .value()
const errors = _
      .chain(results)
      .filter(['status', 'rejected'])
      .map('reason')
      .filter(err => !isHttpError(err) || err.statusCode !== 404)
      .value()
if (errors.length) {
     throw errors[0] // or whatever we want to do in this case with the errors
}
```

donistz (changes requested):
Do you want to invalidate all existing secrets of the service account

frontend/src/components/messages/RotateServiceAccountSecret.vue
```diff
@@ -6,7 +6,7 @@ SPDX-License-Identifier: Apache-2.0
 
 <template>
   <div>
-    Do you want to rotate the secret of service account <em>{{name}}</em>?<br /> Attention: The current kubeconfig credentials will be revoked
+    Do you want to rotate all secrets of the service account <em>{{name}}</em>?<br /> Attention: The current kubeconfig credentials will be revoked
```
donistz, line 9: Do you want to invalidate all secrets of the service account <em>{{name}}</em>?<br /> Attention: The current kubeconfig credentials will be revoked

backend/lib/services/members/MemberManager.js
holgerkoser: If an error occurs inside the `forEach` loop it will cause an `unhandledRejection` error which leads to process exit. All the function will be called at the same time. The usage of async / await in this case has no effect at all. It would have been much better to use 
```js
for (const name of names) {
  await this.client.core.secrets.delete(namespace, name)
}
```
instead to run the delete one by one and forward the rejection to the outer function.

backend/lib/services/members/MemberManager.js
holgerkoser: All errors are forwarded to the outer async function here but you are right we not handle 404 like we do for delete serviceaccount.

donistz (approved):
/lgtm

backend/lib/services/members/MemberManager.js
```diff
@@ -234,6 +241,19 @@ class MemberManager {
...
+    const errors = _
+      .chain(results)
+      .filter(['status', 'rejected'])
+      .map('reason')
+      .filter(err => !isHttpError(err) || err.statusCode !== 404)
+      .value()
+
+    if (errors.length) {
+      throw new Error(errors)
+    }
```
holgerkoser, line 254:
```suggestion
    if (errors.length === 1) {
      throw errors[0]
    } else if (errors.length > 1) {
      const message = errors.map(err => err.message).join('\n')
      throw new AggregateError(errors, message)
    }
```

backend/lib/services/members/MemberManager.js
```diff
@@ -234,6 +241,19 @@ class MemberManager {
...
+  }
```
holgerkoser, line 255: This should be a standalone function or a static method.
```suggestion
  static getFulfilledValues (results) {
    const errors = _
      .chain(results)
      .filter(['status', 'rejected'])
      .map('reason')
      .filter(err => !isHttpError(err) || err.statusCode !== 404)
      .value()

    if (errors.length === 1) {
      throw errors[0]
    } else if (errors.length > 1) {
      const message = errors.map(err => err.message).join('\n')
      throw new AggregateError(errors, message)
    }
    
    return _
      .chain(results)
      .filter(['status', 'fulfilled'])
      .map('value')
      .value()
  }
```

backend/test/services.members.spec.js
```diff
@@ -428,24 +428,26 @@ describe('services', function () {
...
         })
 
-        it('should not delete a service account secret if there is one more secret attached', async function () {
+        it('should delete all service account secrets if there is more than one secret attached', async function () {
           const id = 'system:serviceaccount:garden-foo:robot-multiple'
           const item = memberManager.subjectList.get(id)
-          await expect(memberManager.deleteServiceAccountSecret(item)).rejects.toThrow(UnprocessableEntity)
+          await memberManager.deleteServiceAccountSecrets(item)
+          expect(client.core.secrets.delete).toBeCalledWith('garden-foo', 'secret-1')
+          expect(client.core.secrets.delete).toBeCalledWith('garden-foo', 'secret-2')
```
holgerkoser, line 450:
```suggestion
          expect(client.core.secrets.delete).toHaveBeenNthCalledWith(1, 'garden-foo', 'secret-1')
          expect(client.core.secrets.delete).toHaveBeenNthCalledWith(2, 'garden-foo', 'secret-2')
```

backend/__fixtures__/serviceaccounts.js
```diff
@@ -14,8 +14,10 @@ const { getTokenPayload } = require('./auth')
 
 const serviceAccountList = [
   getServiceAccount({ namespace: 'garden-foo', name: 'robot', createdBy: 'foo@example.org' }),
+  getServiceAccount({ namespace: 'garden-foo', name: 'robot-two-secrets', createdBy: 'foo@example.org' }),
```
holgerkoser, line 17: I think it is better to not test these cases `missing-secret` and `two-secretes` in acceptance tests. We should rather test these cases in a unit test of the member manager.

backend/__fixtures__/serviceaccounts.js
```diff
@@ -40,6 +42,14 @@ function getServiceAccount ({
       name: [name, 'token', toHex(name).substring(0, 5)].join('-')
     }]
   }
+  if (name.includes('missing-secret')) {
```
holgerkoser, line 45: This function should not be changed

backend/__fixtures__/secrets.js
```diff
@@ -260,6 +261,10 @@ const mocks = {
           const item = secrets.getShootSecret(namespace, name)
           return Promise.resolve(item)
         }
+        if (name === 'newest-secret') {
```
holgerkoser, line 264: We should not make assumptions on certain names. Only assumptions on k8s conventions or gardener conventions.

backend/lib/services/members/MemberManager.js
```diff
@@ -187,35 +187,42 @@ class MemberManager {
...
+      .map(({ name }) => this.client.core.secrets.get(namespace, name))
+      .thru(promises => Promise.allSettled(promises))
+      .value()
+
+    this.handleRejectedPromises(results)
+
+    const secret = _
+      .chain(results)
+      .filter(['status', 'fulfilled'])
+      .map('value')
```
holgerkoser, line 221:
```suggestion
    const values = this.constructor.getFulfilledValues(results)

    const secret = _
      .chain(values)
```

backend/lib/services/members/MemberManager.js
```diff
@@ -187,35 +187,42 @@ class MemberManager {
...
-      })
-      .head()
-      .get('name')
+      .get('extensions.secrets')
+      .map(({ name }) => this.client.core.secrets.delete(namespace, name))
+      .thru(promises => Promise.allSettled(promises))
       .value()
-    await this.client.core.secrets.delete(namespace, name)
+
+    this.handleRejectedPromises(results)
```
holgerkoser, line 203:
```suggestion
  this.constructor.getFulfilledValues(results)
```

backend/__fixtures__/serviceaccounts.js
grolu: moved to unit tests

backend/test/services.members.spec.js
```diff
@@ -181,24 +207,29 @@ describe('services', function () {
...
           return Promise.resolve(_.set(body, 'metadata.creationTimestamp', 'now'))
         }),
-        delete: jest.fn().mockImplementation((namespace, name) => {
-          const item = _.find(serviceAccounts, { metadata: { name, namespace } })
-          if (!item) {
-            return Promise.reject(createError(404))
-          }
-          return Promise.resolve(item)
-        }),
+        delete: mockFindOrReject(serviceAccounts),
```
holgerkoser, line 227:
```suggestion
        delete: jest.fn().mockImplementation(findObjectFn(serviceAccounts)),
```

backend/test/services.members.spec.js
```diff
@@ -181,24 +207,29 @@ describe('services', function () {
...
         mergePatch: jest.fn().mockResolvedValue()
       }
       client.core.secrets = {
-        delete: jest.fn().mockResolvedValue()
+        delete: mockFindOrReject(secrets),
+        get: mockFindOrReject(secrets)
```
holgerkoser, line 232:
```suggestion
        delete: jest.fn().mockImplementation(findObjectFn(secrets)),
        get: jest.fn().mockImplementation(findObjectFn(secrets))
```

backend/test/services.members.spec.js
```diff
@@ -442,12 +473,26 @@ describe('services', function () {
...
+      describe('#getKubeconfig', function () {
+        it('should return kubeconfig with token of newest secret', async function () {
+          const id = 'system:serviceaccount:garden-foo:robot-multiple'
+          const item = memberManager.subjectList.get(id)
+          const kubeConfig = await memberManager.getKubeconfig(item)
+
+          expect(client.core.secrets.get).toHaveBeenNthCalledWith(1, 'garden-foo', 'secret-1')
+          expect(client.core.secrets.get).toHaveBeenNthCalledWith(2, 'garden-foo', 'secret-2')
+          expect(client.core.secrets.get).toHaveBeenNthCalledWith(3, 'garden-foo', 'missing-secret')
+          expect(kubeConfig).toContain('secret-2')
```
holgerkoser, line 495: parse kubeconfig and check the correct value

petersutter (approved):
/lgtm

holgerkoser (approved):
\lgtm

grolu:
>Do we still need this after our discussion today?

I think it makes sense to merge it now. The current handling is incomplete otherwise. We can change it later when we finally agreed on how to move forward.

PR #1198, opened 2022-05-02
Show available LB classes

### LINKED ISSUE #1057: Show available LB classes in the shoot detail pane for CCloud clusters
Endusers would like to see which LB classes are available, when they create services of type LB.

Domains with more than one subnet will usually have several loadbalancer classes configured in their `cloudprofile` , e.g.
```yaml
  providerConfig:
    constraints:
      floatingPools:
      - name: FloatingIP*
        region: eu-de-2
        loadBalancerClasses:
        - floatingSubnetName: FloatingIP-internet-*
          name: internet
        - floatingSubnetName: FloatingIP-sap-*
          name: sap
          purpose: default
      - name: FloatingIP*
        region: eu-nl-1
        loadBalancerClasses:
        - floatingSubnetName: FloatingIP-internet-*
          name: internet
        - floatingSubnetName: FloatingIP-sap-*
          name: intranet
          purpose: default
```
There can be region specific differences in the LB classes.
Furthermore the user can specify the LB classes to use in the shoot manifest via.
```yaml
spec
  provider:
    type: openstack
    controlPlaneConfig:
        loadBalancerClasses:
        - name: sap
          floatingSubnetName: FloatingIP-sap-*
```
If the user defines the LB classes in the shoot mainfest, they completely replace the ones defined in the
cloudprofile.

After shoot creation the available LB classes should be displayed in the dashboard and read either from the shoot mainfest if configured, or from the cloudprofile.
When reading the Floating Pool and region from the cloudprofile, the best match must be used

In addition there is always one default LB class. This class should be marked in the UI. The following rules apply:
* If only one LB class is defined, it's the default

In case of several LB classes:
If no `purpose: default` is given, we look for a LB class name called `default`.
If this is also not present the first LB class from the list is taken as default

grolu:

Fixes #1057

**Release note**:

```bugfix user
Fixed an issue caused by wildcard Floating IP names not mapped correctly, therefore available load balancer classes were not shown on shoot details page. Also the UI did not flag the default load balancer class
```


frontend/src/utils/wildcard.js
```diff
@@ -0,0 +1,71 @@
...
+      endsWithWildcard,
+      customWildcard,
+      isWildcard: startsWithWildcard || endsWithWildcard || customWildcard,
+      regex: new RegExp('^' + pattern + '$')
+    }
+  })
+}
+
+export function bestMatchForString (wildCardObjects, wildCardString) {
+  const matches = filter(wildCardObjects, (item) => {
```
holgerkoser, line 48: We should first sort the pattern and then find the first match and return it.

holgerkoser, line 48: We should first sort the pattern and then find the first match and return it.

frontend/src/utils/wildcard.js
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    return {
+      value,
+      originalValue: item,
+      startsWithWildcard,
+      endsWithWildcard,
+      customWildcard,
+      isWildcard: startsWithWildcard || endsWithWildcard || customWildcard,
+      regex: new RegExp('^' + pattern + '$')
```
holgerkoser, line 42: This is a problem since we do not escape the pattern. This could lead to wrong matching and will maybe occur in security scan.

holgerkoser, line 42: This is a problem since we do not escape the pattern. This could lead to wrong matching and will maybe occur in security scan.

grolu: removed the regex

grolu: removed the regex

frontend/src/utils/wildcard.js
grolu: I did not change this method for now. If you have a proposal please go ahead.

frontend/src/utils/wildcard.js
```diff
@@ -39,14 +41,32 @@ export function wildcardObjectsFromStrings (wildcardStrings) {
       endsWithWildcard,
       customWildcard,
       isWildcard: startsWithWildcard || endsWithWildcard || customWildcard,
-      regex: new RegExp('^' + pattern + '$')
+      test: function (value) {
```
holgerkoser, line 44:
```suggestion
      test (value) {
```

frontend/src/utils/wildcard.js
```diff
@@ -9,27 +9,29 @@
...
+import includes from 'lodash/includes'
 
 export function wildcardObjectsFromStrings (wildcardStrings) {
   return map(wildcardStrings, item => {
     let startsWithWildcard = false
     let endsWithWildcard = false
     let customWildcard = false
-    const value = trim(item, '*')
-    let pattern = value
+    let value = item
```
holgerkoser, line 21: rename item to value and remove this line

frontend/src/utils/wildcard.js
```diff
@@ -39,14 +41,32 @@ export function wildcardObjectsFromStrings (wildcardStrings) {
       endsWithWildcard,
       customWildcard,
       isWildcard: startsWithWildcard || endsWithWildcard || customWildcard,
-      regex: new RegExp('^' + pattern + '$')
+      test: function (value) {
+        if (!this.isWildcard) {
+          return value === this.originalValue
```
holgerkoser, line 46:
```suggestion
          return value === this.value
```
We don't need the originalValue

frontend/src/utils/wildcard.js
```diff
@@ -39,14 +41,32 @@ export function wildcardObjectsFromStrings (wildcardStrings) {
...
+        }
+        if (this.startsWithWildcard && this.endsWithWildcard) {
+          return includes(value.substring(1, value.length - 1), this.value)
+        }
+        if (this.startsWithWildcard) {
+          return endsWith(value.substring(1, value.length), this.value)
+        }
+        if (this.endsWithWildcard) {
+          return startsWith(value.substring(0, value.length - 1), this.value)
+        }
```
holgerkoser, line 59:
```suggestion
        if (this.startsWithWildcard && this.endsWithWildcard) {
          return includes(value, this.value)
        }
        if (this.startsWithWildcard) {
          return endsWith(value, this.value)
        }
        if (this.endsWithWildcard) {
          return startsWith(value, this.value)
        }
```
I am not the expert here but I think this is wrong.

frontend/tests/unit/utils.wildcard.spec.js
```diff
@@ -0,0 +1,104 @@
...
+        expect(wildcardObject.endsWithWildcard).toBe(true)
+        expect(wildcardObject.customWildcard).toBe(false)
+        expect(wildcardObject.test('__Barz')).toBe(false)
+        expect(wildcardObject.test('Barz__')).toBe(false)
+        expect(wildcardObject.test('__Barz__')).toBe(true)
+        expect(wildcardObject.test('Barz')).toBe(false)
+        expect(wildcardObject.test('x')).toBe(false)
+      })
+
+      it('should detect customWildcard', () => {
```
holgerkoser, line 62:
```suggestion

      it('should detect endsWith wildcard', () => {
        const wildcardObject = wildcardObjects[0]
        expect(wildcardObject.value).toEqual('Foo')
        expect(wildcardObject.isWildcard).toBe(true)
        expect(wildcardObject.endsWithWildcard).toBe(true)
        expect(wildcardObject.startsWithWildcard).toBe(false)
        expect(wildcardObject.customWildcard).toBe(false)
        expect(wildcardObject.test('Foo__')).toBe(true)
        expect(wildcardObject.test('__Foo__')).toBe(false)
        expect(wildcardObject.test('Foo')).toBe(true)
        expect(wildcardObject.test('x')).toBe(false)
      })

      it('should detect startsWith wildcard', () => {
        const wildcardObject = wildcardObjects[1]
        expect(wildcardObject.value).toEqual('Bar')
        expect(wildcardObject.isWildcard).toBe(true)
        expect(wildcardObject.startsWithWildcard).toBe(true)
        expect(wildcardObject.endsWithWildcard).toBe(false)
        expect(wildcardObject.customWildcard).toBe(false)
        expect(wildcardObject.test('__Bar')).toBe(true)
        expect(wildcardObject.test('__Bar__')).toBe(false)
        expect(wildcardObject.test('Bar')).toBe(true)
        expect(wildcardObject.test('x')).toBe(false)
      })

      it('should detect startsWith and endsWith wildcard', () => {
        const wildcardObject = wildcardObjects[2]
        expect(wildcardObject.value).toEqual('Barz')
        expect(wildcardObject.isWildcard).toBe(true)
        expect(wildcardObject.startsWithWildcard).toBe(true)
        expect(wildcardObject.endsWithWildcard).toBe(true)
        expect(wildcardObject.customWildcard).toBe(false)
        expect(wildcardObject.test('__Barz')).toBe(true)
        expect(wildcardObject.test('Barz__')).toBe(true)
        expect(wildcardObject.test('__Barz__')).toBe(true)
        expect(wildcardObject.test('Barz')).toBe(true)
        expect(wildcardObject.test('x')).toBe(false)
      })

      it('should detect custom wildcard', () => {
```

frontend/src/utils/wildcard.js
grolu: I'm also not the expert. I think this has been wrong in the previous implementation. At least we show the suffix / prefix as required on the UI

frontend/src/utils/wildcard.js
grolu: we need to keep the originalValue for line 39

frontend/src/utils/wildcard.js
```diff
@@ -0,0 +1,91 @@
...
```
holgerkoser, line 59: Ok. Now I have understood why you have done it like this. For me the following code would be more intuitive. But I am also fine with the current code if you add some comments. I have have not understood why you remove only one character from start and/or end like it is done for the pattern. 
```suggestion
        const index = value.indexOf(this.value)
        const lengthDifference = value.length - this.value.length

        if (this.startsWithWildcard && this.endsWithWildcard) {
          // included but does not start or end with
          return index > 0 && index < lengthDifference
        }
        if (this.startsWithWildcard) {
          // ends with and not equal
          return index === lengthDifference && lengthDifference > 0
        }
        if (this.endsWithWildcard) {
          // starts with and not equal
          return index === 0 && lengthDifference > 0
        }
```

holgerkoser (approved):
/lgtm

PR #1199, opened 2022-05-06
Improve container runtime handling when machine image changes

### LINKED ISSUE #1111: OS update recommendation should take CRI support into account
As a user, I'd like to get recommendations for updating my OS only if it supports the currently selected container runtime for my Shoot.

**Why is this needed**:
In the future, we might release e.g. a new gardenlinux version without support for `docker`. Shoot owners who have `docker` configured for their worker should not get the recommendation to update to this version.

grolu:
- When selecting a machine image that does not support the currently selected container runtime, the Dashboard does no longer implicitly set the container runtime to something else
- If the user switches back to another image that supports the previously selected runtime it is restored
- This also applies for new clusters, I think it makes sense as the user may not notice it otherwise
- Container Runtime is now a required field also for not new clusters. There might still exist some very old worker groups without a container runtime set explicitly. However, modifying those would cause Gardener to add this now required field automatically anyway

Fixes #1111

**Release note**:

```bugfix user
When selecting a machine image that does not support the currently selected container runtime, the Dashboard does no longer implicitly set the container runtime to something else
```

### DIFF
--- a/frontend/src/components/ShootWorkers/ContainerRuntime.vue
+++ b/frontend/src/components/ShootWorkers/ContainerRuntime.vue
@@ -34,8 +34,8 @@ SPDX-License-Identifier: Apache-2.0
 </template>
 
 <script>
-import { requiredIf } from 'vuelidate/lib/validators'
-import { getValidationErrors, defaultCriNameByKubernetesVersion } from '@/utils'
+import { required } from 'vuelidate/lib/validators'
+import { getValidationErrors } from '@/utils'
 import find from 'lodash/find'
 import map from 'lodash/map'
 import get from 'lodash/get'
@@ -44,15 +44,9 @@ import unset from 'lodash/unset'
 import includes from 'lodash/includes'
 import isEmpty from 'lodash/isEmpty'
 
-const validationErrors = {
-  criName: {
-    required: 'Container Runtime is required'
-  }
-}
-
 const validations = {
   criName: {
-    required: requiredIf('criNameIsRequired')
+    required
   }
 }
 
@@ -69,20 +63,23 @@ export default {
     kubernetesVersion: {
       type: String,
       required: true
-    },
-    criNameIsRequired: {
-      type: Boolean,
-      default: false
     }
   },
   data () {
     return {
-      validationErrors,
       valid: undefined
     }
   },
   validations,
   computed: {
+    validationErrors () {
+      const originalCriName = this.originalCri.name
+      return {
+        criName: {
+          required: `The selected machine image does not support previously selected Container Rumtime '${originalCriName}'`
+        }
+      }
+    },
     criNames () {
       return map(this.machineImageCri, 'name')
     },
@@ -138,12 +135,20 @@ export default {
   mounted () {
     this.$v.$touch()
     this.validateInput()
+    this.originalCri = get(this.worker, 'cri', {})
   },
   watch: {
     criNames (criNames) {
       if (!includes(criNames, this.criName)) {
-        this.criName = defaultCriNameByKubernetesVersion(criNames, this.kubernetesVersion)
-        this.onInputCriName()
+        if (includes(criNames, this.originalCri.name)) {
+          // Restore original container runtime settings if possible
+          set(this.worker, 'cri', this.originalCri)
+        } else {
+          // To not implicitly change to other container runtime
+          set(this.worker, 'cri', undefined)
+        }
+        this.$v.$touch()
+        this.validateInput()
       }
     }
   }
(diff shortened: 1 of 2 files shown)


frontend/src/components/ShootWorkers/ContainerRuntime.vue
```diff
@@ -46,13 +45,7 @@ import isEmpty from 'lodash/isEmpty'
...
 const validationErrors = {
   criName: {
-    required: 'Container Runtime is required'
-  }
-}
-
-const validations = {
-  criName: {
-    required: requiredIf('criNameIsRequired')
+    validCriName: 'The selected machine image does not support the selected Container Rumtime'
```
petersutter, line 48:
```suggestion
    validCriName: 'The selected machine image does not support the selected container runtime'
```

frontend/src/components/ShootWorkers/ContainerRuntime.vue
```diff
@@ -69,22 +62,34 @@ export default {
...
     }
   },
-  validations,
+  validations () {
+    return this.validators
+  },
   computed: {
-    criNames () {
-      return map(this.machineImageCri, 'name')
+    criItems () {
```
holgerkoser, line 77: I think it is wrong to do the validation in a getter and rely in the validator on that cched value for valid. Why do we need these items objects at all?
```html
    <v-select
      color="primary"
      item-color="primary"
      :items="criNames"
      :error-messages="getErrorMessages('criName')"
      @blur="$v.criName.$touch()"
      v-model="criName"
      label="Container Runtime"
    ></v-select>
```
```js
export default {
   ...
  data () {
    return {
      valid: undefined,
      validationErrors
    }
  },
  validations () {
    return this.validators
  },
  computed: {
    validCriNames () {
      return map(this.machineImageCri, 'name')
    },
    criNames () {
      const names = this.validCriNames
      if (!includes(names, this.criName)) {
        names.push(this.criName)
      }
      return names
    },
    criContainerRuntimeTypes () {
      const containerRuntime = find(this.machineImageCri, ['name', this.criName])
      const ociRuntimes = get(containerRuntime, 'containerRuntimes', [])
      return map(ociRuntimes, 'type')
    },
    criName: {
      get () {
        return get(this.worker, 'cri.name')
      },
      set (value) {
        this.$set(this.worker, 'cri', {
          ...this.worker.cri,
          name: value
        })
        this.selectedCriContainerRuntimeTypes = undefined
        this.$v.criName.$touch()
        this.validateInput()
      }
    },
    selectedCriContainerRuntimeTypes: {
      get () {
        const criContainerRuntimes = get(this.worker, 'cri.containerRuntimes')
        return criContainerRuntimes
          ? map(criContainerRuntimes, 'type')
          : undefined
      },
      set (value) {
        if (!isEmpty(value)) {
          const criContainerRuntimes = map(value, type => ({ type }))
          set(this.worker, 'cri.containerRuntimes', criContainerRuntimes)
        } else {
          unset(this.worker, 'cri.containerRuntimes')
        }
      }
    },
    validators () {
      return {
        criName: {
          validCriName: value => {
            return includes(this.validCriNames, value)
          }
        }
      }
    }
  },
  methods: {
    getErrorMessages (field) {
      return getValidationErrors(this, field)
    },
    validateInput () {
      if (this.valid !== !this.$v.$invalid) {
        this.valid = !this.$v.$invalid
        this.$emit('valid', { id: this.worker.id, valid: this.valid })
      }
    }
  },
  mounted () {
    this.$v.$touch()
    this.validateInput()
  },
  watch: {
    criNames () {
      this.$v.$touch()
      this.validateInput()
    }
  }
}
```

frontend/src/components/ShootWorkers/ContainerRuntime.vue
```diff
@@ -69,22 +62,34 @@ export default {
...
+      const criItems = map(this.machineImageCri, cri => {
+        return {
+          value: cri.name,
+          text: cri.name,
+          valid: true
+        }
+      })
+      if (!includes(criItems, this.criName)) {
```
holgerkoser, line 85: Shouldn't we disable that invalid value?

frontend/src/components/ShootWorkers/ContainerRuntime.vue
```diff
@@ -138,13 +153,12 @@ export default {
   mounted () {
     this.$v.$touch()
     this.validateInput()
+    this.originalCri = get(this.worker, 'cri', {})
```
holgerkoser, line 156: We don't use this porperty?

frontend/src/components/ShootWorkers/ContainerRuntime.vue
grolu: Yes, you are right of course, I changed it to disabled and removed the valid property. However we still have to do the error validation manually as far as I know

frontend/src/components/ShootWorkers/ContainerRuntime.vue
grolu: removed leftover

frontend/src/components/ShootWorkers/ContainerRuntime.vue
grolu: changed the validation as suggested

petersutter (approved):
/lgtm

PR #1200, opened 2022-05-06
Enh/remove node forge

holgerkoser:
* Use the built-in class [X509Certificate](https://nodejs.org/dist/latest-v18.x/docs/api/crypto.html#class-x509certificate) to parse the client certificate and extract the commonName (CN) from the subject.  
* Bump [gtoken](https://www.npmjs.com/package/gtoken) to version `5.3.2`

Fixes https://github.com/gardener/dashboard/security/dependabot/51
Closes #1187

The `X509Certificate` class returns the subject as  string and not as an object. Unfortunately the format ist specified in the documentation. I have found the following [example](https://www.geeksforgeeks.org/node-js-x509-issuer-property/) code.

**Release note**:

```bugfix user
NONE
```


grolu (approved):
/lgtm

petersutter (approved):
/lgtm

PR #1206, opened 2022-05-12
Add "deletion scheduled" to shoot status

### LINKED ISSUE #849: Add "deletion scheduled" to shoot status
It would be helpful if a currently creating cluster that I'm deleting would show that the cluster is marked for deletion in the respective progression status.
Currently a short notification bubble is shown at the bottom right which vanishes after a second.
The status could show i.e. "Create Processing, Deletion Scheduled" or something similar.
image

**Why is this needed**:
We repeatedly find users that delete a cluster that is e.g. in creation, but don't have "trust" that gardener will really delete that cluster eventually and therefore they start to shoot down resources in their IaaS account manually (i.e. because they want to recreate the shoot but only have too low quotas).
A clear text in the shoots status signalling that it is marked/scheduled for deletion could help in some situations.
(The trigger to display that text could be simply the existence of a `deletionTimestamp` on the shoot resource)

/cc @vlerenc

grolu:
Screen Shot 2022-05-12 at 18 02 58

Fixes #849

**Release note**:

```feature user
The cluster status now shows if a cluster has been marked for deletion
```


petersutter (approved):
/lgtm, also the caption is good in my opinion

holgerkoser (approved):
/lgtm

PR #1234, opened 2022-06-03
Fetch service account token using TokenRequest

### LINKED ISSUE #1218: Fetch service account token using TokenRequest
On the project `Members` screen it is possible to download the kubeconfig for a service account. Currently the dashboard reads the service account token secret. Instead, the [TokenRequest](https://kubernetes.io/docs/reference/kubernetes-api/authentication-resources/token-request-v1/) API should be used to acquire service account tokens.

- [] Dashboard should have a configuration option that is in sync with the `--service-account-max-token-expiration` gardener kube-apiserver flag
- [] Use TokenRequest API to acquire service account tokens
- [] The user should be able to change the token expiration (within the allowed boundaries)
- [] Remove the feature to rotate the service account secret 

**Why is this needed**:
https://github.com/gardener/dashboard/issues/1217

petersutter:

Fixes #1218

**Release note**:

```bugfix user

```


backend/lib/services/members/MemberManager.js
```diff
@@ -187,41 +174,24 @@ class MemberManager {
...
-      .map(({ name }) => this.client.core.secrets.delete(namespace, name))
-      .thru(promises => Promise.allSettled(promises))
-      .value()
-
-    this.constructor.getFulfilledValues(results)
-  }
-
   async getKubeconfig (item) {
-    const { namespace, name } = Member.parseUsername(item.id)
+    const defaultTokenExpiration = _.get(config, 'frontend.serviceAccountDefaultTokenExpiration', 7776000) // default is 90 days
```
holgerkoser, line 178: Why do we define the default in the code and not in the chart? Do we need a default in our deployment at all and let the this empty per default. This would mean we rely on max value of the apiserver.

holgerkoser (commented):
Only the default expiration is unclear to me.

grolu (approved):
/lgtm

PR #1236, opened 2022-06-08
Fixed seed selector logic

### LINKED ISSUE #906: Seed does not match CloudProfile when no label is present
The contract on `CloudProfile.SeedSelector` is currently written as "An empty list means that all seeds of the same provider type are supported". But [this code](https://github.com/gardener/dashboard/blob/1538be54ab96c76b32f8c5ad556a195554d24056/backend/lib/services/cloudprofiles.js#L42) seems to filter out all seeds that do no have a matching label. In other words, I believe the seeds must have a label match for this code to properly add the seed to the `CloudProfile`/ seed name array tuple that is returned.

In testing, `Seed` objects without labels are not properly mapped to `CloudProfile` objects that only have the matching types and [this log message](https://github.com/gardener/dashboard/blob/1538be54ab96c76b32f8c5ad556a195554d24056/backend/lib/services/cloudprofiles.js#L69) appears in the dashboard logs.

grolu:
This PR fixes the logic that selects eligible seeds for a cloud profile.

Fixes #906

The issue here was that with the current implementation, cloud profiles with an empty matchLabels list would map to any seed that has at least **one** label defined. Seeds with completely empty label list were not added to the list.
This is because

```js
_.filter(seeds, { metadata: { labels: {} } })
```

matches arbitrary label objects, however they must not be empty.

**Release note**:

```bugfix operator
Fixed a bug that caused seeds with empty label section not to be mapped to cloud profiles with empty `matchLabels` 
```


petersutter (approved):
/lgtm

PR #1238, opened 2022-06-14
Sort supported versions up within same patch group

### LINKED ISSUE #1034: In the update version dialog show firstly the supported version and then the others
When I tried to update a k8s version of a shoot cluster, I see this dialog 
image

It will be nice the supported version to be always on top and after that the deprecated versions to be shown.
For the preview versions, I have no opinion whether they should be before or after the supported version.

**Why is this needed**:
To achieve better user experience, as the supported version will be always the first choice.

grolu:
Sort supported versions up within same patch group, use colors to highlight supported / deprecated versions

Screen Shot 2022-06-28 at 15 55 06

Fixes #1034

**Release note**:

```feature user
Sort supported versions up within same patch group, use colors to highlight supported / deprecated versions
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 #### Builder ####
-FROM node:16-alpine3.15 as builder
+FROM node:18-alpine3.16 as builder
 
 WORKDIR /usr/src/app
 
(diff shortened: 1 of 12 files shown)


grolu:
Instead of using the success color for the supported version, we could also make it bold.

frontend/src/components/ShootVersion/ShootVersionUpdate.vue
```diff
@@ -93,14 +101,21 @@ export default {
...
           } else if (b.header) {
             return 1
-          } else {
-            if (semver.eq(a.version, b.version)) {
-              return 0
-            } else if (semver.gt(a.version, b.version)) {
-              return 1
-            } else {
+          }
+          if (semver.diff(a.version, b.version) === 'patch') {
```
holgerkoser, line 105: The versions are already grouped by update type `['patch', 'minor', 'major']`. I would propose to sort the versions for each group and afterwards add the header to the group. In this case all the entries would not need to have the type property I think.

holgerkoser: How do we handle `pre` release versions?

petersutter (approved):
/lgtm

holgerkoser (approved):
/lgtm

PR #1239, opened 2022-06-15
Show user error as dedicated icon

### LINKED ISSUE #1032: Use the icon for user issues also for hibernated clusters
When a hibernated cluster is experiencing user issue, it has a red hibernated icon.
To find that the issue is classified as user problem, it is necessary to hover over the status icon

image

It would be nice, if the user issue icon have higher priority and is shown even for hibernated clusters.
Optionally, both icons can be shown.

**Why is this needed**:
The gardener landscape operators can easily identify clusters with user issues.

grolu:
Show user error as badge so that the underlying error is still visible, see also linked issue.

Screen Shot 2022-07-04 at 12 49 32

I'm not sure about the layout though. It looks kind of unsteady. We should discuss this.

Fixes #1032

**Release note**:

```feature operator
Added dedicated icon for user errors in shoot status so that the underlying shoot status is still visible
```

### DIFF
--- a/frontend/src/components/ShootListRow.vue
+++ b/frontend/src/components/ShootListRow.vue
@@ -62,12 +62,12 @@ SPDX-License-Identifier: Apache-2.0
         </v-tooltip>
       </template>
       <template v-if="cell.header.value === 'purpose'">
-        <div class="d-flex justify-center pr-4">
+        <div class="d-flex justify-center">
           <purpose-tag :purpose="shootPurpose"></purpose-tag>
         </div>
       </template>
       <template v-if="cell.header.value === 'lastOperation'">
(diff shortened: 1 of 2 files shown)


holgerkoser:
We always have the same problem of wanting to represent many different things in one property or icon. Maybe we should make a separate column for user issue. And we should also have some kind of advanced search for clusters. I wonder if we shouldn't address a general solution to the problem.

grolu:
>Maybe we should make a separate column for user issue

So you propose to add an additional column. Or shall we just add an additional icon just as @vpnachev proposed here: https://github.com/gardener/dashboard/issues/1032

holgerkoser (approved):
/lgtm

PR #1247, opened 2022-06-30
terminal: use TokenRequest to get service account token

### LINKED ISSUE #1219: Fetch `attach` token using TokenRequest
This is a cross-repo issue, as also the g/terminal-controller-manager needs to be adapted.

Instead of reading the service account token secret, the TokenRequestAPI should be used to get the `attach` token for the Webterminal.

The g/terminal-controller-manager should create the kubeconfig that is used within the terminal pod on the host cluster by creating a TokenRequest instead of reading the `access` service account token secret.

**Why is this needed**:
https://github.com/gardener/dashboard/issues/1217

petersutter:
use TokenRequest to get service account token for the attach service account that is required to attach to the terminal pod.

Fixes #1219

**Release note**:

```other operator

```

### DIFF
--- a/backend/test/acceptance/__snapshots__/api.terminals.spec.js.snap
+++ b/backend/test/acceptance/__snapshots__/api.terminals.spec.js.snap
@@ -562,17 +562,16 @@ Array [
   Array [
     Object {
       ":authority": "api.host.garden.shoot.cluster",
-      ":method": "get",
-      ":path": "/api/v1/namespaces/term-host-1/serviceaccounts?watch=true&fieldSelector=metadata.name%3Dterm-attach-1",
+      ":method": "post",
+      ":path": "/api/v1/namespaces/term-host-1/serviceaccounts/term-attach-1/token",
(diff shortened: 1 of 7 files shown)


backend/test/acceptance/__snapshots__/api.terminals.spec.js.snap
```diff
@@ -562,17 +562,16 @@ Array [
...
-  Array [
     Object {
-      ":authority": "api.host.garden.shoot.cluster",
-      ":method": "get",
-      ":path": "/api/v1/namespaces/term-host-1/secrets/term-attach-1-token-74657",
-      ":scheme": "https",
+      "apiVersion": "authentication.k8s.io/v1",
+      "kind": "TokenRequest",
+      "spec": Object {
+        "expirationSeconds": 86400,
```
holgerkoser, line 573:
```suggestion
        "expirationSeconds": 43200,
```

holgerkoser (approved):
/lgtm

PR #1252, opened 2022-07-07
Show terminating and failed projects

### LINKED ISSUE #1045: Show Terminating and Failed Projects
Currently when a Project is deleted, it stays in Terminating until the BackupEntries inside are deleted. For more details see https://github.com/gardener/gardener/issues/1784.
Currently the dashboard does not show Projects that are in Terminating state. This creates confusion for end users as they delete a Project => the dashboard does no longer show it => the end user tries to create a Project with the same name => the creation fails with reason that Project with the same name already exists. The dashboard should make the Project lifecycle transparent to the user, it should not hide it instead.

**Why is this needed**:
To make the Project lifecycle transparent to the end users and to prevent any confusion about pending Project deletion.

grolu:
This PR removes the filter for projects with phase != Ready from the backend.
The dashboard now shows projects that are in terminating, initial or failed state.

Screen Shot 2022-07-07 at 11 33 52
Screen Shot 2022-07-07 at 08 35 23

Fixes #1045

- We need to discuss if we want to stop the user from working with unready projects
- I completely removed the phase filtering logic and qs param from the backend. I think we don't need it anymore, need to check
- I also fixed a bug regarding the project list width for long user names
- I renamed computed property `project` to `selectedProject` in `MainNavigation.vue` as it completely confused me 

**Release note**:

```feature user
Projects are now also visible if the phase is not `Ready`. In this case the project has an warning icon next to the name 
```


backend/lib/services/projects.js
```diff
@@ -148,22 +151,12 @@ exports.list = async function ({ user, qs = {} }) {
     return hasGroupMembership || hasUserMembership || hasServiceAccountMembership
   }
 
-  const phases = _
-    .chain(qs)
-    .get('phase', 'Ready')
-    .split(',')
-    .compact()
-    .value()
```
petersutter, line 156: > We need to discuss if we want to stop the user from working with unready projects

> I completely removed the phase filtering logic and qs param from the backend. I think we don't need it anymore, need to check

We can't be sure if the user can work with the project.

A project can be in failed state in case the namespace could not be claimed. In this case you can't work with the project https://github.com/gardener/gardener/blob/master/pkg/controllermanager/controller/project/project_control_reconcile.go#L63-L70. 
It can also be in a failed state when the RBAC resources could not be reconciled https://github.com/gardener/gardener/blob/master/pkg/controllermanager/controller/project/project_control_reconcile.go#L110-L114. E.g. you are assigned to the project as member with a specific role, but the rolebinding was not yet updated and thus you have no permissions or the other way around.

grolu: As discussed, I changed it to exclude only projects that are in phase `Pending`. Projects may turn from `Pending` to `Failed`, e.g. as you described. However, we agreed that users should still see those projects. Gardener should provide appropriate error messages if the user tries to perform an action that includes creating resources in the namespace, like creating a shoot or secret.

frontend/src/components/NotReadyProjectWarning.vue
```diff
@@ -0,0 +1,50 @@
...
+
+<template>
+  <v-tooltip top v-if="projectDetails.isNotReady">
+    <template v-slot:activator="{ on }">
+      <v-icon v-if="projectDetails.phase === 'Terminating'" :small="small" v-on="on" color="primary" class="ml-1">mdi-delete-sweep</v-icon>
+      <v-icon v-else-if="projectDetails.phase === 'Initial'" :small="small" v-on="on" color="primary" class="ml-1">mdi-plus-circle-outline</v-icon>
+      <v-icon v-else :small="small" v-on="on" color="warning" class="ml-1">mdi-alert-circle-outline</v-icon>
+    </template>
+    <div>
+      The project phase is <v-chip color="primary" label x-small>{{projectDetails.phase}}</v-chip>
```
petersutter, line 15:
```suggestion
      The project phase is <v-chip color="primary" label x-small class="px-1">{{projectDetails.phase}}</v-chip>
```

old:


Screenshot 2022-07-26 at 19 22 28

new:
Screenshot 2022-07-26 at 19 22 14

and maybe we can highlight / make the header bold?

frontend/src/components/NotReadyProjectWarning.vue
```diff
@@ -0,0 +1,50 @@
+<!--
+SPDX-FileCopyrightText: 2021 SAP SE or an SAP affiliate company and Gardener contributors
+
+SPDX-License-Identifier: Apache-2.0
+-->
+
+<template>
+  <v-tooltip top v-if="projectDetails.isNotReady">
```
holgerkoser, line 8: Why do we have a property `isNotReady` but not `isTerminating` and not `isInitial`. Shouldn't we also use `projectDetails.phase !== 'Ready'` for this case?

frontend/src/utils/index.js
```diff
@@ -311,7 +314,9 @@ export function getProjectDetails (project) {
     description,
     purpose,
     staleSinceTimestamp,
-    staleAutoDeleteTimestamp
+    staleAutoDeleteTimestamp,
+    phase,
+    isNotReady
```
holgerkoser, line 319: If we need this boolean property I would use a getter in this case.
```suggestion
    get isNotReady () { return this.phase !== 'Ready' }
```

frontend/src/store/modules/projects.js
```diff
@@ -47,9 +47,9 @@ const actions = {
   },
   async delete ({ commit, state }, { metadata }) {
     const namespace = metadata.namespace
-    await deleteProject({ namespace })
-    commit('ITEM_DELETED', metadata)
-    return state.all
+    const res = await deleteProject({ namespace })
+    // commit('ITEM_DELETED', metadata)
```
holgerkoser, line 51: Why is this line commented out. If we don't need it anymore it should be removed. Do we really need the return value of this action?

backend/lib/services/projects.js
```diff
@@ -37,9 +37,12 @@ function fromResource ({ metadata, spec = {}, status = {} }) {
       createdBy: fromSubject(createdBy),
       owner: fromSubject(owner),
       description,
-      purpose,
+      purpose
+    },
+    status: {
```
holgerkoser, line 42: The idea behind the mapping of resources has been, to only have `data` and `metadata`. If we want to move back to the resource structure we should not do this silently but in one PR for the hole project.

backend/lib/services/projects.js
```diff
@@ -148,23 +151,14 @@ exports.list = async function ({ user, qs = {} }) {
...
     .filter(project => {
       if (!isAdmin && !isMemberOf(project)) {
         return false
       }
-      if (!_.isEmpty(phases)) {
-        const phase = _.get(project, 'status.phase', 'Initial')
-        return _.includes(phases, phase)
-      }
-      return true
+      const phase = _.get(project, 'status.phase', 'Pending')
```
holgerkoser, line 160: This needs to be aligned with the #1246.

holgerkoser: If we introduce a reusable function `filterProjects` maybe in `utils/index.js` we could reuse it here and in my PR.
```js
function filterProjects (projects, user, isAdmin = false) {
  const isMemberOf = project => {
    const hasGroupMembership = _
      .chain(project)
      .get('spec.members')
      .filter(['kind', 'Group'])
      .map('name')
      .intersection(user.groups)
      .size()
      .gt(0)
      .value()

    const hasUserMembership = _
      .chain(project)
      .get('spec.members')
      .filter(['kind', 'User'])
      .map('name')
      .includes(user.id)
      .value()

    const hasServiceAccountMembership = _
      .chain(project)
      .get('spec.members')
      .filter(['kind', 'ServiceAccount'])
      .map(({ namespace, name }) => `system:serviceaccount:${namespace}:${name}`)
      .includes(user.id)
      .value()

    return hasGroupMembership || hasUserMembership || hasServiceAccountMembership
  }

  return _
    .chain(projects)
    .filter(project => {
      if (!isAdmin && !isMemberOf(project)) {
        return false
      }
      const phase = _.get(project, 'status.phase', 'Pending')
      return phase !== 'Pending'
    })
    .value()
}
```

This function would look like:
```js
exports.list = async function ({ user }) {
  const isAdmin = await authorization.isAdmin(user)
  const projects = filterProjects(cache.getProjects(), user, isAdmin)
  return _.map(projects, fromResource)
}
```

frontend/src/components/NotReadyProjectWarning.vue
```diff
@@ -0,0 +1,50 @@
...
+      <span v-else>The project is in an unready state. The project might not be functional.</span>
+    </div>
+  </v-tooltip>
+</template>
+
+<script>
+import { getProjectDetails } from '@/utils'
+
+export default {
+  name: 'notReadyProjectWarning',
```
holgerkoser, line 29: See https://eslint.vuejs.org/rules/multi-word-component-names.html
```suggestion
  name: 'not-ready-project-warning',
```

frontend/src/components/NotReadyProjectWarning.vue
```diff
@@ -0,0 +1,50 @@
...
+SPDX-FileCopyrightText: 2021 SAP SE or an SAP affiliate company and Gardener contributors
+
+SPDX-License-Identifier: Apache-2.0
+-->
+
+<template>
+  <v-tooltip top v-if="projectDetails.isNotReady">
+    <template v-slot:activator="{ on }">
+      <v-icon v-if="projectDetails.phase === 'Terminating'" :small="small" v-on="on" color="primary" class="ml-1">mdi-delete-sweep</v-icon>
+      <v-icon v-else-if="projectDetails.phase === 'Initial'" :small="small" v-on="on" color="primary" class="ml-1">mdi-plus-circle-outline</v-icon>
```
holgerkoser, line 11: Pending

holgerkoser, line 11:
```suggestion
      <v-icon v-else-if="projectDetails.phase === 'Pending'" :small="small" v-on="on" color="primary" class="ml-1">mdi-plus-circle-outline</v-icon>
```

frontend/src/store/modules/projects.js
holgerkoser: write a comment why we do not delete from store.

holgerkoser: delete mutation

frontend/src/store/modules/projects.js
```diff
@@ -47,9 +47,9 @@ const actions = {
...
+    return res.data
```
holgerkoser, line 52: remove return value

frontend/src/components/NotReadyProjectWarning.vue
holgerkoser: The lower smaller text

backend/lib/services/projects.js
grolu: @holgerkoser will do in his PR

petersutter (approved):
/lgtm

holgerkoser (approved):
/lgtm

backend/lib/services/projects.js
holgerkoser: done https://github.com/gardener/dashboard/pull/1246/commits/7eecd1653eebce9a29269dec5da54b9e994dc288

PR #1257, opened 2022-07-16
Disable experimental ESM loader

holgerkoser:
With node version `18.6.0` the experimental ESM loader of yarn always throws an exception because it does not return the required property `shortCircuit: true` (see https://nodejs.org/api/esm.html#hooks).    

See issue https://github.com/yarnpkg/berry/issues/4629

Fixes #

**Release note**:

```bugfix user

```


petersutter (approved):
/lgtm

PR #1235, opened 2022-06-07
Focus Cluster List

grolu:
**What this PR does / why we need it**:
Allow operators to freeze the all all projects cluster list
- Clusters no longer change position, sorting is fixed
- Items still get updates
- Gone clusters are kept as stale objects in the list
- New clusters are not added but a badge next to switch informs about new clusters

Screen Shot 2022-07-12 at 13 54 57

**Which issue(s) this PR fixes**:
Fixes #1121

**Special notes for your reviewer**:
Build currently fails due to coverage not met

**Release note**:

```feature operator
New focus mode for the `all projects` cluster list: This feature enables operators to get a static overview of clusters with issues. When the focus mode is active, the cluster list behavior changes:
- Clusters no longer change position, the sorting is fixed even if data changes
- Gone clusters are kept as stale objects in the list to keep the list steady
- New clusters are not added but a badge next to the focus switch informs about new clusters
```

PR #1258, opened 2022-07-19
Table menu should be scrollable

petersutter:
On smaller window sizes the table menu (e.g. on the ALL PROJECT page) could get so large, that you can't get to the last items in the menu, as it is not scrollable.
With this PR I set a max-height of 80% for the menu which will make the menu scrollable in case it does not fit on the screen.

Fixes #

**Release note**:

```bugfix user
The table menu is now scrollable in case the content does not fit on the window
```

### DIFF
--- a/frontend/src/components/TableColumnSelection.vue
+++ b/frontend/src/components/TableColumnSelection.vue
@@ -5,7 +5,7 @@ SPDX-License-Identifier: Apache-2.0
 -->
 
 <template>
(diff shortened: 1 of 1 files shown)


grolu:
Actually, I also fixed this in my PR here https://github.com/gardener/dashboard/pull/1235 :)

grolu (approved):
/lgtm
I will merge this into my PR

PR #1263, opened 2022-07-28
Hide oidc login if not configured

### LINKED ISSUE #1233: OIDC visible on Login screen if not configured
**What happened**:
Screenshot 2022-06-03 at 09 28 34
OIDC is still visible if there is no oidc configuration.

Only `Token` login should be visible.

Remove the oidc configuration from the dashboard and open the login screen.

related PR https://github.com/gardener/dashboard/pull/957 where the oidc setting was changed to not be required anymore.

**Environment**:

petersutter:
If oidc is not configured only token login should be available

Fixes #1233 

**Release note**:

```bugfix operator
Fixed an issue where the OIDC login was not hidden on the login screen if it was not configured by the gardener dashboard administrator
```

### DIFF
--- a/backend/lib/config/gardener.js
+++ b/backend/lib/config/gardener.js
@@ -110,7 +110,7 @@ module.exports = {
       assert.ok(_.get(config, path), `Configuration value '${path}' is required`)
     })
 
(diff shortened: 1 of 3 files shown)


holgerkoser (approved):
/lgtm

PR #1267, opened 2022-07-28
Split config into public and authenticated endpoint

petersutter:
For the `Login` page the `loginTypes` and `landingPageUrl` properties will be moved to the public `/login-config.json` endpoint. All other configuration that is not required for unauthenticated access is moved under `/api/config`

Fixes #

**Release note**:

```other developer
The frontend config is now split into an authenticated (`/api/config`) and unauthenticated endpoint (`/login-config.json`) 
```


backend/test/acceptance/__snapshots__/config.spec.js.snap
```diff
@@ -1,5 +1,30 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`config should reject unauthenticated requests 1`] = `Array []`;
+
+exports[`config should reject unauthenticated requests 2`] = `
+Object {
+  "code": 401,
+  "details": Object {
+    "name": "UnauthorizedError",
+    "stack": "UnauthorizedError: No authorization token was found
```
holgerkoser, line 10: A stack trace in a snap is not a good idea

backend/test/acceptance/config.spec.js
```diff
@@ -36,4 +50,15 @@ describe('config', function () {
 
     expect(res.body).toMatchSnapshot()
   })
+
+  it('should reject unauthenticated requests', async function () {
```
holgerkoser, line 54: We do not need this test. The middleware of `/api` is already tested in an own acceptance test. Test only the good case.

frontend/src/router/guards.js
```diff
@@ -6,13 +6,17 @@
...
+  getConfiguration,
+  getPublicConfiguration
+} from '@/utils/api'
 
 export default function createGuards (store, userManager, localStorage) {
   return {
     beforeEach: [
       setLoading(store, true),
-      ensureConfigurationLoaded(store),
+      ensureConfigurationLoaded(store, userManager),
```
holgerkoser, line 19: We can move this down if we have a static asset for public config

charts/gardener-dashboard/values.yaml
```diff
@@ -101,8 +101,11 @@ oidc:
   #  # clientSecret is the public client secret use by kubelogin and all users
   #  clientSecret: ~
 
-frontendConfig:
+# frontendConfigPublic is the configuration that is accessible without authentication
+frontendConfigPublic:
```
holgerkoser, line 105: I would keep the `landingPageUrl` in `frontendConfig` and handle what is public in the tmeplate for now.

backend/test/acceptance/config.spec.js
petersutter: ✅

charts/gardener-dashboard/values.yaml
petersutter: ✅

backend/test/acceptance/__snapshots__/config.spec.js.snap
petersutter: ✅

frontend/src/router/guards.js
petersutter: ✅

backend/lib/app.js
```diff
@@ -57,12 +57,11 @@ app.use(helmet.hsts())
 app.use(noCache(['/js', '/css', '/fonts', '/img', '/static']))
 app.use('/auth', auth.router)
 app.use('/webhook', githubWebhook.router)
-app.use(['/api', '/config.json'], compression({
+app.use(['/api'], compression({
```
holgerkoser, line 60:
```suggestion
app.use('/api', compression({
```

frontend/src/router/guards.js
```diff
@@ -36,10 +36,14 @@ function setLoading (store, value) {
 
 function ensureConfigurationLoaded (store) {
   return async function (to, from, next) {
+    const meta = to.meta || {}
```
holgerkoser, line 39: If we move `ensureConfigurationLoaded` inside of `ensureDateLoaded` https://github.com/gardener/dashboard/blob/42fca41d290705e4bbe608376166f3bb743c34ae/frontend/src/router/guards.js#L101 we don't need this. And I think now we can consider config as data.

frontend/src/views/Login.vue
```diff
@@ -122,6 +126,10 @@ export default {
     })
   },
   methods: {
+    async getLoginConfiguration () {
+      const { data: cfg } = await getLoginConfiguration()
+      assign(this.cfg, cfg)
```
holgerkoser, line 131:
```suggestion
      Object.assign(this.cfg, cfg)
```

frontend/src/views/Login.vue
```diff
@@ -164,7 +172,8 @@ export default {
       this.$snotify.error(message, 'Login Error', config)
     }
   },
-  mounted () {
+  async mounted () {
```
holgerkoser, line 175: Why not in created?

frontend/src/router/guards.js
petersutter: ✅

frontend/src/views/Login.vue
petersutter: ✅

holgerkoser (approved):
/lgtm

grolu (approved):
/lgtm

PR #1271, opened 2022-08-04
Filter intentionally broken clusters

### LINKED ISSUE #1269: Filter intentionally broken clusters
It would be helpful to be able to annotate a shoot with e.g. `dashboard.gardener.cloud/ignore-issues` and the dashboard subsequently filtering away such clusters in the "All Projects" view (either when “Show only clusters with issues” is checked or a new filter item for this purpose would be added)

**Why is this needed**:
Customers regularly intentionally break clusters (e.g. because of chaos-testing, because of testing some knowingly broken machine images, ...). At the same time, operators proactively check the landscapes for clusters having issues (as shown in the "All Projects" view with filter “Show only clusters with issues” enabled), hence operators unnecessarily bother with intentionally broken clusters.

petersutter:
Clusters that have the `dashboard.gardener.cloud/ignore-issues` annotation are filtered on the `All Projects` page in case the `Hide no operator action required issues` filter is checked

Fixes #1269

**Release note**:

```feature operator
Clusters that have the `dashboard.gardener.cloud/ignore-issues` annotation are filtered on the `All Projects` page in case the `Hide no operator action required issues` filter is checked
```

### DIFF
--- a/frontend/src/store/modules/shoots/index.js
+++ b/frontend/src/store/modules/shoots/index.js
@@ -290,11 +290,12 @@ function setFilteredItems (state, rootState, rootGetters) {
         }
         const lastErrors = get(item, 'status.lastErrors', [])
         const allLastErrorCodes = errorCodesFromArray(lastErrors)
+        if (isTemporaryError(allLastErrorCodes)) {
+          return false
+        }
         const conditions = get(item, 'status.conditions', [])
         const allConditionCodes = errorCodesFromArray(conditions)
-        const noUserError = !isUserError(allLastErrorCodes) && !isUserError(allConditionCodes)
-        const noTemporaryError = !isTemporaryError(allLastErrorCodes)
-        return noUserError && noTemporaryError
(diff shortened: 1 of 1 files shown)


grolu:
Do we need to provide an easy way to set this flag (via UI) or is it sufficient that we just filter the issues @dguendisch
I think that we should at least show somewhere that the cluster is flagged as `ignore-issues` so that it is clear why a cluster is filtered WDYT?

petersutter:
unless there is a clear idea of where/which places and how to show this information I would see this as a separate task. This PR was just a quick fix for the issue. Improving the usability can be done later on top

dguendisch:
On the long run it makes sense to flag such clusters somehow visually, however I'd think that would be of lower prio (main prio is to get our operators rid of proactively checking such clusters at all), hence I'd agree to Peter to treat flagging as a separate task.

grolu (approved):
/lgtm

holgerkoser (approved):
/lgtm

Issue #1282, opened 2022-08-24
dashboard chart: Add support for projected service account tokens


petersutter:
**What would you like to be added**:
The dashboard meanwhile supports the tokenFile within the clientConfig ([ref](https://github.com/gardener/dashboard/pull/1232)). 
The chart should also be adapted to support projected service account tokens ([ref](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#service-account-token-volume-projection)).
- The ability to configure the dashboard to use service account token volume projection
- The ability to configure a user instead of a service account subject in the clusterrolebinding definition when using a "virtual garden" setup. This will enable other possibilities for authentication to the virtual garden, i.e., leveraging [oidc-webhook-authenticator](https://github.com/gardener/oidc-webhook-authenticator).

depends-on: https://github.com/gardener/dashboard/issues/1281

**Why is this needed**:
To make use of short lived tokens that the kubelet projects into the pod. The service account token will also become invalid against the API when the Pod or the ServiceAccount is deleted.

petersutter:
/close 
resolved with https://github.com/gardener/dashboard/pull/1283

Issue #545, opened 2020-01-14
Restrict permissions of dashboard's service user to garden cluster


petersutter:
Currently, the dashboard has [cluster-admin](https://github.com/gardener/dashboard/blob/b1c1afd5ca6d3486d78e315eaa664235f4b73c30/charts/gardener-dashboard/templates/rbac.yaml#L11-L14) privileges which is not necessary. 

We should create a new dashboard role that only contains the necessary privileges.

petersutter:
These should be the required privileges for the dashboard. We probably have to split the dashboard chart into application and runtime chart, similar like with the gardener charts

```yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: gardener-dashboard-role
rules:
- apiGroups:
  - authentication.k8s.io
  resources:
  - tokenreviews
  verbs:
  - create
- apiGroups:
  - core.gardener.cloud
  resources:
  - quotas
  verbs:
  - list
  - watch
- apiGroups:
  - core.gardener.cloud
  resources:
  - projects
  verbs:
  - list
  - watch
- apiGroups:
  - core.gardener.cloud
  resources:
  - shoots
  verbs:
  - watch
- apiGroups:
  - apiregistration.k8s.io
  resources:
  - apiservices
  verbs:
  - get
# terminal bootstrapping
- apiGroups:
  - ""
  resources:
  - secrets
  verbs:
  - get
- apiGroups:
  - core.gardener.cloud
  resources:
  - shoots
  verbs:
  - get
```

petersutter:
this issue is blocked until gardener/monitoring#11 is resolved

petersutter:
> this issue is blocked until [gardener/monitoring#11](https://github.com/gardener/monitoring/issues/11) is resolved

this is now tracked in https://github.com/gardener/dashboard/issues/1338

The permissions of the dashboard user are restricted with https://github.com/gardener/dashboard/pull/1283
/close

PR #1283, opened 2022-08-24
Chart improvements: support virtual-garden setup, restrict permissions, support projected service account tokens

### LINKED ISSUE #1281: dashboard chart: split into application and runtime parts
The dashboard chart in its current form does not conveniently support a "virtual garden" setup and assumes a single cluster deployment.

**Why is this needed**:
The chart should be split into an application chart (for the virtual garden) and a runtime chart for the runtime cluster. We can then adapt the gardener-dashboard-role accordingly https://github.com/gardener/dashboard/issues/545. Currently, it's not possible to deploy the gardener-dashboard-role using the chart to the virtual garden.

petersutter:
- Split Dashboard Helm chart into application and runtime resources
- Restrict permissions of dashboard user
- The chart was adapted to support projected service account tokens ([ref](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#service-account-token-volume-projection)).

It is now also possible to configure a user (`.Values.global.virtualGarden.userName`) instead of a service account subject in the clusterrolebinding definition when using a "virtual garden" setup. This will enable other possibilities for authentication to the virtual garden, i.e., leveraging [oidc-webhook-authenticator](https://github.com/gardener/oidc-webhook-authenticator).

Fixes #1281, #545, #1282

To easier spot what has changed, enable `Hide whitespace` checkbox on the `Files changed` view

**Release note**:

```breaking operator
The `gardener-dashboard` Helm chart has been split into two different subcharts (`runtime` and `application`). This change required that all chart values that previously were on top-level have now moved under the `.global` key. However, the structure/types of all existing keys has not changed. Please [check the `values.yaml`](https://github.com/gardener/dashboard/blob/master/charts/gardener-dashboard/values.yaml).
```

```feature operator
The `gardener-dashboard` Helm chart now supports configuration for enabling service account token volume projection. It is exposed through the `.Values.global.serviceAccountTokenVolumeProjection` section in the `values.yaml` file.
```

```feature operator
It is now possible to configure a `User` instead of a `ServiceAccount` subject in the `ClusterRoleBinding` for the Gardener Dashboard when using virtual garden setup by setting `.Values.global.virtualGarden.userName` in the `values.yaml` file.
```

```other operator
The `gardener-dashboard` user is no longer bound to the `cluster-admin` `ClusterRole`. Instead, it is bound to the newly introduced `gardener.cloud:system:dashboard` `ClusterRole`, which grants only those permissions that are required for the dashboard user.
```


charts/__tests__/gardener-dashboard/clusterrolebinding.spec.js
```diff
@@ -9,7 +9,7 @@
 const { basename } = require('path')
 const { helm } = fixtures
 
-const chart = basename(__dirname)
+const chart = basename(__dirname) + '/charts/application' // TODO discuss
```
holgerkoser, line 12: I would propose to change the function `renderChartTemplates` in `__fixtures__/helm.js` like this:
```js
function renderTemplatesFn (...paths) {
  const chart = paths[0]
  const cwd = path.resolve(__dirname, '..', ...paths)
  let name, defaults
  try {
    { defaults = {}, releaseName } = require('./' + chart)
    { name = releaseName } = yaml.load(fs.readFileSync(path.join(cwd, 'Chart.yaml'), 'utf8'))    
  } catch (err) {
    assert.fail(`Invalid chart "${chart}"`)
  }
  return async (templates, values) => {
    defaultsDeep(values, defaults)
    ...
  }
}
```
and use it like this:
```js
const renderTemplates = helm.renderTemplateFn(basename(__dirname), 'charts', 'application')
```

charts/gardener-dashboard/charts/application/Chart.yaml
```diff
@@ -0,0 +1,8 @@
+# SPDX-FileCopyrightText: 2022 SAP SE or an SAP affiliate company and Gardener contributors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+apiVersion: v1
+description: A Helm chart to deploy the Gardener dashboard application related components
+name: application
+version: 0.2.0
```
holgerkoser, line 8: We should change the version to `0.1.0` or `0.3.0`

charts/__fixtures__/gardener-dashboard/charts/application.js
```diff
@@ -0,0 +1,16 @@
+//
```
holgerkoser, line 1: I think we don't need these files if we change the renderTemplates function.

charts/gardener-dashboard/charts/runtime/templates/configmap.yaml
```diff
@@ -0,0 +1,341 @@
...
+kind: ConfigMap
+metadata:
+  name: gardener-dashboard-configmap
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app: gardener-dashboard
+    chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+    release: "{{ .Release.Name }}"
+    heritage: "{{ .Release.Service }}"
+data:
```
holgerkoser, line 15: We could wrap it with 
```
{{- with .Values.global }}
...
{{- end }}
```
to not have this very long pathnames

charts/__tests__/gardener-dashboard/clusterrolebinding.spec.js
holgerkoser, line 12: Do something like this:
```js
function renderTemplatesFn (...paths) {
  const cwd = path.resolve(__dirname, '..', ...paths)
  let name, defaults
  try {
    defaults = require('./' + paths[0]).defaults
    const data = fs.readFileSync(path.join(cwd, 'Chart.yaml'), 'utf8')
    name = yaml.loadAll(data).name
  } catch (err) {
    assert.fail(`Invalid chart "${path.join(paths)}"`)
  }
  return async (templates, values) => {
    defaultsDeep(values, defaults)
    const dirname = process.env.HELM_VALUES_DIRNAME
    assert.ok(fs.statSync(dirname).isDirectory(), `Invalid helm values directory "${dirname}"`)
    const filename = path.join(dirname, `values-${randomNumber()}.yaml`)
    try {
      fs.writeFileSync(filename, yaml.dump({ global: values }, { skipInvalid: true }))
      const cmd = [
        'helm',
        'template',
        name,
        '-n',
        'garden',
        ...flatMap(templates, template => ['-s', `templates/${template}.yaml`]),
        '.',
        '-f',
        filename
      ]
      const { stdout } = await exec(cmd.join(' '), { cwd })
      return yaml.loadAll(stdout)
    } finally {
      fs.unlinkSync(filename)
    }
  }
}
```
and use it like this:
```js
const renderTemplates = helm.renderTemplatesFn(basename(__dirname), 'charts', 'application')
```

charts/gardener-dashboard/charts/runtime/templates/configmap.yaml
petersutter: we agreed that, for better search-ability of the properties, we do not simplify here even though it would improve the readability

charts/gardener-dashboard/charts/application/Chart.yaml
petersutter: Changed to `0.1.0` for all charts as it does not affect the helm chart installation

holgerkoser (approved):
/lgtm

grolu (approved):
/lgtm

PR #1286, opened 2022-08-30
Chart: Support for projected service account tokens

### LINKED ISSUE #1282: dashboard chart: Add support for projected service account tokens

petersutter:
The chart was adapted to support projected service account tokens ([ref](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#service-account-token-volume-projection)).

Fixes #1282 

**Release note**:

```bugfix user

```

### DIFF
--- a/charts/__tests__/gardener-dashboard/__snapshots__/deployment.spec.js.snap
+++ b/charts/__tests__/gardener-dashboard/__snapshots__/deployment.spec.js.snap
@@ -121,6 +121,11 @@ Object {
                 "name": "gardener-dashboard-login-config",
                 "subPath": "login-config.json",
               },
+              Object {
+                "mountPath": "/var/run/secrets/projected/serviceaccount",
+                "name": "service-account-token",
+                "readOnly": true,
+              },
             ],
           },
         ],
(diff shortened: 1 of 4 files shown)


petersutter:
cherry picked commits into https://github.com/gardener/dashboard/pull/1283

PR #1291, opened 2022-09-13
Read controllerregistration version from annotation

### LINKED ISSUE #1033: About Dialogue doesn't show Extension version with new ControllerDeployment resource
**What happened**:
About Dialogue doesn't show Extension version when using new ControllerDeployment resource.
Version next to Extension name displayed.
Deploy any extension using new ContorllerDeployment resource.

**Environment**:

grolu:
The Dashboard currently does not show the version of deployed extensions.
With this PR we adapted the dashboard to look for an annotation `dashboard.gardener.cloud/version` in the `controllerregistration` resource. This annotation has to be set in the landscape deployment process.

Background: We used to read the version from the `controllerregistration` spec using the image tag. However since the extension deployment process changed this is no longer possible. We discussed with @rfranzke that there is currently no feasible way to automatically determine the version with the information accessible by the dashboard. So the only solution is to explicitly set the version in the `controllerregistration` resource itself.

Fixes #1033

**Release note**:

```feature operator
Extension version can be specified for each `controllerregistration` resource using the `dashboard.gardener.cloud/version` annotation. The dashboard will show the defined version in the about dialog
```

### DIFF
--- a/backend/__fixtures__/controllerregistrations.js
+++ b/backend/__fixtures__/controllerregistrations.js
@@ -6,7 +6,7 @@
 
 'use strict'
 
-const { cloneDeep, find } = require('lodash')
+const { cloneDeep, find, set } = require('lodash')
 
 function getControllerRegistration ({ uid, name, version, resources }) {
   const metadata = {
@@ -18,15 +18,7 @@ function getControllerRegistration ({ uid, name, version, resources }) {
     spec.resources = resources
   }
   if (version) {
-    spec.deployment = {
-      providerConfig: {
-        values: {
-          image: {
-            tag: version
-          }
-        }
-      }
-    }
+    set(metadata, ['annotations', 'dashboard.gardener.cloud/version'], version)
   }
   return { metadata, spec }
 }
--- a/backend/lib/services/controllerregistrations.js
+++ b/backend/lib/services/controllerregistrations.js
@@ -18,7 +18,8 @@ exports.listExtensions = async function ({ user }) {
   for (const { metadata, spec = {} } of controllerregistrations) {
     const name = metadata.name
     if (allowed) {
-      const version = _.get(spec, 'deployment.providerConfig.values.image.tag')
+      const annotations = _.get(metadata, 'annotations', {})
+      const version = _.get(annotations, 'dashboard.gardener.cloud/version')
       const resources = spec.resources
       extensions.push({ name, version, resources })
     } else {


backend/lib/services/controllerregistrations.js
```diff
@@ -18,7 +18,8 @@ exports.listExtensions = async function ({ user }) {
   for (const { metadata, spec = {} } of controllerregistrations) {
     const name = metadata.name
     if (allowed) {
-      const version = _.get(spec, 'deployment.providerConfig.values.image.tag')
+      const annotations = _.get(metadata, 'annotations', {})
+      const version = _.get(annotations, 'dashboard.gardener.cloud/version')
```
petersutter, line 22:
```suggestion
      const version = _.get(metadata, 'annotations["dashboard.gardener.cloud/version"]')
```

petersutter (approved):
/lgtm

however there should be a breaking change release note or the current release note has to be adapted to include the breaking change information

petersutter:
We could also think to use the recommended label for versions as suggested here https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/

backend/__fixtures__/controllerregistrations.js
```diff
@@ -18,15 +18,7 @@ function getControllerRegistration ({ uid, name, version, resources }) {
...
-    spec.deployment = {
-      providerConfig: {
-        values: {
-          image: {
-            tag: version
-          }
-        }
-      }
-    }
+    set(metadata, ['labels', 'app.kubernetes.io/version'], version)
```
holgerkoser, line 21:
```suggestion
    metadata.labels = { 'app.kubernetes.io/version': version }
```

backend/lib/services/controllerregistrations.js
```diff
@@ -18,7 +18,7 @@ exports.listExtensions = async function ({ user }) {
   for (const { metadata, spec = {} } of controllerregistrations) {
     const name = metadata.name
     if (allowed) {
-      const version = _.get(spec, 'deployment.providerConfig.values.image.tag')
+      const version = _.get(metadata, ['labels', 'app.kubernetes.io/version'])
```
holgerkoser, line 21:
```suggestion
      const version = metadata?.labels['app.kubernetes.io/version']
```

backend/__fixtures__/controllerregistrations.js
```diff
@@ -6,7 +6,7 @@
 
 'use strict'
 
-const { cloneDeep, find } = require('lodash')
+const { cloneDeep, find, set } = require('lodash')
```
petersutter, line 9:
```suggestion
const { cloneDeep, find } = require('lodash')
```

holgerkoser (approved):
/lgtm

PR #1300, opened 2022-09-27
Show gardenctl target cluster for non-admins

### LINKED ISSUE #1298: Add the `Target Cluster` to non-admin users
As a Gardener operator, I frequently use the `target Cluster` feature of the dashboard since it's very quick and useful. However, while debugging with a customer today, I found out that this option is not available for non-admin users.

So, as a Gardener non-admin user, I want to have the `Target Cluster` feature available so that I can quickly and easily target my shoot cluster via gardenctl-v2

**Why is this needed**:
Would be very useful, mostly now that customers can disable the static token kubeconfig for their cluster. However, I would leave the `Target Control Plane` command to admin users only.

petersutter:
The gardenctl `Target Cluster` command is shown for users with `create` `shoots/adminkubeconfig` permission (e.g. for users having the `admin` project role). Previously this was only shown for gardener admins.

Fixes #1298

**Release note**:

```feature user
The gardenctl `Target Cluster` command is now shown for users with `create` `shoots/adminkubeconfig` permission (e.g. for users having the `admin` project role). Previously this was only shown for gardener admins.
```

### DIFF
--- a/frontend/src/components/ShootDetails/GardenctlCommands.vue
+++ b/frontend/src/components/ShootDetails/GardenctlCommands.vue
@@ -96,6 +96,7 @@ export default {
       'cfg'
     ]),
     ...mapGetters([
+      'isAdmin',
       'projectFromProjectList',
       'gardenctlOptions'
     ]),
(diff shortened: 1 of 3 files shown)


grolu (approved):
/lgtm

holgerkoser (approved):
/lgtm

PR #1302, opened 2022-09-30
Create orphaned service account on member update

petersutter:
Create orphaned service account on member update

Screenshot 2022-09-30 at 15 54 42

Screenshot 2022-10-03 at 11 45 13

Fixes #

**Release note**:

```other user
When updating a service account project member the service account will be created in case the service account is listed as project member but does not actually exist
```

### DIFF
--- a/frontend/src/components/dialogs/MemberDialog.vue
+++ b/frontend/src/components/dialogs/MemberDialog.vue
@@ -65,6 +65,15 @@ SPDX-License-Identifier: Apache-2.0
               ></v-text-field>
             </v-col>
           </v-row>
+          <v-alert
+            v-if="isUpdateDialog && orphaned"
+            :value="true"
+            type="info"
+            color="primary"
+            outlined
+          >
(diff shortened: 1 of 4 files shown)


frontend/src/components/dialogs/MemberDialog.vue
```diff
@@ -65,6 +65,15 @@ SPDX-License-Identifier: Apache-2.0
...
             </v-col>
           </v-row>
+          <v-alert
+            v-if="isUpdateDialog && orphaned"
+            :value="true"
+            type="info"
+            color="primary"
+            outlined
+          >
+            The service account will be created in case it does not exist
```
grolu, line 75: This sounds like we do not know if the sa exits. maybe something like
`The service account does not exist anymore and will be re-created if you update the roles.`
or just `The service account will be re-created`

grolu (approved):
/lgtm

holgerkoser (approved):
/lgtm

PR #1307, opened 2022-10-05
Disable delete button for deleted service account

petersutter:
With this PR the delete button is disabled for service accounts that were deleted (deletionTimestamp is set). There might be finalizers on the service account preventing the final deletion of the object. In this case we disable the delete button.

Screenshot 2022-10-05 at 10 54 16

Fixes #

**Release note**:

```other user
Members page: The delete button for service accounts is now disabled in case the service account was already marked for deletion (`deletionTimestamp` is set)
```

### DIFF
--- a/backend/lib/services/members/Member.js
+++ b/backend/lib/services/members/Member.js
@@ -8,7 +8,7 @@
 
 const { pick } = require('lodash')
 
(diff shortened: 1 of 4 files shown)


holgerkoser (approved):
/lgtm

PR #1320, opened 2022-10-17
Adapted DNS Provider List getters to work with latest Gardener Releases

### LINKED ISSUE #1305: Missing DNS provider when adding DNS secret in Gardener version > 1.54.0
**What happened**:
There is no DNS providers available when we want to add new DNS Secret into the project.

OpenStack designate should appear after click on + icon in DNS secret part

Install gardener version > v1.54 without any DNSProvider registration kind

**Environment**:
Gardener 1.54.1

If I understand it correctly, dashboard looks for DNSProvider records in controllerregistrations. But DNSProvider has been removed in gardener since v1.54.0 (https://github.com/gardener/gardener/releases/tag/v1.54.0). Is there a plan to implement the detection of DNS providers in different way? Or is there other way how to solve missing  DNSProvider in the current version (v1.61.0) of dashboard?

grolu:
This PR adds compatibility with the latest gardener release.
As https://github.com/gardener/gardener/issues/5270 moved the DNSProvider capabilities out of g/g, we need a different approach to fetch the DNS Provider List

Fixes #1305

**Release note**:

```bugfix user
NONE
```

### DIFF
--- a/backend/__fixtures__/controllerregistrations.js
+++ b/backend/__fixtures__/controllerregistrations.js
@@ -55,10 +55,11 @@ const controllerRegistrationList = [
   }),
   getControllerRegistration({
     uid: 2,
-    name: 'DNS Registration',
+    name: 'Provider-Foo',
     resources: [{
-      kind: 'DNSProvider',
-      type: 'gardenland'
+      kind: 'DNSRecord',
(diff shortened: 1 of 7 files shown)


petersutter (approved):
/lgtm

PR #1326, opened 2022-10-21
Forward request options correctly to the session pool

### LINKED ISSUE #1325: http/2 session heartbeat not running
**What happened**: 
The http/2 session heartbeat for the kube-apiserver watches is not started, because the corresponding client options are not passed to the http/2 request. As a result the created session does not initiate the heartbeat interval.

**Environment**:

holgerkoser:
The http/2 session heartbeat for the kube-apiserver watches is not started, because the corresponding session options are not passed to the http/2 request. As a result the created session does not initiate the heartbeat interval. Dead connections remain in the pool and the dashboard frontend does not get real-time updates anymore.

Fixes #1325

**Release note**:

```bugfix user
Fixed a problem with the heartbeat of `kube-apiserver` watches. Due to this bug the list did not get any realtime updates and the dashboard pod had to be restarted.  
```


grolu (approved):
/lgtm

PR #1330, opened 2022-10-26
Handle non JSON responses when json is expected

### LINKED ISSUE #1329: Handle non JSON responses when json is expected
**What happened**:
If we send a request to a server with the following header
```
Accept: application/json
```
sometimes a server responds with content-type `text/plain`. I this case the client should not fail but return the text as body.

**Environment**:

holgerkoser:
If we send a request to a server with the following header
```
Accept: application/json
```
sometimes a server responds with content-type `text/plain`. In this case the client should not fail but return the text as body.

Fixes #1329 

**Release note**:

```bugfix operator
When the upstream server responds with content-type `text/plain` where `application/json` is expected an error message like `Unexpected token x in JSON at position y` could be seen in the logs. You are now able to see what the actual server response was
```


petersutter (approved):
test is failing with
```
#14 3.942 /app/packages/request/__tests__/client.test.js
#14 3.942   13:7   error  'exp' is assigned a value but never used                                                         no-unused-vars
#14 3.942   13:13  error  'constants' module was deprecated since v6.3.0. Use 'constants' property of each module instead  n/no-deprecated-api
#14 3.942 
#14 3.942 ✖ 2 problems (2 errors, 0 warnings)
```

PR #1334, opened 2022-11-04
Fixed KUBECONFIG file system watcher

### LINKED ISSUE #1331: Changes in tokenFile are not picked up
**What happened**:
When the `Values.serviceAccountTokenVolumeProjection` is enabled the kubelet makes sure to refresh the token as it approaches expiration.
However the changed token file content was not picked up by the dashboard backend and thus resulted in `Unauthorized` errors after the token has expired

The dashboard backend should use the latest token available, once it changes

Related PR: https://github.com/gardener/dashboard/pull/1232, which as of now is not part of a released version

**Environment**:

holgerkoser:
This PR fixes a bug in the file system watcher implementation when change events occur for files or directories that are not expected. With this fix these events are ignored and error during readFile are only logged as errors but do not destroy the watcher.

Fixes #1331

**Release note**:

```bugfix user
NONE
```

### DIFF
--- a/packages/kube-config/__tests__/watcher.test.js
+++ b/packages/kube-config/__tests__/watcher.test.js
@@ -7,6 +7,7 @@
 'use strict'
 
 const EventEmitter = require('events')
+const { basename } = require('path')
 const chokidar = require('chokidar')
 const { globalLogger: logger } = require('@gardener-dashboard/logger')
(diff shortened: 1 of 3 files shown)


petersutter (approved):
/lgtm

grolu (approved):
/lgtm

PR #1337, opened 2022-11-04
[hotfix] Fixed DNS Provider List not visible for non-admin users

grolu:
Regular users cannot access `controllerregistration` resources. Need to add changed `controllerregistration` resources for DNS Providers to list of allowed resources.

Fixes #

**Release note**:

```bugfix user
Fixed DNS Provider List not visible for regular users (non-operators)
```


petersutter (approved):
/lgtm

PR #1343, opened 2022-11-11
Use Clipboard API

### LINKED ISSUE #1341: Copy button does not work
**What happened**:
The copy button for the kubeconfig, gardenctl target command does not work on the cluster access dialog, however it works on the shoot details page
Screenshot 2022-11-11 at 11 13 05

kubeconfig / target command content should be in the clipboard

- Navigate to the shoot list
- Click on the key symbol of any cluster
- Click any copy button

**Workaround**:
Navigate to the shoot details and use the copy buttons there

**Environment**:

holgerkoser:
Due to a [change](https://github.com/vuetifyjs/vuetify/commit/89e3850c5478c7bf0ae6081a95f6d2b39e690e8f) in vuetify the clipboard javascript library does not work anymore when using inside of `v-dialog`. With this PR we have removed this library and directly use the new browser API [navigator.clipboard.writeText](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/writeText) to copy data into the clipboard. The first time this API is used the user now has to confirm that the website is allowed to access the system clipboard. 

Fixes #1341 

**Release note**:

```bugfix user
Fixed a problem that occurs on the cluster access dialog when trying to copy kubeconfig or target command content  into the clipboard. This regression was introduced with release 1.62.0.
```


petersutter:
also remove `"clipboard": "^2.0.11",` from `package.json`

holgerkoser:
> also remove `"clipboard": "^2.0.11",` from `package.json`

✔️

petersutter (approved):
/lgtm

grolu (approved):
/lgtm

PR #1344, opened 2022-11-11
Read SubjectRules also if no project / namespace is selected

### LINKED ISSUE #1342: Can't create projects
**What happened**:
Screenshot 2022-11-11 at 11 19 17
When part of no project, you are not able to create any projects, even though you are authorised to do so

```bash
kubectl auth can-i create projects -A
yes
```

The button should not be disabled

Requirement:
- You a not yet assigned to any project as member
- You are not a gardener operator
- `kubectl auth can-i create projects -A` reports `yes`

Procedure:
- open the dashboard and you are directed to the `home` page.
- The button `+ CREATE YOUR FIRST PROJECT` is disabled

- Issue introduced with dashboard version `1.62.0`
- The subjectRules are not refreshed when no namespace is set https://github.com/gardener/dashboard/blob/f8c5cdea644fa1feab30f9d3c605f9dccd5034d1/frontend/src/router/guards.js#L82-L85

**Workaround**:
To workaround this, any namespace has to be set as query parameter. 
`https://dashboard.example.com/?namespace=magic`

**Environment**:

holgerkoser:
With release `1.62.0` we introduced a regression whereby it was no longer possible to create new projects. The button was always disabled even for users who have the appropriate permission. This PR 

Fixes #1342

**Release note**:

```bugfix user
Fixes a problem when creating new projects that was introduced with release 1.62.0. Now the button is enabled again if the user has the appropriate permission.
```

### DIFF
--- a/frontend/src/router/guards.js
+++ b/frontend/src/router/guards.js
@@ -79,7 +79,10 @@ function ensureDataLoaded (store, localStorage, logger) {
       ]
 
       const namespace = to.params.namespace || to.query.namespace
-      if (namespace && namespace !== store.state.namespace) {
+      if (!namespace) {
+        // fetch cluster scoped roles only (e.g can create projects)
+        promises.push(store.dispatch('refreshSubjectRules', undefined))
+      } else if (namespace !== store.state.namespace) {
         store.commit('SET_NAMESPACE', namespace)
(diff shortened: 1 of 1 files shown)


petersutter (approved):
/lgtm

PR #1346, opened 2022-11-14
Solve problem with issue polling

holgerkoser:
By accident the function `getIssueNumbers` has been removed from the ticketCache. I thought it was no longer used but it is of course used in landscape where  issue polling is enabled. This PR adds this function back to the ticketCache.
 
Fixes #1347

**Release note**:

```bugfix user
Fixed a problem with issue polling for landscapes that cannot be accessed from github webhooks. This regression was introduced with release 1.62.0.
```

### DIFF
--- a/backend/lib/cache/tickets.js
+++ b/backend/lib/cache/tickets.js
@@ -50,7 +50,7 @@ function init () {
   }
 
   function getIssueNumbers () {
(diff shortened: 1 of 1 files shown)


petersutter (approved):
/lgtm

PR #1349, opened 2022-11-17
Robust token refresh test

holgerkoser:
This PR makes the token refresh test more robust. From time to time the `iat` value differed from the expected value. Now the value is expected to be in a range of 5 seconds.

Fixes #

**Release note**:

```bugfix user
NONE
```


grolu (approved):
/lgtm

PR #1351, opened 2022-11-17
Allow update of roles for foreign serviceaccounts

### LINKED ISSUE #1350: Update of foreign service accounts not possible
**What happened**:
It is not possible to change the roles of a foreign serviceaccount in the dashboard. It looks like an existing serviceaccount is always assumed to be in the project itself and not invited from a foreign project. If try to update a foreign serviceaccount an error message occurs.

Screenshot 2022-11-17 at 14 45 44
If a serviceaccount with the same name exists in the project, the wrong serviceaccount is updated.

It should be possible to also update the roles of a foreign serviceaccount.

**Environment**:

holgerkoser:
With this PR it is possible to update the roles of an existing foreign serviceaccount in a project. The description of a foreign serviceaccount cannot be edited anymore.

Fixes #1350

**Release note**:

```bugfix user
With this PR it is possible to update the roles of an existing foreign serviceaccount in a project 
```


petersutter (approved):
/lgtm

however I have noticed that when removing all roles of a foreign service account no error is shown and after a reload on the UI it also disappears from the list. We should improve this (create new issue?)

holgerkoser:
> however I have noticed that when removing all roles of a foreign service account no error is shown and after a reload on the UI it also disappears from the list.

Created issue #1356

grolu (approved):
/lgtm

PR #1353, opened 2022-11-18
Allow filtering and completion for machine types

### LINKED ISSUE #1352: Allow filtering and completion for machine types
The list of available machine types can be very large and it difficult for the user to find and select the desired machine type. It should be possible to reduce the list by entering filter criteria. 

**Why is this needed**:

holgerkoser:
The list of available machine types can be very large and it difficult for the user to find and select the desired machine type. It should be possible to reduce the list by entering filter criteria.

##### Examples:

Only machine types where the name contains  `m5n` 
m5n

Only machine types where the name contains  `m5` and `128Gi` memory 
m5 and 128Gi

Only machine types where the name contains `m5` and `8xl` 
m5 and 8xl

Fixes #1352 

**Release note**:

```feature user
It is now possible to filter the dropdown list of available machine types by name, cpu, gpu and memory
```


dguendisch:
Looks good, very much welcomed, thank you!

vlerenc:
That was fast. 🚀 Thank you! 🎉

frontend/src/components/ShootWorkers/MachineType.vue
```diff
@@ -71,62 +63,110 @@ export default {
...
+      }
+      const { name, cpu, gpu, memory, storage } = item
+      const terms = query.split(/\s+/)
+      const properties = [cpu, gpu, memory]
+      if (storage) {
+        properties.push(storage.type)
+        properties.push(storage.class)
+        properties.push(storage.size)
       }
+      return terms.every(term => name?.includes(term) || properties.includes(term))
```
grolu, line 146:
```suggestion
      return terms.every(term => name?.includes(term) || some(properties, prop => prop?.includes(term)))
```
I would change it to something like this to allow partial match also for the properties. Moreover, you could do a `toLowerCase()` for all the strings before comparing them.

frontend/src/components/ShootWorkers/MachineType.vue
```diff
@@ -71,62 +63,110 @@ export default {
...
-        this.$emit('valid', { id: this.worker.id, valid: this.valid })
+    filter (item, query) {
+      if (!item) {
+        return false
+      }
+      if (typeof query !== 'string') {
+        return true
+      }
+      const { name, cpu, gpu, memory, storage } = item
+      const terms = query.split(/\s+/)
```
petersutter, line 139: is there a difference to `query.split(" ")`?

holgerkoser: > is there a difference to `query.split(" ")`?

image

petersutter (approved):
/lgtm

frontend/src/components/ShootWorkers/MachineType.vue
holgerkoser: May be we should do:
```js
'   a\tb  c\nd\re   '.replace(/(^\s+|\s+$)/g, '').split(/\s+/)
```

frontend/src/components/ShootWorkers/MachineType.vue
holgerkoser: ✔️

frontend/src/components/ShootWorkers/MachineType.vue
holgerkoser: ✔️

PR #1357, opened 2022-11-24
Cleanup secret if secretBinding creation fails

### LINKED ISSUE #1299: Cleanup Secret if SecretBinding creation fails
**What happened**:
When creating a new infrastructure secret the k8s `Secret` is not cleaned up when the creation of the `SecretBinding` fails. https://github.com/gardener/dashboard/blob/94c0908f09ae8a1a672ff69dfc75f62488ef88f7/backend/lib/services/cloudProviderSecrets.js#L222-L224

The `Secret` should be cleaned up

Create an infrastructure secret for azure using the dashboard with an invalid `subscriptionID`. The Validating webhook will fail with `admission webhook "validation.azure.provider.extensions.gardener.cloud" denied the request: field "subscriptionID" in secret garden-example/mysecret must be a valid GUID`. Trying to create the secret with the same name but with a valid `subscriptionID` will fail with the error `Infrastructure Secret name 'mysecret' is already taken. Please try a different name. 
secrets "mysecret" already exists`

**Environment**:

petersutter:
Cleanup secret if secretBinding creation fails

Fixes #1299

**Release note**:

```bugfix user
Fixed an issue where the secret was not cleaned up if the creation of the secretBinding fails when trying to create a new infrastructure secret
```

### DIFF
--- a/backend/lib/services/cloudProviderSecrets.js
+++ b/backend/lib/services/cloudProviderSecrets.js
@@ -220,7 +220,15 @@ exports.create = async function ({ user, namespace, body }) {
   metadata.secretRef = { namespace, name: metadata.name }
   const secret = await client.core.secrets.create(namespace, toSecretResource(body))
 
-  const secretBinding = await client['core.gardener.cloud'].secretbindings.create(namespace, toSecretBindingResource(body))
+  let secretBinding
+  try {
+    secretBinding = await client['core.gardener.cloud'].secretbindings.create(namespace, toSecretBindingResource(body))
+  } catch (err) {
+    logger.error('failed to create SecretBinding, cleaning up secret %s/%s', namespace, secret.metadata.name)
+    await client.core.secrets.delete(namespace, secret.metadata.name)
+
+    throw err
+  }
 
   const cloudProfileName = _.get(secretBinding, 'metadata.labels["cloudprofile.garden.sapcloud.io/name"]')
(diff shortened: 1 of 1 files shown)


backend/lib/services/cloudProviderSecrets.js
```diff
@@ -220,7 +220,15 @@ exports.create = async function ({ user, namespace, body }) {
...
 
-  const secretBinding = await client['core.gardener.cloud'].secretbindings.create(namespace, toSecretBindingResource(body))
+  let secretBinding
+  try {
+    secretBinding = await client['core.gardener.cloud'].secretbindings.create(namespace, toSecretBindingResource(body))
+  } catch (error) {
+    logger.error('failed to create SecretBinding, cleaning up secret %s/%s', namespace, secret.metadata.name)
+    await client.core.secrets.delete(namespace, secret.metadata.name)
+
+    throw error
```
holgerkoser, line 230:
```suggestion
  } catch (err) {
    logger.error('failed to create SecretBinding, cleaning up secret %s/%s', namespace, secret.metadata.name)
    await client.core.secrets.delete(namespace, secret.metadata.name)

    throw err
```

petersutter: ✅

holgerkoser (approved):
/lgtm

PR #1360, opened 2022-11-25
Reset service account

### LINKED ISSUE #1358: New "Reset" button for project service accounts
New "Reset" button for the project service accounts that belong to the project (namespace)
**Why is this needed**:
To simplify the invalidation of all active tokens generated for a service account, for example, when a privileged member of the project leaves the team/changes responsibilities. Without such feature another privileged member will have to delete the current service account and create a new one with the same roles for the same purpose. 
Visualization and simplification of this activity using a special button will help educating consumers about the reset option and will most probably encourage the reset usage when it is necessary.
**What to do on click of the button:** Delete the service account and create a service account with the same name and the same permissions (the same roles).

petersutter:
This PR adds a "Reset Service Account" Button. The service account will be deleted and recreated with the previous owner information and description.

Screenshot 2022-11-25 at 14 41 50
Screenshot 2022-11-25 at 14 42 02

Fixes #1358 

**Release note**:

```feature user
You can now "reset" a service account on the project `Member` page to invalidate all tokens related to this service account.
```

### DIFF
--- a/backend/lib/routes/members.js
+++ b/backend/lib/routes/members.js
@@ -8,6 +8,7 @@
 
 const express = require('express')
 const { members } = require('../services')
+const { UnprocessableEntity } = require('http-errors')
 
 const router = module.exports = express.Router({
   mergeParams: true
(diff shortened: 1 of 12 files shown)


holgerkoser (approved):
/lgtm

grolu (approved):
/lgtm
However, I do not like the term `reset service account` but I was unable to find something better. I know that this is what be do (delete and recreate) but I'm not 100% sure if a user that looks for a way to revoke service account tokens connects this action with `reset` in the first place.

petersutter:
I'm also fine with calling it "Rotate Service Account", wdyt @grolu @donistz ?

grolu:
>I'm also fine with calling it "Rotate Service Account", wdyt @grolu @donistz ?

It was called like that before, right? I think it makes it clearer as this is what we want to achieve...

donistz:
@petersutter and @grolu An action can be named "Rotate...." when it is possible to generate more than one valid credential and to be able to decide wich one to delete and when. What we will offer is not a rotation at all. It is a "reset" becuse we delete something existing and create something new which could cause a downtime and this has to be clear streight from the name of the action. 
Rotation can be called an action organized by the consumers when they [1] create a new robot user with proper permissions  [2] request tokens for it [3] configure these tokens instead of the tokens of the old robot user and finally [4] delete the old robot user. => This cannot be organized with a click of a button.

PR #1361, opened 2022-11-25
gardener-dashboard chart: comment out oidc config

### LINKED ISSUE #1327: Comment out `oidc` in gardener-dashboard values helm chart
Comment out `oidc` property from the gardener-dashboard values helm chart.
https://github.com/gardener/dashboard/blob/248faec9c291a57c65fea4db6313b4a95a198127/charts/gardener-dashboard/values.yaml#L81-L108

**Why is this needed**:
In the past the `oidc` configuration was required. This is not the case anymore. Meanwhile the dashboard can also be used with token login only.

petersutter:
In the past the `oidc` configuration was required. This is not the case anymore. Meanwhile the dashboard can also be used with token login only.

Fixes #1327

**Release note**:

```breaking operator
`.Values.global.oidc` is not defaulted anymore in the `values.yaml` file of the gardener-dashboard helm chart. Especially there is no default value for `.Values.global.oidc.clientId`, which was previously `dashboard`.
```

### DIFF
--- a/charts/gardener-dashboard/values.yaml
+++ b/charts/gardener-dashboard/values.yaml
@@ -93,34 +93,35 @@ global:
 
   # sessionSecret is used for signing and encrytion of session data
   sessionSecret: ~
-  oidc:
-    # well-known URL for OpenID Provider Issuer Discovery
-    issuerUrl: ~
-    # clientId is the public identifier of the gardener-dashboard aplication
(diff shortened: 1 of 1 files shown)


holgerkoser (approved):
/lgtm

grolu (approved):
/lgtm

PR #1362, opened 2022-11-25
Automatic login to dashboard if OIDC is available

holgerkoser:
If the primary login type is `oidc` is available it can be annoying to have to press the login button again and again. No other user input is required in this case and many operators never use the `token` login type. To support this use case without usage of browser based UserScripts, to automatically press the LoginButton via this script, we now directly to configure this behavior in the dashboard settings.

 
Screenshot 2022-11-25 at 15 24 22

Fixes #

**Release note**:

```bugfix user
Added configuration setting for automatic login to dashboard if primary is via an OIDC Provider
```


frontend/src/views/Login.vue
```diff
@@ -101,39 +110,55 @@ export default {
...
+      try {
+        const { data } = await api.getLoginConfiguration()
+        cfg = data
+      } catch (err) {
+        cfg = {
+          loginTypes: ['token'] // at least allow the token login
+        }
       }
-    })
+      const primaryLoginType = head(cfg.loginTypes) || 'oidc'
```
petersutter, line 144: use `getPrimaryLoginType` function?

frontend/src/views/Login.vue
```diff
@@ -101,39 +110,55 @@ export default {
...
-    next(vm => {
-      if (err) {
-        vm.showSnotifyLoginError(err.message)
-        vm.$router.replace('/login')
+      let cfg
+      try {
+        const { data } = await api.getLoginConfiguration()
+        cfg = data
+      } catch (err) {
+        cfg = {
```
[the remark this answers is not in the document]
holgerkoser: ✔️

frontend/src/views/Login.vue
holgerkoser: ✔️

PR #1366, opened 2022-11-28
Keep dashboard URL when user session is expired

### LINKED ISSUE #1365: The dashboard URL is lost if the user token is expired
**What happened**:
If the user has a session cookie with an expired token and tries to open a dashboard URL, he is redirected to the OIDC provider to login again but the dashboard URL is lost in this case. 

After successful login an the OIDC provider the user should be redirected to initially requested dashboard URL.

**Environment**:

holgerkoser:
Fixed a problem with lost dashboard URL during re-login when the user session was expired. This regression was introduced with release 1.62.0.

Fixes #1365 

**Release note**:

```bugfix user
Fixed a problem with lost dashboard URL during re-login when the user session was expired. This regression was introduced with release 1.62.0.
```

### DIFF
--- a/frontend/src/router/guards.js
+++ b/frontend/src/router/guards.js
@@ -40,18 +40,13 @@ function ensureUserAuthenticatedForNonPublicRoutes (store, userManager, logger) 
       return next()
     }
     const user = userManager.getUser()
-    if (!user) {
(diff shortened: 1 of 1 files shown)


petersutter (approved):
/lgtm

PR #1368, opened 2022-11-30
Default to verbose logging if nothing is defined in localStorage

### LINKED ISSUE #1367: Default log level must be verbose
**What happened**:
The default log level should be verbose even if no localStorage entry exists.

If localStorage is deleted verbose log output should be visible

**Environment**:

holgerkoser:
Show all log messages (LogLevel: verbose) if the user has not yet defined anything in his settings

Fixes #1367

**Release note**:

```bugfix user
Show all log messages (LogLevel: verbose) if the user has not yet defined anything in his settings
```


petersutter (approved):
/lgtm

PR #1369, opened 2022-12-02
Retry operation for failed shoots

petersutter:
When a shoot cluster is in failed state, it does not make sense to set the reconcile operation annotation. Instead the retry operation annotation has to be set.
When a user presses the reconcile button under `Lifecycle` it will show a hint that for failed shoots the operation is retried

Screenshot 2022-12-02 at 16 10 57

Fixes #

**Release note**:

```other user
The `retry` operation will be set when triggering the reconcile for failed `Shoot`s
```


grolu (approved):
/lgtm

grolu (approved):
/lgtm

PR #1370, opened 2022-12-02
Fix member dialog validation

### LINKED ISSUE #1356: Improve handling of foreign service accounts
**What happened**:
When removing all roles of a foreign service account no error is shown and after a reload on the UI it also disappears from the list. The whole behavior should be reconsidered and optimized for user-friendliness.

**Environment**:

petersutter:
Fix member dialog validation. A User and foreign service account should always have roles assigned. Otherwise they have to be removed from the project membership.

Fixes #1356

**Release note**:

```bugfix user
Fixed roles validation for invited service accounts and user project member
```

### DIFF
--- a/frontend/src/components/dialogs/MemberDialog.vue
+++ b/frontend/src/components/dialogs/MemberDialog.vue
@@ -173,25 +173,21 @@ export default {
     },
     validators () {
       const validators = {
-        internalRoles: {},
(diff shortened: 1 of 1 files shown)


holgerkoser (approved):
/lgtm

grolu (approved):
/lgtm

PR #1386, opened 2023-01-09
Added machine type filters

grolu:
Follow-up of #1353
Added additional dropdown filters for machine types

Screenshot 2023-01-09 at 12 54 00

Fixes #

**Release note**:

```feature user
Machine types can be filtered for CPU, GPU and memory via dropdown selection
```

### DIFF
--- a/frontend/src/components/ShootWorkers/MachineType.vue
+++ b/frontend/src/components/ShootWorkers/MachineType.vue
@@ -23,6 +23,34 @@ SPDX-License-Identifier: Apache-2.0
       :hint="hint"
       persistent-hint
     >
+      <template v-slot:prepend-item>
+        <div class="d-flex filter-div">
+          <v-select
+            v-if="cpuItems.length > 2"
+            class="ma-1"
+            :items="cpuItems"
+            v-model="cpuFilter"
+            label="CPU Filter"
+          >
+          </v-select>
+          <v-select
+            v-if="memoryItems.length > 2"
+            class="ma-1"
+            :items="memoryItems"
+            v-model="memoryFilter"
+            label="Memory Filter"
+          >
+          </v-select>
+          <v-select
+            v-if="gpuItems.length > 2"
+            class="ma-1"
+            :items="gpuItems"
+            v-model="gpuFilter"
+            label="GPU Filter"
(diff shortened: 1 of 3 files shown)


frontend/src/components/ShootWorkers/MachineType.vue
```diff
@@ -121,14 +166,40 @@ export default {
...
+      return ['all', ...gpuItems.sort((a, b) => (a - b))]
+    },
+    memoryItems () {
+      const memoryItems = uniq(map(this.machineTypes, 'memory'))
+      return ['all', ...memoryItems.sort((a, b) => this.rawMemoryValue(a) - this.rawMemoryValue(b))]
     }
   },
   validations,
   methods: {
+    rawMemoryValue (memoryString) {
```
petersutter, line 185: - we should add tests for this function
- also support more units? https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#meaning-of-memory
- what if the unit is unknown/unsupported?

frontend/src/components/ShootWorkers/MachineType.vue
```diff
@@ -121,14 +166,40 @@ export default {
     },
     hint () {
       return this.notInList ? 'This machine type may not be supported by your worker' : ''
+    },
+    cpuItems () {
+      const cpuItems = uniq(map(this.machineTypes, 'cpu'))
+      return ['all', ...cpuItems.sort((a, b) => (a - b))]
```
petersutter, line 172: what if cpu is in milli format? e.g. 1000m=1(, 100m=0.1) https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#meaning-of-cpu

grolu: I this the case for machines? I thought this only applies to pods?

grolu:
We agreed to not merge this for now. Let's wait for feedback if we need something like this. If yes, we should consider improving the layout (add insets, dividers etc.)
Moreover, we should sync the filters with the v-autocomplete input (xx Gi, yy CPU etc.) or do not support filtering for machine type parameters via the v-autocomplete

PR #1399, opened 2023-01-24
Fixes an occasionally failing test

### LINKED ISSUE #1398: Test `"auth › should successfully refresh a token"` fails occasionally
**What happened**:
The backend test `auth › should successfully refresh a token` in file `test/acceptance/auth.spec.js` fails occasionally. The `iat` claim should only be assumed to be within a certain range.

**Environment**:

holgerkoser:
The backend test `auth › should successfully refresh a token` in file `test/acceptance/auth.spec.js` fails occasionally. With this PR the `iat` claim is assumed to be within a certain range and not a fixed time.

Fixes #1398

**Release note**:

```bugfix user
NONE
```

### DIFF
--- a/backend/test/acceptance/auth.spec.js
+++ b/backend/test/acceptance/auth.spec.js
@@ -404,7 +404,7 @@ describe('auth', function () {
     expect(payload).toEqual({
       jti: expect.stringMatching(/^[a-z0-9-]+$/),
       id,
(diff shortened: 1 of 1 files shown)


grolu (approved):
/lgtm

petersutter (approved):
/lgtm

PR #1402, opened 2023-01-25
Docs: Use gardenlogin kubeconfig

petersutter:
- Updates "Connect Kubectl" documentation to use the gardenlogin kubeconfig
- The `Kubeconfig - Gardenlogin` section is now always shown.

Fixes #

**Release note**:

```bugfix user

```


grolu:
The new screenshots now use dark mode. Is it ok that we mix dark mode and non dark mode images in the docs? I mean the old ones were outdated inc case of color scheme anyway... But I think we should agree for new screenshots to always use dark or light scheme. WDYT?

grolu (approved):
/lgtm

PR #1408, opened 2023-01-30
Shoot owners and Gardener operators see the same monitoring links

istvanballok:
**What this PR does / why we need it**:

Shoot owners and Gardener operators see the same monitoring links

Previously, based on the user's role: Gardener Operator (cluster-admin) or
regular shoot owner, the Gardener Dashboard showed different links in the
'Logging and Monitoring' panel of a shoot cluster.

Shoot owners saw a link to a Grafana and optionally to an alert manager instance
if alerting was configured for the specific shoot cluster. (See [alerting](https://gardener.cloud/docs/gardener/monitoring/alerting/))

Gardener operators saw a link to a different Grafana, a Prometheus and
optionally to an alert manager instance.

The Gardener Dashboard also provides access to the password that is needed to
access the links above.

For shoot owners, the shoot cluster's monitoring secret in the project namespace
in the virtual-garden cluster was used. For Gardener operators, the credentials
were taken from a secret in the shoot's control plane in the seed cluster.
Accessing the secret in the seed cluster required some special handling in the
Gardener Dashboard.

With the Gardener PR [7007](https://github.com/gardener/gardener/pull/7007), both shoot owners and Gardener operators shall have
equal access to the observability components. The two different Grafana
instances are replaced by a single Grafana instance that is accessed by both
roles.

With this Gardener Dashboard PR, both shoot owners and Gardener operators will
see the same monitoring links: a link to a Grafana, a link to a Prometheus and
optionally a link to an alert manager instance. The password will be taken from
the monitoring secret in the project namespace for both roles.

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

/cc @grolu @holgerkoser @petersutter @rickardsjp 

The Gardener dashboard no longer needs to access the shoot's control plane in
the seed cluster to get the password for the Gardener operators. The majority of
the source code changes in this PR is related to removing the no longer needed
functionality (`/seed-info`). Please check the individual commits as well.

Note that this PR should be released in the same cycle as Gardener PR 7007.

<details>
<summary>Documentation about a "local" development setup to test both roles (cluster admin and project viewer) in the Dashboard UI with a development version of Gardener.</summary>

> ### Local setup
>
> Connect the locally executed Gardener Dashboard to a Gardener installation in
> the remote local setup.
>
> https://github.com/gardener/gardener/blob/master/docs/deployment/getting_started_locally.md#remote-local-setup
>
> #### Connect kubectl to the remote local setup
>
[...]

Issue #1318, opened 2022-10-17
Unify the user and operator grafanas


petersutter:
**What would you like to be added**:
With https://github.com/gardener/gardener/pull/6749 the ingress with the `gu` prefix that was used for the grafana operators will be dropped. Instead the ingress with the `gu` prefix will be used both for the operators and the users.
The dashboard needs to be adapted accordingly

**Why is this needed**:

istvanballok:
minor typo:

> With https://github.com/gardener/gardener/pull/6749 the ingress with the -gu- go prefix

grolu:
fixed with #1408

PR #1409, opened 2023-01-31
Add plutono URLs in shoots dashboards

nickytd:

This PR add plutono urls as part of the migration of grafana/loki to plutono/vali. After expiring the logs retention period of two weeks we shall remove grafana URLs and remain only with plutono URLs.

Adding plutono URLs shall be aligned with the gardener release containing plutono workload.
Later (after the standard log retention period) removing grafana URLs shall be aligned with the gardener release removing grafana and loki workloads

Fixes https://github.com/gardener/backlog/issues/62

**Release note**:

```feature user
This release of gardener dashboards introduces the plutono URLs in the Logging and Monitoring tile.
```

### DIFF
--- a/frontend/src/components/ClusterMetrics.vue
+++ b/frontend/src/components/ClusterMetrics.vue
@@ -8,9 +8,9 @@ SPDX-License-Identifier: Apache-2.0
   <v-list>
     <link-list-tile
       icon="mdi-developer-board"
-      app-title="Grafana"
-      :url="grafanaUrl"
-      :url-text="grafanaUrl"
+      app-title="Plutono"
+      :url="plutonoUrl"
+      :url-text="plutonoUrl"
       :is-shoot-status-hibernated="isShootStatusHibernated"
     ></link-list-tile>
     <link-list-tile
@@ -45,8 +45,8 @@ export default {
   },
   mixins: [shootItem],
   computed: {
-    grafanaUrl () {
-      return get(this.shootItem, 'info.grafanaUrl', '')
+    plutonoUrl () {
+      return get(this.shootItem, 'info.plutonoUrl', '')
     },
     prometheusUrl () {
       return get(this.shootItem, 'info.prometheusUrl', '')
--- a/frontend/src/store/modules/shoots/index.js
+++ b/frontend/src/store/modules/shoots/index.js
@@ -176,7 +176,7 @@ const actions = {
 
       if (info.seedShootIngressDomain) {
         const baseHost = info.seedShootIngressDomain
-        info.grafanaUrl = `https://gu-${baseHost}`
+        info.plutonoUrl = `https://pl-${baseHost}`
 
         info.prometheusUrl = `https://p-${baseHost}`
 


petersutter:
How will this PR play along with https://github.com/gardener/dashboard/pull/1408? Here we won't fetch the operator credentials from the seed anymore. Also there will only be one grafana and prometheus link for the operators and the users, same as with the credentials.

istvanballok:
> How will this PR play along with https://github.com/gardener/dashboard/pull/1408? Here we won't fetch the operator credentials from the seed anymore. Also there will only be one grafana and prometheus link for the operators and the users, same as with the credentials.

I discussed with @nickytd *offline* that this PR should come after #1408:
- once https://github.com/gardener/gardener/pull/7007 is merged, https://github.com/gardener/gardener/pull/7318 will be rebased, and it will be adopted to the new state that there is only a single Grafana instance
- consequently, after #1408 is merged, this PR will also have to be rebased and reworked

and I am also proposing a different approach:
- make sure that https://github.com/gardener/gardener/pull/7318 creates an ingress for Plutono both with the old `gu-` prefix and with the new `plu-` prefix, both for old and new shoot clusters
- this means that the current Dashboard links (`gu-`) will work when https://github.com/gardener/gardener/pull/7318 is deployed, for old and new shoot clusters, before and after reconciliation
- in a later release cycle, we shall replace the Dashboard links (`gu-` -> `plu-`)
- when the new Dashboard version is deployed in the landscape, all the Dashboard links will work, because by that point in time, all the shoot clusters will have a `plu-` ingress
- in a later release cycle, the `gu-` prefix can be removed from Gardener as a cleanup activity

I think the proposed approach is superior to this PR, because with this PR, the Dashboard would show URLs for both Plutono and Grafana, i.e. 2 links, but from the end user's perspective, it is mostly the same thing (except for the logs, see below), furthermore, depending on the shoot cluster (if it is new or old, before or after reconciliation) only one or the other link would work, which would lead to a poor user experience. So I propose the rework the Gardener change (https://github.com/gardener/gardener/pull/7318/commits/ab2e01c827b87118ddd41eedd5b3bb4d530d3c0f) to include both ingress prefixes, and then rework / replace this PR to change `gu-` to `plu-` in place in a later release.

> After expiring the logs retention period of two weeks we shall remove grafana URLs and remain only with plutono URLs.

Regarding the motivation for providing access to the old logs via Grafana because they are not migrated from Loki to Vali, maybe we could consider other alternatives to avoid the situation that we show 2 links to the end users and in some cases, only one of them works. If both work, the logs would be different (loki vs. vali) but the metrics (prometheus) would be the same, which is also somewhat confusing.

Maybe we could discuss this and other approaches later also offline. cc @rickardsjp

nickytd:
👍 Fully agree with the proposal by @istvanballok. We shall show only the plutonto URL in the dashboard. Hence this PR will be adopted accordingly to the https://github.com/gardener/gardener/pull/7007

holgerkoser:
Please fix the problem in `frontend/src/components/ClusterMetrics.vue`
```
#18 [builder 12/24] RUN yarn workspace @gardener-dashboard/frontend    run lint
#18 sha256:4bbbf052167d3e0cacf6b5c34a753a9f6dd49c1dda303b6649f788acb11ecedd
#18 21.66 
#18 21.66 /app/frontend/src/components/ClusterMetrics.vue
#18 21.66   30:21  error  'v-else' directives require being preceded by the element which has a 'v-if' or 'v-else-if' directive  vue/valid-v-else
#18 21.66 
#18 21.66 ✖ 1 problem (1 error, 0 warnings)
#18 21.66 
#18 ERROR: executor failed running [/bin/sh -c yarn workspace @gardener-dashboard/frontend    run lint]: exit code: 1
```
I think what you are trying to achieve is something like:
```html
<template v-if="isAdmin">
    <link-list-tile
      icon="mdi-developer-board"
      app-title="Grafana"
      :url="grafanaUrlOperators"
      :url-text="grafanaUrlOperators"
      :is-shoot-status-hibernated="isShootStatusHibernated"
    ></link-list-tile>
    <link-list-tile
      icon="mdi-developer-board"
      app-title="Plutono"
      :url="plutonoUrlOperators"
      :url-text="plutonoUrlOperators"
      :is-shoot-status-hibernated="isShootStatusHibernated"
    ></link-list-tile>
</template>
<template v-else>
    <link-list-tile
      icon="mdi-developer-board"
      app-title="Grafana"
      :url="grafanaUrlUsers"
      :url-text="grafanaUrlUsers"
      :is-shoot-status-hibernated="isShootStatusHibernated"
    ></link-list-tile>
    <link-list-tile
      icon="mdi-developer-board"
      app-title="Plutono"
      :url="plutonoUrlUsers"
      :url-text="plutonoUrlUsers"
      :is-shoot-status-hibernated="isShootStatusHibernated"
    ></link-list-tile>
</template>
```
anyway I think it would be much better to do this `if (this.isAdmin) { ... } else { ... }` in the code and not in the template because markup is not affected at all.
```html
    <link-list-tile
      icon="mdi-developer-board"
      app-title="Grafana"
      :url="grafanaUrl"
      :url-text="grafanaUrl"
      :is-shoot-status-hibernated="isShootStatusHibernated"
    ></link-list-tile>
    <link-list-tile
      icon="mdi-developer-board"
      app-title="Plutono"
      :url="plutonoUrl"
      :url-text="plutonoUrl"
      :is-shoot-status-hibernated="isShootStatusHibernated"
    ></link-list-tile>
```
and with the computed properties:
```js
    grafanaUrl () {
      const role = this.isAdmin ?  'Operator' : 'User'
      return get(this.shootItem, `info.grafanaUrl${role}s`, '')
    },
    plutonoUrl () {
      const role = this.isAdmin ?  'Operator' : 'User'
      return get(this.shootItem, `info.plutonoUrl${role}s`, '')
    },
```

nickytd:
Thanks for the feedback. 👍 The proposal really improves clarity of the code. Let me know if the current proposal is ok. If it is ok I will wait for the planned updates in gardener/gardener and will adjust the PR here accordingly.

holgerkoser:
There are still lint errors:
```
#18 [builder 12/24] RUN yarn workspace @gardener-dashboard/frontend    run lint
#18 sha256:3d9172004561587350397c996381a94a27498129830eb1b8331597e0c3253a2a
#18 21.76 
#18 21.76 /app/frontend/src/components/ClusterMetrics.vue
#18 21.76   78:34  error  Multiple spaces found before ''Operator''  no-multi-spaces
#18 21.76   82:34  error  Multiple spaces found before ''Operator''  no-multi-spaces
#18 21.76 
#18 21.76 ✖ 2 problems (2 errors, 0 warnings)
#18 21.76   2 errors and 0 warnings potentially fixable with the `--fix` option.
#18 21.76 
```
and I think something in template is still not as it should be:
```diff
-    <link-list-tile
-      icon="mdi-developer-board"
-      app-title="Plutono"
-      :url="plutonoUrlOperators"
-      :url-text="plutonoUrlOperators"
-      :is-shoot-status-hibernated="isShootStatusHibernated"
-    ></link-list-tile>
-   <link-list-tile v-if="isAdmin"
+  <link-list-tile
      icon="mdi-developer-board"
      app-title="Grafana"
      :url="grafanaUrl"
      :url-text="grafanaUrl"
      :is-shoot-status-hibernated="isShootStatusHibernated"
    ></link-list-tile>
-  <link-list-tile v-if="isAdmin"
+  <link-list-tile
      icon="mdi-developer-board"
      app-title="Plutono"
      :url="plutonoUrl"
      :url-text="plutonoUrl"
      :is-shoot-status-hibernated="isShootStatusHibernated"
    ></link-list-tile>
```

frontend/src/components/ClusterMetrics.vue
```diff
@@ -67,11 +74,13 @@ export default {
     ...mapGetters([
       'isAdmin'
     ]),
-    grafanaUrlOperators () {
-      return get(this.shootItem, 'info.grafanaUrlOperators', '')
+    grafanaUrl () {
+      const role = this.isAdmin ?  'Operator' : 'User'
```
holgerkoser, line 78:
```suggestion
      const role = this.isAdmin ? 'Operator' : 'User'
```

frontend/src/components/ClusterMetrics.vue
```diff
@@ -67,11 +74,13 @@ export default {
...
+      return get(this.shootItem, `info.grafanaUrl${role}s`, '')
     },
-    grafanaUrlUsers () {
-      return get(this.shootItem, 'info.grafanaUrlUsers', '')
+    plutonoUrl () {
+      const role = this.isAdmin ?  'Operator' : 'User'
```
holgerkoser, line 82:
```suggestion
      const role = this.isAdmin ? 'Operator' : 'User'
```

frontend/src/store/modules/shoots/index.js
```diff
@@ -183,7 +183,9 @@ const actions = {
       if (info.seedShootIngressDomain) {
         const baseHost = info.seedShootIngressDomain
         info.grafanaUrlUsers = `https://gu-${baseHost}`
+        info.plutonoUrlUsers = `https://plu-${baseHost}`
         info.grafanaUrlOperators = `https://go-${baseHost}`
+        info.plutonoUrlOperators = `https://plo-${baseHost}`
```
holgerkoser, line 188: Why is the prefix for Users `plu` but for Operators `plo`?

istvanballok:
Hi @nickytd,

I think we could rework this PR:
- as we agreed above, to prevent confusing the shoot owners, we should try to avoid showing both the Grafana and the Plutono URLs at the same time
- otherwise, https://github.com/gardener/dashboard/pull/1408 is soon to be merged, so there are going to be no operator and user versions of Grafana / Plutono any more.
- @nickytd if you don't yet have a local development setup for the Dashboard, that should help to catch and fix the linter issues as pointed out by @holgerkoser above, please check the description of #1408 where I documented the local setup I used. I could also help you off-line to get it up and running.

Furthermore, we could try to make the Grafana/Plutono URL handling dynamic. Currently, the Grafana/Plutono prefix is hard coded in the Dashboard (changing it is the main purpose of this PR) but it is also available in the `<shoot>.monitoring` secret as the URL annotation and this approach is also used by the Gardener integration test [here](https://github.com/gardener/gardener/blob/6fae0b0e6c4c336474a0f1cf96015520940136da/test/e2e/gardener/shoot/internal/rotation/observability.go#L94-L122) to get the valid observability endpoint. So I'd suggest that we (@nickytd, @rickardsjp) try to change the Dashboard logic to pick the URL from the monitoring secret. That way the Grafana / Plutono prefix will no longer be hard coded in the code base and the link that will be shown by the Dashboard will be correct even after the Dashboard upgrade and before the shoot cluster is reconciled. With the current approach we can not consider the "Gardener version" of the shoot cluster and hence can not know if the Dashboard should show a Grafana or a Plutono link before and after the shoot cluster is reconciled.

nickytd:
@istvanballok 👍 fully agree

nickytd:
After #1318 the change now only renames the single grafana url to plutono.

grolu:
@nickytd does this depend on a gardener PR. Can you link the PR so that we can track it

nickytd:
This PR depends on https://github.com/gardener/gardener/pull/7318 and shall be merged only after #7318. The current state of the PR is Draft.

nickytd:
It is a draft version, there is no need to close/open.

holgerkoser:
We have finished the migration of the frontend to the new major version of vuejs and vuetify. This means this PR is completely outdated and we would like to close it. If the changes are still required I would propose to create a new PR based to the current master branch. What do you think?

PR #1414, opened 2023-02-04
Show machine types / volume types independent of zone selection

grolu:
Machine types / volume types depend on the selected infrastructure / cloud profile. They can be flagged as unavailable for each zone independently in the cloud profile. The old code filtered the machine types / volume types according to the selected zone(s) for a worker. 
However, we recently learned that the machine / volume types available for a zone can be different depending on the infrastructure account. This is because virtual zones are mapped to physical zones for each provider account individually. It seems that we are at the moment only able to filter items that are not available in all zones (in the whole region) as the subset of available zones should be identical independent of the provider account.
Gardener will likely remove the unavailable machine / volume types from the zones in cloud profile and add a list of unavailable items for each region.

Fixes #

**Release note**:

```bugfix user
Gardener Dashboard now shows all available machine types / volume types for a region independent of the zone selection. As virtual zones are mapped to physical zones for each infrastructure account individually, Gardener Dashboard is unable to determine which zone - type combination will work for the shoot workers
```


petersutter (approved):
/lgtm

holgerkoser (approved):
/lgtm

PR #1416, opened 2023-02-06
Fixed a couple of issues with focus mode

grolu:
A couple of fixes
Focus Mode
- Fixed greyed out not reactive as v-if inside tooltip
- Tooltip for disabled filters did not work
- sortItems did cause mutation error as items are passed from store while generating sorted uids array, so the function must not alter the passed item array

Cluster List
- Shoot Messages on cluster list did increase row height

Fixes #

**Release note**:

```bugfix user
Focus Mode: Stale clusters sometimes did not show as greyed out
```


frontend/src/store/modules/shoots/getters.js
```diff
@@ -256,7 +256,7 @@ export default {
       const inverse = sortOrder === 'desc' ? -1 : 1
       switch (sortBy) {
         case 'k8sVersion': {
-          items.sort((a, b) => {
+          return [...items].sort((a, b) => {
```
petersutter, line 259: was this part of the fix? why is it necessary?

grolu: Yes, because of

>sortItems did cause mutation error as items are passed from store while generating sorted uids array, so the function must not alter the passed item array

However we should discuss this. maybe there is a better solution

holgerkoser: This is not the right place to fix this issue. This function should not assume that the items are references to the store state. The clone should be done before calling this function. There we now it comes from the store.

grolu: moved the copy to function caller

petersutter (approved):
/lgtm

PR #1422, opened 2023-02-08
Added access restrictions to issue template

### LINKED ISSUE #974: Add "Access Restrictions" to cluster list and to templates/text when creating cluster issues
**Why is this needed**:
The "Access Restrictions" (https://github.com/gardener/dashboard/issues/689) are pretty important when consumers have them in place.
Hence admins/operators should have a direct visual feedback when the cluster in question has access restrictions imposed.

I'd want to have a visual feedback on clusters with imposed "Access Restrictions":
- some kind of icon in the cluster list clearly showing that "Access Restrictions" are in place (ideally with some hover/popup on the details of the "Access Restrictions").
- some kind of textual hint on a github issue created directly from the gardener dashboard, e.g. "[Restricted Access!][project/shoot] bla" or a label or some hint in the issue content, so that also admin/ops that interact with github primarily are prominently made aware of access restrictions.

grolu:
If a cluster has access restrictions defined, we should also make this information available in the issue template.
Also fixed a layout issue with the access restrictions chips.

Fixes #974

See attached issue:
>some kind of icon in the cluster list clearly showing that "Access Restrictions" are in place (ideally with some hover/popup on the details of the "Access Restrictions")

We already added a `Access Restrictions` column where we show the chips. I thought about possible improvements like the possibility to hide Access Restrictions (via settings) in case they do not apply to sb or to just show an icon (a lock etc.) with a popper to keep the information on the list as condensed as possible. In the end I decided against it as this could lead to situations where a user is unaware about access restrictions without remembering that they are just hidden for him. Moreover, the column can be removed from the list if someone does not need it.

**Release note**:

```feature operator
Added access restrictions to issue template
```


charts/gardener-dashboard/values.yaml
```diff
@@ -157,13 +157,20 @@ global:
     #   # - `${projectName}`: name of the project
     #   # - `${utcDateTimeNow}`: current date-time in utc format
     #   # - `${seedName}`: shoot's seed name
+    #   # - `${accessRestrictions}`: shoot's selected accessRestrictions
```
petersutter, line 160: As we pass an object here we should give details about it's properties, WDYT?

frontend/src/components/ShootAccessRestrictions/AccessRestrictionChips.vue
```diff
@@ -6,7 +6,7 @@ SPDX-License-Identifier: Apache-2.0
 
 <template>
   <div>
-    <v-row v-for="{ key, title, description, options } in selectedAccessRestrictions" :key="key" class="mx-0 mb-1">
+    <div v-for="{ key, title, description, options } in selectedAccessRestrictions" :key="key" class="d-flex">
```
holgerkoser, line 9: Why do we need the outer div? Isn't one div sufficient?

grolu: The outer div is because the inner div is in a v-for loop and template must contain only one element. However, the v-row caused layout issues when used inside the v-list component. So I replaced it with div

grolu: The outer one adds the row layout

charts/gardener-dashboard/values.yaml
```diff
@@ -157,13 +157,20 @@ global:
...
     #   issueDescriptionTemplate: |
     #     ## Which cluster is affected?
 
     #     `Cluster Details Dashboard Link`: [${projectName}/${shootName}](${shootUrl})
     #     `Operating System`: ${machineImageNames}
     #     `Platform`: ${providerType}
-
+    #     <% if(accessRestrictions.length) { %>
```
holgerkoser, line 167: Is this a good place for documentation of such complicated things. I would propose to create a standalone docs page for this topic and refer it by link from here.

charts/gardener-dashboard/values.yaml
grolu: added additional descriptions

petersutter (approved):
/lgtm

PR #1424, opened 2023-02-09
Adjust the shoot ingress domain calculation

istvanballok:

Adjust the shoot ingress domain calculation

With Gardener 1.64, a bug was fixed that affected technical IDs and the ingress calculations.

In older shoot clusters with a single `-` in the technical ID, the ingress domain contained the prefix `shoot-` (this was a bug), because Gardener tried to remove `shoot--` and that was not found so the `shoot-` prefix stayed in the ingress domain.

Now that the bug is fixed, the `shoot-` prefix will be removed and the logic in the Gardener Dashboard has to be adjusted as well.

See https://github.com/gardener/gardener/pull/7454

Fixes #

/cc @holgerkoser @petersutter 

**Release note**:

```bugfix user
Adjust the shoot ingress domain calculation to a bugfix in Gardener 1.64
```

### DIFF
--- a/backend/lib/services/shoots.js
+++ b/backend/lib/services/shoots.js
@@ -293,7 +293,7 @@ exports.info = async function ({ user, namespace, name }) {
 
   if (shoot.spec.seedName) {
     const seed = getSeed(getSeedNameFromShoot(shoot))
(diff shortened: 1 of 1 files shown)


holgerkoser (approved):
/lgtm

PR #1429, opened 2023-02-17
Remove the use of the enableBasicAuthentication field

dimitar-kostadinov:
Remove the use of the enableBasicAuthentication field

Part of https://github.com/gardener/gardener/issues/6911

**Release note**:

```bugfix user
NONE
```

### DIFF
--- a/docs/usage/project-operations.md
+++ b/docs/usage/project-operations.md
@@ -252,8 +252,6 @@ To create a new cluster using the command line, you need a YAML definition of th
             location: Europe/Kiev
       kubernetes:
         allowPrivilegedContainers: true
-        kubeAPIServer:
-          enableBasicAuthentication: false
         kubeControllerManager:
           nodeCIDRMaskSize: 24
         kubeProxy:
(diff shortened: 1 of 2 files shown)


petersutter:
@dimitar-kostadinov from what I can see in the referenced link it only drops the **defaulting** of the `EnableBasicAuthentication` field.

In the dashboard we already disable this field by default. But we let users choose to enable it again if they wish. We also discourage users to enable it.
As long as it is allowed by gardener to change the field we will offer this possibility to endusers.

dimitar-kostadinov:
> @dimitar-kostadinov from what I can see in the referenced link it only drops the **defaulting** of the `EnableBasicAuthentication` field.
> 
> In the dashboard we already disable this field by default. But we let users choose to enable it again if they wish. We also discourage users to enable it. As long as it is allowed by gardener to change the field we will offer this possibility to endusers.

@petersutter The final goal is to remove the `EnableBasicAuthentication` field from the shoot spec in Gardener API, as basic authentication has been removed in Kubernetes v1.19+. Currently Gardener supports k8s [v.1.20+](https://github.com/gardener/gardener/blob/master/docs/usage/supported_k8s_versions.md).
In preparation, all uses of this field in source code, examples, and docs should be removed.

petersutter (approved):
/lgtm

grolu (approved):
/lgtm

PR #1430, opened 2023-02-22
Added optional region field to aws dns secret dialog

### LINKED ISSUE #1410: Provide optional field for region in secretbinding in Dashboard
an optional field for region is required in secretbinding in Dashboard for AWS
**Why is this needed**:
if there is no region field in secret, DNS controller will connect to endpoint of route53 of default region.  It's not big problem for global AWS. but if we use AWS China IaaS, the endpoint url is different. In this case, user needs to set region in secret.

grolu:
Added optional region field to aws dns secret dialog

Screenshot 2023-02-22 at 16 50 46

Fixes #1410

**Release note**:

```feature user
Added optional region field to aws dns secret dialog
```


petersutter:
It would be good if we can provide a hint on why the user would want to set this optional field. 
Also, does a value help make sense here?

grolu:
> It would be good if we can provide a hint on why the user would want to set this optional field. Also, does a value help make sense here?

I improved the hint: `Overwrite default region of Route 53 endpoint. Required for certain regions. Example value: eu-central-1`

holgerkoser (approved):
/lgtm

petersutter (approved):
/lgtm

PR #1438, opened 2023-03-10
Support alpine architecture `x86_64` and `aarch64`

### LINKED ISSUE #1437: Docker build does not work on ARM
**What happened**:
The Dockerimage currently only works on `x86_64` architecture. We should at least support also ARM because the new Macbooks have an ARM processor.

**Environment**:

holgerkoser:
Support alpine architecture `x86_64` and `aarch64`

Fixes #1437

**Release note**:

```bugfix developer
The make build command now also works for new macbooks with ARM processor
```

### DIFF
--- a/.ci/pipeline_definitions
+++ b/.ci/pipeline_definitions
@@ -12,6 +12,10 @@ dashboard:
           'inject-commit-hash'
         inject_effective_version: true
       publish:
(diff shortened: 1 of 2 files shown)


grolu (approved):
/lgtm

petersutter (approved):
/lgtm

PR #1400, opened 2023-01-24
enh: listen on lease obj to trigger ticket polling + restructure GH w…

sven-petersen:
…ebhook

**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Release note**:

```feature operator
Support using GitHub webhooks when running multiple backend instances. Previously, only one randomly selected instance would receive the webhook, causing only that instance to update its data accordingly. Now, if any instance receives a webhook, all instances are notified and can synchronize with GitHub, ensuring eventual consistency. This improvement will enhance scalability in the future.
```

PR #1454, opened 2023-03-24
fix: deleting just created comment was not propagated to UI

sven-petersen:

Fixes an issue with the previously merged [PR#1400](https://github.com/gardener/dashboard/pull/1400).
Previously the `updated_at` field, provided in the data from the GitHub webhook, from the the issue/comment was used as value for the `renewTime` field in the patch operation of the k8s Lease-object. If an issue (or comment) is created and deleted immediatly after the `updated_at` field remains the same. Hence the data in the patch call is identical to the data of the resource which does not trigger our change listener on the lease object.

So we now always use the current system time instead to ensure that there always is a change when patching the resource.

Fixes #

**Release note**:

```bugfix user

```

### DIFF
--- a/backend/test/acceptance/webhook.spec.js
+++ b/backend/test/acceptance/webhook.spec.js
@@ -53,32 +53,39 @@ describe('github', function () {
       jest.resetAllMocks()
     })
 
-    it('should handle valid github webhook', async () => {
-      // do not enable that globally (e.g. in beforeEach) because express error handlers
-      // won't be called when fake timers are enabled.
-      jest.useFakeTimers().setSystemTime(now)
-      const body = JSON.stringify({ action: 'opened', issue: newGithubIssue })
+    describe('handle valid github webhook', () => {
+      beforeAll(() => {
+        // Be careful when adding other tests in this block. E.g. express error handlers
+        // won't be work when fake timers are enabled as the test in that case never complete.
+        jest.useFakeTimers().setSystemTime(now)
+      })
 
-      await agent
-        .post('/webhook')
-        .set('x-github-event', githubEvent)
-        .set('x-hub-signature-256', fixtures.github.createHubSignature(body))
-        .type('application/json')
-        .send(body)
-        .expect(204)
+      afterAll(() => {
+        jest.useRealTimers()
+      })
 
-      jest.useRealTimers()
+      it('should succeed if system time matches new value for lease renewTime', async () => {
+        const body = JSON.stringify({ action: 'opened', issue: newGithubIssue })
 
-      expect(mockRequest).toBeCalledTimes(1)
-      expect(mockRequest).toBeCalledWith(
-        expect.anything(),
-        {
-          spec: {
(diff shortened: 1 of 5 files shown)


gardener-robot-ci-2:
Thank you @sven-petersen for your contribution. Before I can start building your PR, a member of the organization must set the required label(s) {'reviewed/ok-to-test'}. Once started, you can check the build status in the PR checks section below.

backend/test/github.spec.js
```diff
@@ -29,29 +27,32 @@ const advanceTimersAndFlushPromises = async (ms) => {
...
+    })
+
     beforeEach(() => {
       mergePatchStub = jest.spyOn(dashboardClient['coordination.k8s.io'].leases, 'mergePatch')
       mergePatchStub.mockResolvedValue({})
     })
 
-    it('should throw an error in case of unknown event', async () => {
-      await expect(handleGithubEvent('unknown', null)).rejects.toThrow(UnprocessableEntity)
+    afterAll(() => {
```
petersutter, line 50: move immediately under `beforeAll`

petersutter (approved):
/lgtm

backend/test/acceptance/webhook.spec.js
```diff
@@ -64,33 +67,37 @@ describe('github', function () {
...
         .post('/webhook')
-        .set('x-github-event', githubEvent)
+        .set('x-github-event', 'unknown_event')
         .set('x-hub-signature-256', fixtures.github.createHubSignature(body))
         .type('application/json')
         .send(body)
-        .expect(422)
+        .expect((res) => {
+          expect(res.status).toEqual(422)
+        })
```
holgerkoser, line 95:
```suggestion
        .expect(422)
```

backend/test/acceptance/webhook.spec.js
```diff
@@ -42,18 +45,18 @@ describe('github', function () {
...
 
     afterEach(() => {
       jest.resetAllMocks()
     })
 
-    it('should handle valid github webhook', async function () {
+    it('should handle valid github webhook', async () => {
+      // do not enable that globally (e.g. in beforeEach) because express error handlers
+      // won't be called when fake timers are enabled.
+      jest.useFakeTimers().setSystemTime(now)
```
holgerkoser, line 59: I would prefer to do this in beforeAll and AfterAll or do the reset in finally.

grolu (approved):
/lgtm

backend/test/acceptance/webhook.spec.js
sven-petersen: Enabling the fake timers by `beforeAll`/`afterAll` of the parent `describe` block would break the test `should error on invalid github webhook event` because the fake timers somehow mess with express's error handling and cause the test case to never complete.

As only the test `should handle valid github webhook` requires the fake timers I assumed it is sensible to enable/disable them on the lowest "layer" possible to prevent side effects with other and future tests.

holgerkoser (approved):
/lgtm

PR #1462, opened 2023-03-29
fix animation for dialogs that have the data when mounted

grolu:
updated() not called in case component has the data in mounted()
=>Need to set this.animateOnAppear in mounted for components that map data from store (shootStaging)

Fixes #

**Release note**:

```bugfix user

```


holgerkoser (approved):
/lgtm

petersutter (approved):
/lgtm

PR #1490, opened 2023-05-11
Adapt `NetworkPolicy`s

rfranzke:

/area security networking
/kind enhancement

This PR adapts the `NetworkPolicy`s as following:

- egress traffic to DNS and to the virtual garden kube-apiserver is allowed
- egress to public networks is allowed (to reach GitHub API and seed/shoot API server)

**Release note**:

```other operator
The Helm chart are now adapted such that they work well in garden cluster with enabled `NetworkPolicy` protection (default since `gardener/gardener@v1.71` when garden cluster is managed by `gardener-operator`).
```


petersutter (approved):
/lgtm

PR #1493, opened 2023-05-17
Set dashboard ingress.spec.ingressClassName to nginx

StenlyTU:

This PR adds `ingressClassName: nginx` to the Dashboard ingress.

With https://github.com/gardener/gardener/pull/7945 we want to install nginx-ingress-controller through gardener-operator.

The currently deployed runtime nginx-ingress-controller uses the following argument `--watch-ingress-without-class=true` which reads the Class through deprecated kubernetes.io/ingress.class annotation.

With that change `--watch-ingress-without-class=true` argument is not needed when the nginx-controller is installed through the gardener-operator. Will leave the old annotation for now.

Fixes #

**Release note**:

```other operator
Set dashboard ingress.spec.ingressClassName to nginx
```


petersutter (approved):
/lgtm

holgerkoser (approved):
/lgtm

PR #1498, opened 2023-06-01
Remove `kubernetes.io/ingress.class` annotation in values.yaml

petersutter:
Fixes an issue where the helm deployment failed with the error 
```
Helm upgrade failed: failed to create resource: Ingress.extensions "gardener-dashboard-ingress" is invalid: annotations.kubernetes.io/ingress.class: Invalid value: "nginx": can not be set when the class field is also set
```

Fixes #

**Release note**:

```bugfix operator
Fixed an issue where the helm deployment failed with the error `annotations.kubernetes.io/ingress.class: Invalid value: "nginx": can not be set when the class field is also set`
```
```breaking operator
The default ingress class annotation under `Values.global.dashboard.ingress.annotations['kubernetes.io/ingress.class']` will not be set anymore. Instead, the ingress class name will be set using `Values.global.dashboard.ingress.ingressClassName`
```

### DIFF
--- a/charts/__fixtures__/gardener-dashboard.js
+++ b/charts/__fixtures__/gardener-dashboard.js
@@ -26,8 +26,7 @@ const defaults = {
       ingress: {
         annotations: {
           'nginx.ingress.kubernetes.io/ssl-redirect': 'true',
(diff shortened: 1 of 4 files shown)


holgerkoser (approved):
/lgtm

Issue #1418, opened 2023-02-07
[vue3] Upgrade to Vue 3


grolu:
**What would you like to be added**:
We need to upgrade our project to Vue 3. As consequence thereof we need to upgrade Vuetify to version 3 / or switch to a diffrent UI library.
We should not only make our code work but also adapt it to the new features like composition api etc.

**Why is this needed**:
Vue 2 will reach End of Life (EOL) on December 31st, 2023

**Current Status**
- [x] https://github.com/gardener/dashboard/issues/1484
- [x] https://github.com/gardener/dashboard/issues/1483
- [x] https://github.com/gardener/dashboard/issues/1486
- [x] https://github.com/gardener/dashboard/issues/1487
- [x] https://github.com/gardener/dashboard/issues/1488
- [x] https://github.com/gardener/dashboard/issues/1485

PR #1513, opened 2023-08-16
Add ability to retreive adminkubeconfig to dashboard

Wieneo:
For clusters with K8s 1.27+, users can't access their shoots via the dashboard, if they don't want to use gardenctl.

I'm excited for your feedback! :)

**Release note**:
```feature user
Users can now request a time limited kubeconfig from within the dashboard
```


gardener-robot-ci-1:
Thank you @Wieneo for your contribution. Before I can start building your PR, a member of the organization must set the required label(s) {'reviewed/ok-to-test'}. Once started, you can check the build status in the PR checks section below.

petersutter:
Hi @Wieneo, thank you for your contribution.

At the moment, we are primarily focusing on the vue3 migration #1418 (`vue3` branch) and trying to minimize changes on the `master` branch. This is to ensure that we don't duplicate efforts. We're in the final stages of the vue3 migration and are planning to merge this branch into `master` by next week.

This implies that your PR would also need to be adjusted accordingly and based upon the `vue3` branch.

Some more feedback:

- Your current implementation requires the user to configure the lifetime using the settings button first. Then, upon pressing save, the kubeconfig is fetched, and the download, view, and copy buttons appear. This is not what I would expect as a user but probably comes because of a current limitation of the `CopyBtn`, right? It requires the text to be present beforehand. With the new browser API [navigator.clipboard.writeText](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/writeText), it's possible to fetch data asynchronously and then write it into the clipboard, however the CopyBtn does not support this yet. Adapting the `CopyBtn` component to support async text would enable us to have all buttons (download, view, copy, settings) visible from the start. The user can then directly download/view/copy the kubeconfig without setting a lifetime beforehand. A default lifetime (lowest value) is assumed.
- Confirming the name of the shoot cluster when changing the lifetime is not required. We usually do this for "dangerous" operations or operations that can't be reverted easily, e.g. worker settings changes, shoot deletion etc.
- We can also display the current configured lifetime to the user in the description, similar to how we do it for the maintenance time window (Maintenance time window starts in x hours) in the Lifecycle card. For example, Request kubeconfig valid for 30 minutes.
- Now there are three options available for the user to download a kubeconfig. We should give the user hints on which options are recommended and which not and explain the reason to help the user make the right decision. E.g. maybe have a `recommended` / `not recommended` chip with a tooltip / popover explaining the reason.
- Also it should be possible to turn off the feature globally, in case a landscape owner does not want to allow this option.
- The maximum lifetime that a user can set for the kubeconfig depends on how gardener is configured, see `--shoot-admin-kubeconfig-max-expiration` https://github.com/gardener/gardener/blob/master/docs/proposals/16-adminkubeconfig-subresource.md. This means that the lifetime values have to be configurable.
- Please split https://github.com/gardener/dashboard/blob/8b5858df22329300812fb5a17ecdbdb80aac5daa/frontend/src/components/AdminKubeConfigRequest.vue#L43 into value (e.g. 1800) and title (e.g. 30 minutes) so that the value (lifetime seconds) does not show up on the selection. The title can be calculated from the value, which makes also the configuration easier.
- Rename `getShootAdminKubeconfig` to `createAdminKubeconfig` to be consistent with how it is called in the client. 
- Use post instead of put, also https://github.com/gardener/dashboard/blob/8b5858df22329300812fb5a17ecdbdb80aac5daa/frontend/src/store/index.js#L999 should be `post` and not `put`.

holgerkoser:
> * Use post instead of put, also https://github.com/gardener/dashboard/blob/8b5858df22329300812fb5a17ecdbdb80aac5daa/frontend/src/store/index.js#L999
>    should be `post` and not `put`.

**[HTTP PUT](https://www.rfc-editor.org/rfc/rfc9110.html#name-put)**
PUT puts a file or resource at a specific URI, and exactly at that URI. If there's already a file or resource at that URI, PUT replaces that file or resource. If there is no file or resource there, PUT creates one. PUT is idempotent!

**[HTTP POST](https://www.rfc-editor.org/rfc/rfc9110.html#name-post)**
POST sends data to a specific URI and expects the resource at that URI to handle the request. The web server at this point can determine what to do with the data in the context of the specified resource.

Wieneo:
Hi!
Thank you very much for your feedback @petersutter!

I've tried to implement your feedback as best as I can:
- The feature can now be switched on or off
- The maximum expiration seconds can be configured
- The dropdown now only displays the "human readable" versions of the duration ("30m", "1h", etc...)
- Admin kubeconfig now get's fetched async as the user clicks on download / view
- The currently configured lifetime (default is 30m) is displayed to the user
- The dialog doesn't need to be confirmed via shoot name anymore
- Renamed a couple functions and use POST instead of PUT

Currently the CopyBtn is missing as I didn't get it to work with the async data fetching.
I will try to adapt this PR to vue3 after you merge the "vue3" branch into main :)

frontend/src/components/ShootDetails/ShootKubeconfig.vue
```diff
@@ -112,23 +136,85 @@ export default {
...
+        } catch (err) {
+          const errorMessage = 'Could not request admin kubeconfig'
+          const errorDetails = errorDetailsFromError(err)
+          console.error(errorMessage, errorDetails.errorCode, errorDetails.detailedMessage, err)
+        }
+      }
+    },
+    async toggleKubeconfig () {
+      if (this.isAdminKubeconfigLoginType) {
+        await this.createCachedAdminKubeconfig()
```
petersutter, line 187: - button should display as loading/progressing when pressing the view button while the kubeconfig is being fetched
- Only when visible it is required to fetch the kubeconfig
- Do not cache the kubeconfig. Just fetch a new one. E.g. User could leave the page open for some time and click the download button again but would get possibly an expired kubeconfig

frontend/src/components/ShootDetails/ShootKubeconfig.vue
```diff
@@ -88,17 +100,29 @@ export default {
   },
   data () {
     return {
-      kubeconfigExpansionPanel: false
+      kubeconfigExpansionPanel: false,
+      adminKubeConfigExpiration: '30m', // Default 30 minutes
+      adminKubeConfig: undefined
```
petersutter, line 105: remove this data. we do not want to cache it (as explained below)

frontend/src/utils/api.js
```diff
@@ -198,6 +198,12 @@ export function updateShootSeedName ({ namespace, name, data }) {
   return updateResource(`/api/namespaces/${namespace}/shoots/${name}/spec/seedname`, data)
 }
 
+export function createAdminKubeconfig ({ namespace, name, data }) {
+  namespace = encodeURIComponent(namespace)
+  name = encodeURIComponent(name)
+  return callResourceMethod(`/api/namespaces/${namespace}/shoots/${name}/adminkubeconfig`, data)
```
petersutter, line 204:
```suggestion
  return createResource(`/api/namespaces/${namespace}/shoots/${name}/adminkubeconfig`, data)
```

charts/gardener-dashboard/values.yaml
```diff
@@ -199,6 +199,10 @@ global:
       features:
         terminalEnabled: false
         projectTerminalShortcutsEnabled: false
+
+      adminKubeconfig:
```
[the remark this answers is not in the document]
holgerkoser: Yes. I think `shootAdminKubeconfig` is better

backend/lib/routes/shoots.js
```diff
@@ -254,3 +254,17 @@ router.route('/:name/spec/seedName')
...
+
+router.route('/:name/adminkubeconfig')
+  .all(metricsMiddleware)
+  .post(async (req, res, next) => {
+    try {
+      const user = req.user
+      const namespace = req.params.namespace
+      const name = req.params.name
+      const expirationSeconds = req.body.expirationSeconds
+      res.send(await shoots.createAdminKubeconfig({ user, namespace, name, expirationSeconds }))
```
holgerkoser, line 266: pass the complete body as body to the `createAdminKubeconfig` service function.

charts/gardener-dashboard/values.yaml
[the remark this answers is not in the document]
Wieneo: I renamed the key and added a chart test :)

backend/lib/config/gardener.js
```diff
@@ -122,6 +122,11 @@ module.exports = {
       _.unset(config, 'frontend.ticket')
     }
 
+    if (config.adminKubeconfig) {
```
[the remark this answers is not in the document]
Wieneo: Made it consistent

backend/lib/routes/shoots.js
```diff
@@ -254,3 +254,17 @@ router.route('/:name/spec/seedName')
       next(err)
     }
   })
+
+router.route('/:name/adminkubeconfig')
```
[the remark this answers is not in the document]
Wieneo: Added a condition

frontend/src/components/AdminKubeConfigRequest.vue
```diff
@@ -0,0 +1,66 @@
...
+      const confirmed = await this.$refs.actionDialog.waitForDialogClosed()
+      if (confirmed) {
+        await this.updateKubeconfigExpiration()
+      }
+    },
+    async updateKubeconfigExpiration () {
+      this.$emit('kubeconfigExpirationUpdate', this.kubeconfigExpiration)
+    },
+    reset () {
+      if (!this.kubeconfigExpiration) {
```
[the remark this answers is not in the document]
Wieneo: Reset is now always triggered

backend/lib/routes/shoots.js
Wieneo: Changed it

frontend/src/components/ShootDetails/ShootKubeconfig.vue
Wieneo: The current kubeconfig needs to be cached for the computed "kubeconfig" property though or am I missing something?
We can't fetch the admin kubeconfig async from within the computed property.

I changed the behaviour, that a new kubeconfig is always fetched.

frontend/src/store/index.js
```diff
@@ -995,6 +995,9 @@ const getters = {
   canPatchShootsBinding (state) {
     return canI(state.subjectRules, 'patch', 'core.gardener.cloud', 'shoots/binding')
   },
+  canCreateKubeconfig (state) {
+    return canI(state.subjectRules, 'post', 'core.gardener.cloud', 'shoots/adminkubeconfig')
```
holgerkoser, line 999:
```suggestion
    return canI(state.subjectRules, 'create', 'core.gardener.cloud', 'shoots/adminkubeconfig')
```

image

petersutter: you are right, my bad for suggesting `post`

frontend/src/utils/api.js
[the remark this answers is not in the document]
Wieneo: Fixed :)

frontend/src/components/ShootDetails/ShootKubeconfig.vue
```diff
@@ -112,23 +136,85 @@ export default {
...
+      const defaultExpirations = ['30m', '1h', '3h', '6h', '12h', '1d', '3d', '7d']
+      let filteredExpirations = defaultExpirations
+
+      if (this.cfg.adminKubeconfig.maxExpirationSeconds) {
+        filteredExpirations = []
+        defaultExpirations.forEach((val) => {
+          if (this.adminKubeConfigExpirationInSeconds(val) <= this.cfg.adminKubeconfig.maxExpirationSeconds) {
+            filteredExpirations.push(val)
+          }
+        })
```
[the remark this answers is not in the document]
Wieneo: Implemented

frontend/src/components/ShootDetails/ShootKubeconfig.vue
```diff
@@ -112,23 +136,85 @@ export default {
...
       return this.type === 'gardenlogin'
     },
+    isStaticKubeconfigLoginType () {
+      return this.type === 'token'
+    },
+    isAdminKubeconfigLoginType () {
+      return this.type === 'adminkubeconfig'
+    },
+    possibleAdminKubeconfigExpirationSettings () {
+      const defaultExpirations = ['30m', '1h', '3h', '6h', '12h', '1d', '3d', '7d']
```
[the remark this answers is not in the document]
Wieneo: The gardener parameter --shoot-admin-kubeconfig-max-expiration requires a minimum of 1h. That's why I thought 30m as the smallest expiration setting would be sufficient.
Maybe we should introduce a validation within the HELM chart and config, to ensure maxExpirationSeconds is at least 1h too?

I personally like the approach of the dashboard defining, what expiration settings the user can select. This prevents operators / users from creating kubeconfigs with validity of 1year+.

I unfortunately didn't get the wildcard-select component to work with an array of objects (with properties: item (string), value (int)).
I got the error: nvalid prop: type check failed for prop "value". Expected String, got Object

petersutter:
@Wieneo meanwhile we have merged the `vue3` branch, you can now adapt your PR accordingly

backend/lib/config/gardener.js
```diff
@@ -126,6 +126,11 @@ module.exports = {
       _.unset(config, 'frontend.ticket')
     }
 
+    if (config.frontend.shootAdminKubeconfig) {
+      _.set(config, 'frontend.shootAdminKubeconfig.enabled', config.frontend.shootAdminKubeconfig.enabled)
+      _.set(config, 'frontend.shootAdminKubeconfig.maxExpirationSeconds', config.frontend.shootAdminKubeconfig.maxExpirationSeconds)
+    }
+
```
holgerkoser, line 133: Why do we need this?

backend/lib/routes/shoots.js
```diff
@@ -254,3 +255,19 @@ router.route('/:name/spec/seedName')
       next(err)
     }
   })
+
+if (config.frontend.shootAdminKubeconfig && config.frontend.shootAdminKubeconfig.enabled) {
```
holgerkoser, line 259:
```suggestion
if (config.frontend.shootAdminKubeconfig?.enabled) {
```

backend/lib/services/shoots.js
```diff
@@ -159,6 +160,22 @@ exports.replaceSeedName = async function ({ user, namespace, name, body }) {
...
+    kind,
+    apiVersion,
+    spec: {
+      expirationSeconds: body.expirationSeconds
+    }
+  }
+
+  const adminKubeConfigRequest = await client['core.gardener.cloud'].shoots.createAdminKubeconfigRequest(namespace, name, payload)
+  const kubeconfig = Buffer.from(adminKubeConfigRequest.status.kubeconfig, 'base64').toString('utf8')
+  return kubeconfig
```
holgerkoser, line 176:
```suggestion
  const { status } = await client['core.gardener.cloud'].shoots.createAdminKubeconfigRequest(namespace, name, payload)
  return utils.decodeBase64(status.kubeconfig) 
```

frontend/src/components/AdminKubeConfigRequest.vue
```diff
@@ -0,0 +1,66 @@
...
+  props: {
+    expirations: {
+      type: Array
+    }
+  },
+  props: {
+    expirations: {
+      type: Array
+    }
+  },
```
holgerkoser, line 52: Two time the same thing

frontend/src/components/AdminKubeConfigRequest.vue
```diff
@@ -0,0 +1,66 @@
...
+  </action-button-dialog>
+</template>
+
+<script>
+import ActionButtonDialog from '@/components/dialogs/ActionButtonDialog'
+import WildcardSelect from '@/components/WildcardSelect'
+import shootItem from '@/mixins/shootItem'
+
+export default {
+  name: 'admin-kubeconfig-request',
```
holgerkoser, line 30:
```suggestion
```

charts/gardener-dashboard/values.yaml
```diff
@@ -204,6 +204,10 @@ global:
       features:
         terminalEnabled: false
         projectTerminalShortcutsEnabled: false
+
+      shootAdminKubeconfig:
+        enabled: true
```
holgerkoser, line 209: I don't think it should be enabled by default. What do you think @petersutter?

frontend/src/components/AdminKubeConfigRequest.vue
holgerkoser: The component is not used anywhere in the frontend.

petersutter:
there are some files missing, can you double check @Wieneo ?

Wieneo:
@petersutter 
Sorry, I didn't get around to finishing the migration to vue3 yesterday. I pushed the new working state.

frontend/src/composables/useApi/api.js
```diff
@@ -193,6 +193,12 @@ export function updateShootSeedName ({ namespace, name, data }) {
   return updateResource(`/api/namespaces/${namespace}/shoots/${name}/spec/seedname`, data)
 }
 
+export function createAdminKubeconfig ({ namespace, name, data }) {
```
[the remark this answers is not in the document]
Wieneo: Fixed :)

frontend/src/components/AdminKubeConfigRequest.vue
Wieneo: Fixed :)

frontend/src/components/AdminKubeConfigRequest.vue
Wieneo: Fixed :)

petersutter:
Hi @Wieneo, in case you are still working on it, please mark this PR as draft until you are ready for review, thanks

frontend/src/components/ShootDetails/ShootKubeconfig.vue
Wieneo: How would I mark a button as loading / progressing?
I implemented the other points.

backend/lib/config/gardener.js
Wieneo: I thought this was needed for /api/config to serve the shootAdminKubeconfig settings. I removed it.

Wieneo:
Hi @petersutter,
will remember that for next time.

The PR is ready for review :)
I resolved all conversations that I fixed.

petersutter (changes requested):
Also make sure to run [yarn lint](https://github.com/gardener/dashboard/blob/7521759482f1a12d5e6a484ba95b0013e584bc36/frontend/package.json#L28C42-L28C42)

frontend/src/components/ShootDetails/ShootKubeconfig.vue
petersutter: see v-progress-circular. However I noticed that we do not do the same for the kubeconfigs that can be downloaded on the member page. If we add it then it should be added to the member page as well (but not in this PR). I resolve this conversation..

frontend/src/components/ShootDetails/GShootAccessCard.vue
```diff
@@ -250,6 +255,9 @@ export default {
     hasDashboardTokenAuth () {
       return get(this.shootItem, 'spec.addons.kubernetesDashboard.authenticationMode', 'basic') === 'token'
     },
+    hasAdminKubeconfigEnabled () {
+      return get(this.cfg, 'shootAdminKubeconfig.enabled', false)
```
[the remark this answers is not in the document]
Wieneo: I don't know how I missed that, fixed it.
Admin Kubeconfig now only gets displayed within the access card, if its enabled within the config.

frontend/src/components/ShootDetails/GShootKubeconfig.vue
```diff
@@ -141,6 +174,23 @@ export default {
...
+    isStaticKubeconfigType () {
+      return this.type === 'token'
+    },
+    isAdminKubeconfigType () {
+      return this.type === 'adminkubeconfig'
+    },
+    possibleAdminKubeconfigExpirationSettings () {
+      let expirations = ['30m', '1h', '3h', '6h', '12h', '1d', '3d', '7d']
+
+      if (this.shootAdminKubeconfig.maxExpirationSeconds) {
```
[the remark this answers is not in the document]
Wieneo: Fixed it :)
This function also doesn't get called anymore, if the adminkubeconfig is disabled.

frontend/src/components/GAdminKubeConfigRequest.vue
```diff
@@ -0,0 +1,61 @@
...
+    expirations: {
+      type: Array
+    }
+  },
+  methods: {
+    async onConfigurationDialogOpened () {
+      this.reset()
+      const confirmed = await this.$refs.actionDialog.waitForDialogClosed()
+      if (confirmed) {
+        this.$emit('expirationUpdate', this.expiration)
```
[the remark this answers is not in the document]
Wieneo: Fixed

charts/gardener-dashboard/values.yaml
Wieneo: Changed it

frontend/src/components/GAdminKubeConfigRequest.vue
```diff
@@ -0,0 +1,61 @@
...
+
+<template>
+  <g-action-button-dialog
+    :shoot-item="shootItem"
+    @dialog-opened="onConfigurationDialogOpened"
+    ref="actionDialog"
+    width="500"
+    caption="Configure Kubeconfig Lifetime">
+    <template v-slot:actionComponent>
+      <g-wildcard-select
```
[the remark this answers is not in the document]
Wieneo: I tried this but v-select seems to populate the v-model only with the value and not the whole object.
Do you know what I did wrong?

frontend/src/components/ShootDetails/GShootKubeconfig.vue
```diff
@@ -141,6 +174,23 @@ export default {
...
       return this.type === 'gardenlogin'
     },
+    isStaticKubeconfigType () {
+      return this.type === 'token'
+    },
+    isAdminKubeconfigType () {
+      return this.type === 'adminkubeconfig'
+    },
+    possibleAdminKubeconfigExpirationSettings () {
+      let expirations = ['30m', '1h', '3h', '6h', '12h', '1d', '3d', '7d']
```
petersutter, line 184: also add 10 minute expiration. Default can stay at 30minutes

frontend/src/components/ShootDetails/GShootKubeconfig.vue
```diff
@@ -45,26 +45,40 @@ SPDX-License-Identifier: Apache-2.0
...
+          @click.stop="toggleKubeconfig"
         />
       </template>
 
       <g-static-token-kubeconfig-configuration
-        v-if="!isGardenloginType"
+        v-if="isStaticKubeconfigType"
         :shoot-item="shootItem"
       />
+      <g-admin-kube-config-request
```
petersutter, line 76: pass `adminKubeConfigExpiration` to component (so that defaulting works).

frontend/src/components/ShootDetails/GShootKubeconfig.vue
```diff
@@ -45,26 +45,40 @@ SPDX-License-Identifier: Apache-2.0
         </span>
       </div>
     </g-list-item-content>
+    <g-list-item-content v-if="isAdminKubeconfigType">
+      Admin Kubeconfig
+      <div class="text-body-2">
+        <span>
+          Request a kubeconfig valid for {{adminKubeConfigExpiration}}
```
petersutter, line 52: I can't remember if you had implemented previously before switching to vue3, but the selected expiration value should be stored in the local storage

petersutter (changes requested):

frontend/src/components/GAdminKubeConfigRequest.vue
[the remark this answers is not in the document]
petersutter: Yes that's correct. In which case do you need the whole item? As the values should be unique, you can use lodash find to find it from the `possibleAdminKubeconfigExpirationSettings`

```html
        <span>
          Request a kubeconfig valid for {{ adminKubeConfigExpirationTitle }}
        </span>
```

```js
    adminKubeConfigExpirationTitle () {
      const expiration = find(this.possibleAdminKubeconfigExpirationSettings, this.adminKubeConfigExpiration)
      return expiration?.title
    },
    possibleAdminKubeconfigExpirationSettings () {
      let expirations = [
        { value: 600, title: '10m' },
        { value: 1800, title: '30m' },
        { value: 3600, title: '1h' },
        { value: 10800, title: '3h' },
        { value: 21600, title: '6h' },
        { value: 43200, title: '12h' },
        { value: 86400, title: '1d' },
        { value: 259200, title: '3d' },
        { value: 604800, title: '7d' },
      ]

      if (this.shootAdminKubeconfig?.maxExpirationSeconds) {
        expirations = filter(expirations, ({ value }) => {
          return value <= this.shootAdminKubeconfig.maxExpirationSeconds
        })
      }

      return expirations
    },
```

petersutter:
@Wieneo FYI with https://github.com/gardener/dashboard/pull/1546 lazy loading support of copy button is now implemented

grolu:
Hi @Wieneo, as @petersutter already mentioned, lazy loading support is now supported by `GCopyBtn`. Could you please have a look and adapt this PR, so that we can merge it. We would like to integrate this Feature in one of our upcoming releases.

grolu:
closed an favor of #1666

PR #1514, opened 2023-08-16
Support GitHub app authentication

petersutter:
This PR adds support for github app authentication. Also an `authStrategy` has to be set as setting only an `auth` object or a basic authentication string is deprecated since long

> Setting auth to an object or a basic authentication string is now deprecated and will be removed in v17 ([4573ee2](https://github.com/octokit/rest.js/commit/4573ee22d2e4c7ee0f291eeb0bf78b46ed438e95))

Source: octokit release [v16.38.0](https://github.com/octokit/rest.js/releases/tag/v16.38.0)

Example `values.yaml` for github app authentication
```yaml
global:
  dashboard:
    gitHub:
      authentication:
        # authenticate as github app installation. The github app requires the permission to read and write issues and has to be installed in the respective repository
        appId: 1
        clientId: lv1.1234567890abcdef
        clientSecret: 1234567890abcdef1234567890abcdef12345678
        installationId: 123
        privateKey: |
          -----BEGIN RSA PRIVATE KEY-----
          ...
          -----END RSA PRIVATE KEY-----
```

Example `values.yaml` for token authentication
```yaml
global:
  dashboard:
    gitHub:
      authentication:
        token: dummytoken
```

Fixes #

**Release note**:

```feature operator
Github app authentication is now supported (in addition to token authentication) for the ticket feature
```


charts/__tests__/gardener-dashboard/runtime/dashboard/configmap.spec.js
```diff
@@ -268,23 +267,53 @@ describe('gardener-dashboard', function () {
...
+
+        it('should render the template', async function () {
+          const documents = await renderTemplates(templates, values)
+          expect(documents).toHaveLength(2)
+          const [configMap, githubSecret] = documents
+          expect(configMap.metadata.name).toBe(name)
+          expect(githubSecret.metadata.name).toBe('gardener-dashboard-github')
+          const config = yaml.load(configMap.data['config.yaml'])
+          expect(pick(config, ['frontend.ticket', 'gitHub'])).toMatchSnapshot()
+          expect(githubSecret).toMatchSnapshot()
```
petersutter, line 292: only authentication + decode base64

backend/lib/github/octokit.js
```diff
@@ -31,37 +33,32 @@ class OctokitLog {
...
-      username,
-      password
-    }
-  }
-  if (clientId && clientSecret) {
     return {
-      clientId,
-      clientSecret
+      authStrategy: ({ token }) => createTokenAuth(token),
+      auth: {
```
petersutter, line 47: return { auth: token } instead

petersutter: ✅

charts/__tests__/gardener-dashboard/runtime/dashboard/configmap.spec.js
petersutter: ✅

PR #1515, opened 2023-08-18
Clarify misleading wording on potential updates during maintenance

dguendisch:
Clarifies a misleading wording on potential updates during maintenance.
When hitting the maintenance button, a dialog currently reads "The following updates will be performed...".
However the dashboard has no idea that such updates will definitely be performed (and in most cases they won't), hence change it to "might be performed".

Fixes #

**Release note**:

```bugfix user

```


petersutter (approved):
/lgtm thanks

PR #1516, opened 2023-08-18
Do not close tickets on shoot deletion

petersutter:
This PR removes the feature that automatically closes tickets when a shoot is deleted.
This should be handled by a different component, like a robot.

Fixes #

**Release note**:

```breaking operator
Tickets are not longer automatically closed by the dashboard when a shoot is deleted. This should be handled by a different component, like a robot.
```


petersutter:
cc @neo-liang-sap @etiennnr

PR #1529, opened 2023-08-22
Reset filter on Secrets, Members and Shoot list page

### LINKED ISSUE #1524: Filter value does not reset when navigating
**What happened**:
When navigating to a different project, the search / filter value does not reset. If the target project has less data the filter input text field might even not be displayed even though the filter value is still applied.

- Navigate to member page
- Select project with more than 3 members
- Search for a member
- Change to project with 3 members or less
- Filter input text field is hidden, filter value is still applied
- Reload page, filter value gets resetted

**Environment**:

petersutter:
Reset filter on Secrets, Members and Shoot list page when switching the Project

Fixes #1524

**Release note**:

```bugfix user
Fixed an issue where the filter on the `Secrets`, `Members` and `Clusters` page was not reset when switching the Project
```

### DIFF
--- a/frontend/src/views/GMembers.vue
+++ b/frontend/src/views/GMembers.vue
@@ -289,6 +289,7 @@ import {
   computed,
   markRaw,
   inject,
+  watch,
 } from 'vue'
 import { storeToRefs } from 'pinia'
 import { useRouter } from 'vue-router'
@@ -554,6 +555,17 @@ const serviceAccountTableHeaders = computed(() => {
 const visibleServiceAccountTableHeaders = computed(() => {
   return filter(serviceAccountTableHeaders.value, ['selected', true])
 })
+
+watch(namespace, () => {
+  reset()
+})
+
+function reset () {
+  userFilter.value = ''
+  serviceAccountFilter.value = ''
+  userPage.value = 1
+  serviceAccountPage.value = 1
+}
 
 function setKubeconfigError (err) {
   appStore.alert = {
--- a/frontend/src/views/GMembers.vue
+++ b/frontend/src/views/GMembers.vue
@@ -373,11 +373,11 @@ const userPage = ref(1)
 const serviceAccountPage = ref(1)
 
 const {
-  namespace,
   project,
   projectList,
 } = storeToRefs(projectStore)
 const {
+  namespace,
   canManageMembers,
   canManageServiceAccountMembers,
   canCreateServiceAccounts,
--- a/frontend/src/views/GSecrets.vue
+++ b/frontend/src/views/GSecrets.vue
@@ -332,7 +332,10 @@ export default {
         'infrastructureSecretList',
         'dnsSecretList',
       ]),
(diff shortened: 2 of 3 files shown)


grolu:
What about the shoot list filter? When I switch the project, the filter value remains in the search field, however the value is no longer applied (list is not filtered). I think we should reset the filter value in this case, too.

petersutter:
right @grolu, fixed it

grolu (approved):
/lgtm

frontend/src/views/GMembers.vue
```diff
@@ -555,6 +556,17 @@ const visibleServiceAccountTableHeaders = computed(() => {
   return filter(serviceAccountTableHeaders.value, ['selected', true])
 })
 
+watch(namespace, () => {
```
holgerkoser, line 559: Not directly related to this PR but the `namespace` should be imported from the `authzStore`.

frontend/src/views/GSecrets.vue
```diff
@@ -327,12 +327,14 @@ export default {
...
-        'dnsSecretList',
-      ]),
-    ...mapGetters(useAuthzStore, ['canCreateSecrets']),
+    ...mapGetters(useSecretStore, [
+      'infrastructureSecretList',
+      'dnsSecretList',
+    ]),
+    ...mapGetters(useAuthzStore, [
+      'canCreateSecrets',
+      'namespace',
```
holgerkoser, line 336:
```suggestion
      'namespace',
      'canCreateSecrets',
```

frontend/src/views/GShootList.vue
```diff
@@ -281,20 +282,21 @@ export default {
...
     this.cachedItems = this.shootList.slice(0)
-    this.shootSearch = null
+    this.resetShootSearch()
     this.focusModeInternal = false
     next()
   },
   data () {
     return {
       shootSearch: '',
+      delayedShootSearch: '',
```
holgerkoser, line 299: Why do we need two properties? If we need it the name should be `debouncedShootSearch` to make it more clear what is is.

frontend/src/views/GShootList.vue
```diff
@@ -73,6 +73,7 @@ SPDX-License-Identifier: Apache-2.0
             <template #activator="{ props }">
               <v-text-field
                 v-bind="props"
+                v-model="shootSearch"
```
holgerkoser, line 76: Do not use `v-model` and `@update:model-value` in combination

frontend/src/views/GSecrets.vue
petersutter: what is this change about?

petersutter: ✅

frontend/src/views/GShootList.vue
petersutter: ✅  as discussed

holgerkoser (approved):
/lgtm

PR #1532, opened 2023-08-28
Use Custom Serialization in `useLocalStorage` composable

### LINKED ISSUE #1520: Uncaught TypeError: Cannot read properties of undefined
**What happened**:
When reading data from local storage an exceprion can occur if the data is unexpected.

Change value to `null` for local storage `global/color-scheme` key

**Environment**:

holgerkoser:
This PR moves all usages of the vueuse composable `useLocalStorage` to the pinia store names `localStorageStore` and attaches a custom serialization implementation that is error tolerant. Some users can have old incompatible localStorage values for certain key. This could cause uncaught exceptions

Fixes #1520 

**Release note**:

```bugfix user
Fixed deserialization of incompatible or invalid localStorage values. Handle exceptions and fallback to default or previous values. 
```


frontend/src/components/GAlertBanner.vue
```diff
@@ -57,6 +58,8 @@ const props = defineProps({
 })
 
 const { type, color } = toRefs(props)
+const localStorageStore = useLocalStorageStore()
+const hiddenMessages = toRef(localStorageStore, 'hiddenMessages')
```
petersutter, line 62:
```js
const permanentlyHidden = computed({
  get () {
    if (props.identifier) {
      return hiddenMessages.value[props.identifier] === true
    }
    return false
  },
  set (value) {
    if (props.identifier) {
      hiddenMessages.value[props.identifier] = value
    }
  },
})
```
```
GAlertBanner.vue:71 Uncaught (in promise) TypeError: Cannot read properties of null (reading 'cfg.4be0e4219e7ecb2bcc11fc2b18b25f37')
    at ReactiveEffect.get [as fn] (GAlertBanner.vue:71:34)
    at ReactiveEffect.run (reactivity.esm-bundler.js:178:19)
    at get value [as value] (reactivity.esm-bundler.js:1147:33)
    at ReactiveEffect.get [as fn] (GAlertBanner.vue:83:51)
    at ReactiveEffect.run (reactivity.esm-bundler.js:178:19)
    at get value [as value] (reactivity.esm-bundler.js:1147:33)
    at unref (reactivity.esm-bundler.js:1026:29)
    at Object.get (reactivity.esm-bundler.js:1032:35)
    at Proxy._sfc_render (GAlertBanner.vue:9:14)
    at renderComponentRoot (runtime-core.esm-bundler.js:816:16)
```

as the default for hiddenMessages  is gone, hiddenMessages  is null. 
You can reproduce it by setting an alert in your frontend config, e.g.

```yaml
frontend:
  alert:
    type: warning
    message: foo
```

frontend/src/composables/useTerminalSplitpanes.js
```diff
@@ -183,72 +153,57 @@ export const useTerminalSplitpanes = ({ name, namespace, target }) => {
...
+    const onySlotItemsInTree = every(symbolTree.ids(), id => includes(slotItemUUIds.value, id))
+    localStorageStore.terminalSplitpaneTree = onySlotItemsInTree || symbolTree.isEmpty()
+      ? null // clear value
+      : splitpaneTree.value
   }
 
   function targetIdOrDefault (targetId) {
     if (!targetId) {
-      targetId = focusedElementId.value // TODO check if it still works
+      targetId = appStore.focusedElementId // TODO check if it still works
```
petersutter, line 167:
```suggestion
      targetId = appStore.focusedElementId
```
we can remove this comment

frontend/src/views/GMembers.vue
```diff
@@ -772,11 +774,11 @@ function sortItems (items, sortByArr, secondSortCriteria) {
 }
 
 function setSelectedHeaderUserAccount (header) {
-  userAccountSelectedColumns.value[header.key] = !header.selected
+  userSelectedColumns.value[header.key] = !header.selected
```
petersutter, line 777: Column selection on the Members page for both tables does not work anymore

frontend/src/views/GShootItem.vue
```diff
@@ -45,27 +41,21 @@ export default {
...
-      terminalCoordinates: this.terminalCoordinates,
-      defaultTarget: this.defaultTarget,
-      splitpanesState: this.state,
-      moveTo: this.moveTo,
-      add: this.add,
-      setSelections: this.setSelections,
-      removeWithId: this.removeWithId,
-      leavePage: this.leavePage,
+      terminalSplitpanes,
+      ...terminalSplitpanes,
```
petersutter, line 47: was this done temporarily and in the future we only want to provide `terminalSplitpanes`

frontend/src/composables/useTerminalSplitpanes.js
holgerkoser: ✔️

frontend/src/views/GShootItem.vue
holgerkoser: Removed provide of `terminalSplitpanes` since it is not use at the moment.

frontend/src/views/GMembers.vue
holgerkoser: ✔️

frontend/src/components/GAlertBanner.vue
holgerkoser: ✔️

PR #1536, opened 2023-08-29
Bug/fix time input safari

### LINKED ISSUE #1427: Hibernation: wake up in Safari not removable
**What happened**:
I tried to configure a cluster in **Safari** to not having a hibernation wake up configured, but an hibernation, but it was always using 12:30 as a default, when I tried to remove it. After looking at the same cluster in Chrome it looked different, there was a time selector pull down and also the option to remove the wake up schedule.

removing wake up schedule should also work in Safari 

login with Safari, try to maintain a wake up schedule for hiobernation, see you can't remove it.

**Environment**:
canary

**Hibernation config in Chrome:** Screenshot 2023-02-16 at 08 38 47

**Hibernation schedule in Safari:** Screenshot 2023-02-16 at 08 39 27

grolu:
This PR fixes the Hibernation Schedule time input issue in Safari by falling back to a regular text input field.
This is required because Safari browser rendering of text fields with type 'time' is confusing
See also https://bugs.webkit.org/show_bug.cgi?id=233215
So we use type text for Safari and add hint & error handing to support user input

Screenshot 2023-08-29 at 16 01 23

Fixes #1427

**Release note**:

```bugfix user
Fixed hibernation schedule time input for Safari browser
```

### DIFF
--- a/frontend/src/components/GTimeTextField.vue
+++ b/frontend/src/components/GTimeTextField.vue
@@ -1,0 +1,87 @@
+<!--
+SPDX-FileCopyrightText: 2023 SAP SE or an SAP affiliate company and Gardener contributors
+
+SPDX-License-Identifier: Apache-2.0
+-->
+
+<template>
+  <v-text-field
+    v-model="timeValue"
+    :error-messages="errorMessages"
+    :hint="inputHint"
+    :type="textFieldType"
+    @change="v$.timeValue.$touch()"
+  />
+</template>
+
+<script setup>
+
+import { computed } from 'vue'
+import { useVuelidate } from '@vuelidate/core'
+import { useVModel } from '@vueuse/core'
+
+import { getValidationErrors } from '@/utils'
+
+const props = defineProps({
+  modelValue: {
+    type: String,
+  },
+  errorMessages: {
+    type: Array,
+  },
+})
+
+const emit = defineEmits([
+  'update:modelValue',
+])
+
+const timeValue = useVModel(props, 'modelValue', emit)
+
+const isSafari = computed(() => {
+  return /^((?!chrome|android).)*safari/i.test(navigator.userAgent)
+})
+
+const textFieldType = computed(() => {
+  /*
+  Safari browser rendering of text fields with type 'time' is confusing
+  See also https://bugs.webkit.org/show_bug.cgi?id=233215
+  So we use type text for Safari and add hint & error handing to support user input
+  */
+  return isSafari.value ? 'text' : 'time'
+})
+
+const rules = {
+  timeValue: {
+    timeSyntax: (value) => {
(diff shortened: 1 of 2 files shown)


frontend/src/components/ShootHibernation/GHibernationScheduleEvent.vue
```diff
@@ -30,31 +30,29 @@ SPDX-License-Identifier: Apache-2.0
...
             v-model="wakeUpTime"
             color="primary"
             label="Wake up at"
             :error-messages="getErrorMessages('wakeUpTime')"
-            type="time"
             clearable
             variant="underlined"
             @blur="touchIfNothingFocused"
-            @input="onInputWakeUpTime"
+            @update:model-value="onInputWakeUpTime"
```
petersutter, line 42: use `model-value` inststead of `v-model` when using `@update:model-value` for both occurrences

frontend/src/components/ShootHibernation/GHibernationScheduleEvent.vue
```diff
@@ -30,31 +30,29 @@ SPDX-License-Identifier: Apache-2.0
...
           />
         </v-col>
         <v-col cols="2">
-          <v-text-field
+          <g-time-text-field
```
petersutter, line 46: please also use this component for the configure maintenance dialog

frontend/src/components/ShootHibernation/GHibernationScheduleEvent.vue
grolu: ? which occurrence?

petersutter: <g-time-text-field /> component is used twice in this file (line 33 and line 46). Both use v-model

grolu: Ah okay now I think I understand what you mean. Changing `v-model` to `model-value` would require further changes as we require the two-way data binding in this case. Of course we could handle this in the `@update:model-value` event handler but then we should change this everywhere. Right now we use this event handler in lots of places together with v-model, mostly as replacement for `@input` events that now behave differently.

frontend/src/components/GTimeTextField.vue
```diff
@@ -0,0 +1,91 @@
...
+  },
+})
+
+const emit = defineEmits([
+  'update:modelValue',
+])
+
+const timeValue = useVModel(props, 'modelValue', emit)
+
+const isSafari = computed(() => {
```
holgerkoser, line 43: Try to detect the specific behavior you're trying to fix, instead of targeting it with `isSafari`?

holgerkoser: It seems not possible to check for the behavior!

PR #1542, opened 2023-09-01
Fix some navigation issues

### LINKED ISSUE #1522: Router Error loop when navigating from broken yaml
**What happened**:
When a `YAMLException` occurs, the dashboard runs into an endless error loop in the router while navigating

- Go to create cluster screen
- Switch to yaml tab
- change to something invalid
- navigate to a different page (e.g. secrets)

**Environment**:

holgerkoser:
TODO: write a detailed description on what has been changed and the issues which are fixed but still not completely understood.
 
Fixes #1522

**Release note**:

```bugfix user

```

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -67,7 +67,7 @@
     "url-template": "^3.1.0",
     "uuid": "^8.3.2",
     "vue": "^3.3.4",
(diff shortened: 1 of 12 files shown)


grolu (approved):
/lgtm

PR #1547, opened 2023-09-05
Fixed authnStore implementation

### LINKED ISSUE #1534: Autologin not working with "jwt expired" error
**What happened**:
The JWT expires very quickly and the user is redirected to the login screen with the corresponding error message even if autoLogin is enabled.

```sh
29 07:34:56.387 http : ::ffff:127.0.0.1 - - [29/Aug/2023:07:34:56 +0000] "GET /auth?redirectUrl=http%3A%2F%2Flocalhost%3A8080%2F HTTP/1.1" 302 724
2023-08-29 07:34:56.392 warn : Authentication failed: UnauthorizedError - jwt expired
2023-08-29 07:34:56.392 http : ::ffff:127.0.0.1 - - [29/Aug/2023:07:34:56 +0000] "POST /auth/token HTTP/1.1" 401 465
2023-08-29 07:34:56.406 http : ::ffff:127.0.0.1 - - [29/Aug/2023:07:34:56 +0000] "GET /auth/logout?error%5Bname%5D=UnauthorizedError&error%5Bmessage%5D=jwt+expired HTTP/1.1" 302 96
```

**Environment**:

holgerkoser:
The value of the user reference in the authnStore is only updated in the global `beforeEach` navigation guard. The connection status of the socket.io connection is driven by this store property. But on the other hand the request interceptor that ensures an valid JWT must check the current cookie value and not the store property if the token is expired or not. This has been mixed up during the migration of the UserManager class in `@/utils/auth.js` file. This PR uses the old implementation of the UserManager class and migrates it without changing the functionality.

The commit [24592dd](https://github.com/gardener/dashboard/pull/1547/commits/24592dde83a1a081871a448fe7d174eb62887045) fixes situation where the shootItem is not yet or not anymore populated during navigation to or from `ShootItemPlaceholder`. The intention of `required` was that the icon property must be specified when using the component but it can handle the situation the the value of the bound property is undefined.

Fixes #1534 

**Release note**:

```bugfix user

```


petersutter (approved):
/lgtm

grolu (approved):
/lgtm

PR #1551, opened 2023-09-11
Do not pin minor node version

holgerkoser:
Fix the ESM loader problem with import assertions by upgrading to latest yarn version. Revert temporary workaround to enforce the previous minor nodejs version.

Fixes #

**Release note**:

```bugfix user

```


grolu (approved):
/lgtm

PR #1552, opened 2023-09-11
handle `user === null` correctly

holgerkoser:
If there was no cookie the value of the user reference in the authStore was null. The leads to an uncaught exception in the browser.

Fixes #

**Release note**:

```bugfix user

```


petersutter (approved):
/lgtm

Issue #1445, opened 2023-03-21
Configure additional fields for cloud providers


grolu:
**What would you like to be added**:
With PR #1436 we added support for adding cloud providers via dashboard configuration.
However if a cloud provider requires additional input on the cluster creation page, it is currently not possible to configure additional fields.

***Implementation Plan
- [ ] Merge https://github.com/gardener/dashboard/pull/1553
- [x] Upgrade to Vue / Vuetify 3
- [ ] Refactor `NewShootInfrastructureDetails` component to store staging resource #975
- [x] add `generic` component for additional configurable fields

As first step support only plain input fields or maybe simple dropdowns

**Why is this needed**:
Expose require or additional configuration to avoid complicated yaml editing.

Issue #554, opened 2020-01-17
Dashboard Extensibility


grolu:
Allow to extend the dashboard with custom add-ons, like new cloud infrastructure types without modifying the Dashboard core.
This allows operators to customize and extend the dashboard without the need for a custom build which slows down the adoption of new dashboard version.

- [X] Customize appearance (Dashboard Theming) #953 #980
- [x] Custom Cloud Providers
  - [ ] Add additional cloud providers via configuration #1434 #1553
  - [ ] Add secrets page via configuration #1553
  - [x] Customize built-in vendors #2695
  - [ ] Add additional input fields to create cluster page via configuration #1445
- [x] Introduce basic branding functionality #1568

vlerenc:
@grolu Are more changes planned (just asking because all checkboxes are checked)?

grolu:
The PR ist still not reviewed. But hopefully we can merge soon... Other than that there are no more changes planned.

Issue #207, opened 2018-10-12
Sort Cloud Profiles


grolu:
The Garden Administrator needs an option to sort cloud profiles. This could be achieved by adding an annotation to all cloud profiles with an integer value. The annotation should be part of the landscape setup. This value could be used by the Gardener Dashboard (and other Tools / UIs) to sort infrastructures, as well as multiple Cloud Profiles within an infrastructure.

Imagine the following Cloud Profiles:

```
apiVersion: garden.sapcloud.io/v1beta1
kind: CloudProfile
metadata:
  name: aws
  annotations:
    garden.sapcloud.io/priority: 200
spec:
  aws:
...
```

```
apiVersion: garden.sapcloud.io/v1beta1
kind: CloudProfile
metadata:
  name: openstack-1
  annotations:
    garden.sapcloud.io/priority: 102
spec:
  openstack:
...
```

```
apiVersion: garden.sapcloud.io/v1beta1
kind: CloudProfile
metadata:
  name: openstack-2
  annotations:
    garden.sapcloud.io/priority: 101
spec:
  openstack:
...
```

Currently they would be ranked like this:

Infrastructures:
- aws
- openstack:
  - openstack-1
  - openstack-2

With the annotation, the dashboard would order the Cloud Profiles like this:

Infrastructures:
- openstack:
  - openstack-2
  - openstack-1
- aws

This feature helps to highlight / promote specific infrastructures or cloud profiles. It could also help reducing costs by ranking cheaper options first.

grolu:
And alternative way to solve this would be to allow to set preferred / default infrastructure and secret per project in the dashboard. This setting could be persisted as annotation on the project resource and would be shared with all members of a project.

grolu:
@vlerenc @petersutter @holgerkoser I'm not sure about this. I don't think that someone actively requested this feature?

vlerenc:
Hmm... wasn't there a planned application of that feature by @RaphaelVogel , but maybe we solved it differently? I can't recall anymore.

grolu:
https://github.com/gardener/dashboard/pull/1553
will solve this

PR #1434, opened 2023-03-01
Support generic cloud providers

grolu:
**What this PR does / why we need it**:
This PR has several improvements regarding unsupported cloud providers.

***If no (valid) cloud provider is configured that the dashboard supports, this will no longer result in an fatal error 500. Instead this will be rendered as an information in the dashboard***
Screenshot 2023-02-28 at 14 47 42
Screenshot 2023-02-28 at 14 48 04

***Added a generic secret dialog as well as fallback implementation for all required code parts. Start implementing support for a new cloud provider is now as simple as adding the name of the new provider to the list of supported providers***

***Added support for Gardener `local` infrastructure provider***
Screenshot 2023-03-01 at 10 48 36
Screenshot 2023-03-01 at 10 54 44

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Release note**:

```bugfix developer
If no (valid) cloud provider is configured that the dashboard supports, this will no longer result in an fatal error 500. Instead this will be rendered as an information in the dashboard
```
```feature developer
Added a generic secret dialog as well as fallback implementation for all required code parts. Start implementing support for a new cloud provider is now as simple as adding the name of the new provider to the list of supported providers (`supportedProviderTypes` in store/index.js)
```
```feature developer
Added support for `local` infrastructure provider. This allows to use the dashboard with local gardener installations. See also: [Deploying Gardener locally](https://github.com/gardener/gardener/blob/master/docs/deployment/getting_started_locally.md)
```

PR #1553, opened 2023-09-11
[DRAFT] Add cloud providers via configuration

grolu:
On top of PR #1434 this PR adds logic to define additional cloud providers via configuration (`customCloudProviders`).
It is possible to define a custom secret dialog including (simple) validation and help texts.
Current possible validations:
- required
- requiredIf (ref to other field) // TODO add example
- regex validation
- isJSON
- isYAML

A custom cloud provider icon can be defined using an image resource mounted via asset configuration.
This way it is also possible to overwrite icons as well as names for build-in cloud providers (like aws, gpc, etc)
It is possible to provide additional required data for the shoot template.

This PR also adds a configuration option for the available cloud providers (`cloudProviderList`). This allows to add additional cloud providers as well as configure their order in the Dashboard. It is possible to add cloud providers without additional configuration, the secret dialog and logo will fallback to generic input / icon.

***Current Limitations:***
No additional input fields on create cluster page, see #1445

***Example dashboard configuration for a cloud provider with name `custom`***
```yaml
cloudProviderList:
  - gcp
  - aws
  - custom
customCloudProviders:
  custom:
    zoned: false
    shoot:
      specTemplate:
        provider:
          type: custom
          infrastructureConfig:
            apiVersion: custom.provider.extensions.gardener.cloud/v1alpha1
            kind: InfrastructureConfig
            networks:
              vpc:
                cidr: ${workerCIDR}
          controlPlaneConfig:
            apiVersion: custom.provider.extensions.gardener.cloud/v1alpha1
            kind: ControlPlaneConfig
        networking:
          nodes: ${workerCIDR}
          foo: bar
    secret:
      fields:
      - key: namespace
        hint: Enter a valid namespace
        label: Namespace
        validators:
          required: .+
          isNamespace: .+--.+
        validationErrors:
          required: Namespace is required
          isNamespace: Must be a valid namespace with format a--b
      - key: token 
        hint: Enter a valid token
        label: token
        password: true
        validators:
          required: .+
          isToken: ^.{10}$
        validationErrors:
          required: token is required
          isToken: Must be a valid token with length 10
      help: |
        #Custom Cloud Provider
        <br />
        <img src="/static/vendor_assets/custom.png" width="100px" />
        <br />
        ## Namespace
        Please enter a valid namespace with format a--b
        <br />
        ## Token
        Please enter a valid token with exactly 10 chars
vendors:
    aws:
      icon: amazon.svg
      name: Amazon
    custom:
      name: Custom CC
      icon: custom.png
```

Icons can be provided as base64 encoded values when using the [helm chart](https://github.com/gardener/dashboard/blob/master/charts/gardener-dashboard) in the frontendConfig.vendorAssets map. Those icons can then be refrenced in the vendor configurations as well as the help html as shown above.

***Cloud provider `custom` rendered in the Dashboard***
Screenshot 2023-03-06 at 16 51 34
Screenshot 2023-03-06 at 16 53 47
Screenshot 2023-03-06 at 16 54 20

Fixes 

**Release note**:

```feature operator
Define order of cloud providers in dashboard configuration
```
```feature operator
Define additional cloud providers in dashboard configuration
```

### DIFF
--- a/backend/lib/routes/config.js
+++ b/backend/lib/routes/config.js
@@ -57,7 +57,8 @@ function sanitizeFrontendConfig (frontendConfig) {
     } = {},
     vendorHints = [],
     resourceQuotaHelp = '',
-    controlPlaneHighAvailabilityHelp = ''
+    controlPlaneHighAvailabilityHelp = '',
+    customCloudProviders = {}
   } = sanitizedFrontendConfig
 
   convertAndSanitize(alert, 'message')
@@ -87,6 +88,13 @@ function sanitizeFrontendConfig (frontendConfig) {
 
   for (const vendorHint of vendorHints) {
     convertAndSanitize(vendorHint, 'message')
+  }
+
+  for (const key of Object.keys(customCloudProviders)) {
+    const secret = customCloudProviders[key]?.secret
+    if (secret) {
+      convertAndSanitize(secret, 'help')
+    }
   }
 
   return sanitizedFrontendConfig
--- a/charts/__tests__/gardener-dashboard/runtime/dashboard/configmap.spec.js
+++ b/charts/__tests__/gardener-dashboard/runtime/dashboard/configmap.spec.js
@@ -696,6 +696,111 @@ describe('gardener-dashboard', function () {
       })
     })
 
+    describe('cloudProviderList', function () {
+      it('should render the template with cloudProviderList', async function () {
+        const values = {
+          global: {
+            dashboard: {
+              frontendConfig: {
+                cloudProviderList: [
+                  'foo',
+                  'bar'
+                ]
+              }
+            }
+          }
+        }
+        const documents = await renderTemplates(templates, values)
+        expect(documents).toHaveLength(1)
+        const [configMap] = documents
+        const config = yaml.load(configMap.data['config.yaml'])
+        expect(pick(config, ['frontend.cloudProviderList'])).toMatchSnapshot()
+      })
+    })
+
+    describe('customCloudProviders', function () {
+      it('should render the template with customCloudProviders', async function () {
+        const values = {
+          global: {
+            dashboard: {
+              frontendConfig: {
+                customCloudProviders: {
+                  foo: {
+                    zoned: true,
+                    shoot: {
+                      provider: {
+                        type: 'foo'
+                      }
+                    },
+                    secret: {
+                      fields: [
+                        {
+                          key: 'user',
+                          type: 'text'
+                        },
+                        {
+                          key: 'password',
+                          type: 'password'
+                        }
+                      ]
+                    }
+                  },
+                  bar: {
+                    zoned: false,
+                    shoot: {
+                      provider: {
+                        type: 'bar'
+                      }
+                    },
+                    secret: {
+                      fields: [
+                        {
+                          key: 'user',
+                          type: 'text'
+                        },
+                        {
+                          key: 'password',
+                          type: 'password'
+                        }
+                      ]
+                    }
+                  }
+                }
+              }
+            }
+          }
+        }
+        const documents = await renderTemplates(templates, values)
+        expect(documents).toHaveLength(1)
+        const [configMap] = documents
+        const config = yaml.load(configMap.data['config.yaml'])
+        expect(pick(config, ['frontend.customCloudProviders'])).toMatchSnapshot()
+      })
+    })
+
+    describe('vendors', function () {
+      it('should render the template with vendors', async function () {
+        const values = {
+          global: {
+            dashboard: {
+              frontendConfig: {
+                vendors: {
+                  foo: {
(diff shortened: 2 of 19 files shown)


frontend/src/components/GGenericInputFields.vue
```diff
@@ -0,0 +1,243 @@
+<!--
+SPDX-FileCopyrightText: 2023 SAP SE or an SAP affiliate company and Gardener contributors
+
+SPDX-License-Identifier: Apache-2.0
+-->
+
+<template>
```
petersutter, line 7: introduce GGenericInputField.vue component that handles one field for better readability. Not sure if a component like GGenericInputFields.vue is then still necessary

docs/operations/vendors-custom-cps.md
```diff
@@ -0,0 +1,234 @@
...
+              hint: Select at least one predefined value
+              label: Select Value
+              type: select-multiple
+              validators:
+                required:
+                  type: required
+              values:
+                - foo
+                - bar
+            specTemplate:
```
petersutter, line 130: I could not get the `specTemplate` to work, the following was my config (snipped)

```
  customCloudProviders:
    custom-provider:
      shoot:
        specTemplate:
          provider:
            type: DUMMY
          networking:
            nodes: 9.8.7.6
```

Lappihuan:
This is a amazing feature for gardener!
I would be intrested to contribute to this PR, what is the blocker here?

grolu:
Hello @Lappihuan Thank you for bumping this PR!
Actually there is no real blocker. However, we had some internal discussions and questioned the need for it. This PR started as a kind of a TGIF project to reduce the number of vendor specific dialogs containing boilerplate code and to ease the adoption of new cloud providers. However we did not see much need here (not so much new cloud providers lately) - however there might be custom dashboard builds for other vendors that I'm not aware of. For this use case this feature would make it way easier to keep changes as small as possible.
This PR has also grown very large and it incorporates more than just he dynamic cloud provider configuration. Specifically the option to overwrite vendor names and icons should be separated into a dedicated PR.
Moreover, this PR needs to be adapted to the latest changes in master branch first.
Please let me know which features you like most and where you see the use cases.

Lappihuan:
While whitelabeling is a nice-to-have, I believe the shoot form customization is what truly makes this powerful.

grolu:
Well, right now this PR focus on adding some (basic) possibilities to add additional fields to the create cluster dialog. You could provide simple input fields or a selection with values provided by the cloud profile. Feel free to further explain your use cases so that we can think of a way to make this work.

Lappihuan:
> This PR has also grown very large and it incorporates more than just he dynamic cloud provider configuration. Specifically the option to overwrite vendor names and icons should be separated into a dedicated PR.

I agree the whole whitelabel part would be best separated from the form customization feature.

> Please let me know which features you like most and where you see the use cases.

There have been multiple attempts to bridge the gap between the UI and the actual Shoot resource. Particularly with Extensions, while a few are supported directly, many are only accessible through the YAML editor.

There is the usecase of [Shoot Template](https://github.com/gardener/dashboard/issues/208#issuecomment-429385318)(s) either as a whole Default template or arround the Custom Fields / Form Customization.

- Setting a default CNI; currently, it defaults to the first available option (`canico`)
- Specifying a default DNS provider; currently defaults to the first provider listed (`aws-route53`)
- Enabling HA control planes by default (see #2227)
- Requiring end-users to specify a CIDR for the kube API server via the [acl extension](https://github.com/stackitcloud/gardener-extension-acl)
- Offering users a predefined selection of Flux repositories directly applied via the [shoot-flux extension](https://github.com/stackitcloud/gardener-extension-shoot-flux)
- Allowing end-users to conveniently input OIDC configuration via forms
- Pre-configuring registry caches with specific registries
- Simplifying or decluttering the Shoot Access Card by selectively disabling features we don't want to highlight (see #1705).
- Then there is the Provider specific parameters like for example the [csiManila](https://gardener.cloud/docs/extensions/infrastructure-extensions/gardener-extension-provider-openstack/usage/#controlplaneconfig)

Currently, most of these configurations are achievable via YAML editing. However, we can't effectively set defaults or guarantee specific extensions are consistently enabled or configured.

I'm also unclear if the existing [Custom Fields](https://github.com/gardener/dashboard/blob/master/docs/usage/custom-fields.md) already solve some of these scenarios. It would be helpful to clarify how much overlap exists between Custom Fields and the proposed Form Customization feature.

Some of our requirements may indeed be highly specific. I understand that if our customization needs exceed what's feasible upstream, maintaining a fork might be necessary.

The issues mentioned in this PR are still relevant too:

> Fixes https://github.com/gardener/dashboard/issues/1445 Fixes https://github.com/gardener/dashboard/issues/554 Fixes https://github.com/gardener/dashboard/issues/1707 Fixes https://github.com/gardener/dashboard/issues/207

PR #1558, opened 2023-09-13
Refresh rules

petersutter:
Refresh subject rules when changing the permissions of your own (non-admin) user. E.g. you make yourself a viewer instead of project admin. The rules should be updated accordingly

This PR fixed an issue where the wrong function was used (`refreshSubjectRules`), which was a leftover of the pre-vue3 code.

Fixes #

**Release note**:

```bugfix user

```

### DIFF
--- a/frontend/src/store/authz.js
+++ b/frontend/src/store/authz.js
@@ -135,13 +135,22 @@ export const useAuthzStore = defineStore('authz', () => {
       canCreateTerminals.value
   })
 
-  async function fetchRules (namespace, force) {
-    if (force || (namespace && spec.value?.namespace !== namespace)) {
-      const body = { namespace }
-      const response = await api.getSubjectRules(body)
+  // reuse function not exported
+  async function getRules (namespace) {
+    const body = { namespace }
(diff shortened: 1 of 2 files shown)


grolu (approved):
/lgtm

frontend/src/store/authz.js
```diff
@@ -135,8 +135,8 @@ export const useAuthzStore = defineStore('authz', () => {
       canCreateTerminals.value
   })
 
-  async function fetchRules (namespace) {
-    if (namespace && spec.value?.namespace !== namespace) {
+  async function fetchRules (namespace, force) {
+    if (force || (namespace && spec.value?.namespace !== namespace)) {
```
holgerkoser, line 139: I would propose to add a new method
```
// reuse function not exported
async function getRules (namespace) {
}

async function refreshRules () { 
  // use the fetch code
}
```

petersutter: ✅

holgerkoser (approved):
/lgtm

PR #1559, opened 2023-09-14
Fixed DNS popover alignment and link color

grolu:

Fixes #

**Release note**:

```bugfix user

```


frontend/src/components/ShootDns/GDnsProvider.vue
```diff
@@ -28,17 +28,19 @@ SPDX-License-Identifier: Apache-2.0
         />
       </v-chip>
     </template>
-    <v-list class="pa-0">
+    <v-list>
```
petersutter, line 31: we should increase the min-width of the popover as the title could get cut off
Screenshot 2023-09-14 at 15 20 01

grolu: done

petersutter (approved):
/lgtm

frontend/src/components/ShootDns/GDnsProvider.vue
```diff
@@ -28,17 +28,19 @@ SPDX-License-Identifier: Apache-2.0
         />
       </v-chip>
     </template>
-    <v-list class="pa-0">
+    <v-list min-width="300px">
```
holgerkoser, line 31: We usually do not use units for component widths.
```suggestion
    <v-list min-width="300">
```

PR #1561, opened 2023-09-14
terminals: Use gardener provided kube-apiserver endpoints. Drop `terminal-bootstrap` component

### LINKED ISSUE #1471: Delete terminal-bootstrap component
Gardener meanwhile exposes trusted certificates for the `Shoot` cluster kube-apiserver https://github.com/gardener/gardener/pull/7712, if there is a wildcard certificate configured.

It is planned that the same is done for the non-managed `Seed` clusters. 

Once this is done, the [terminal-bootstrap](https://github.com/gardener/dashboard/tree/master/packages/terminal-bootstrap) component can be considered obsolete and we can get rid of it.

- [ ] Remove https://github.com/gardener/dashboard/tree/master/packages/terminal-bootstrap
- [ ] Adapt Helm charts
- [ ] Update Documentation

**Why is this needed**:
This component is then obsolete and we can drop the code.

petersutter:
With https://github.com/gardener/gardener/pull/7764 and https://github.com/gardener/gardener/pull/7712 g/g exposes the `kube-apiserver`s with trusted certificates, which makes the `terminal-bootstrap` component of the dashboard obsolete.

This PR makes the necessary changes and drops the `terminal-bootrap` component altogether.

This PR includes a workaround that involves establishing a TLS connection in case the wss connection failed:

Users experience an `ERR_SSL_CLIENT_AUTH_CERT_NEEDED` error when attempting to open a wss connection to a kube-apiserver. This issue can be observed when using [`nginx.ingress.kubernetes.io/ssl-passthrough=true`](http://nginx.ingress.kubernetes.io/ssl-passthrough=true) on the Ingress.
During the TLS handshake, the server sends a `CertificateRequest` message, asking for client-side SSL certificates for authentication.
The issue arises when the `OnCertificateRequested` function is called when the request context does not already have a client certificate selection for the endpoint. In this case, the request is canceled.
```c
void Delegate::OnCertificateRequested(URLRequest* request,
                                      SSLCertRequestInfo* cert_request_info) {
  // This method is called when a client certificate is requested, and the
  // request context does not already contain a client certificate selection for
  // the endpoint. In this case, a main frame resource request would pop-up UI
  // to permit selection of a client certificate, but since WebSockets are
  // sub-resources they should not pop-up UI and so there is nothing more we can
  // do.
  request->Cancel();
}
```
https://source.chromium.org/chromium/chromium/src/+/main:net/websockets/websocket_stream.cc;drc=54d1c18dc0ae31fbb1d31bcf513d10ae53f4f90a;l=436
This problem does not occur in Safari or Firefox browsers.
The issue is tracked in [#993907](https://bugs.chromium.org/p/chromium/issues/detail?id=993907).

# Cleanup resources
 As gardener took over the responsibility of providing `kube-apiserver` endpoints with trusted certificates the `terminal-bootstrap` component was dropped. As landscape operator, you have to clean up the created resources of the `terminal bootstrap` component.

For each `Seed`, list the resources with label `component=dashboard-terminal`, which the `terminal-bootstrap` component has created. The resource names should start with `dashboard-terminal-kube-apiserver*`
```bash
kubectl get ingresses.networking.k8s.io --selector=component=dashboard-terminal --all-namespaces
kubectl get services --selector=component=dashboard-terminal --all-namespaces
kubectl get endpoints --selector=component=dashboard-terminal --all-namespaces
```

If you are sure that the above result did not contain any other unexpected resources you can delete them using:
```bash
kubectl delete ingresses.networking.k8s.io --selector=component=dashboard-terminal --all-namespaces
kubectl delete services --selector=component=dashboard-terminal --all-namespaces
kubectl delete endpoints --selector=component=dashboard-terminal --all-namespaces
```

Fixes #1471

**Release note**:

```breaking operator
Terminals: The `connectSrc` directive of the Content Security Policy no longer permits the use of `wss:` by default. You need to set up the allowed hosts that the browser can connect to for the web terminal feature. This can be done using `Values.global.terminal.allowedHostSourceList`. For more details see [webterminals.md#allow-list-for-hosts](https://github.com/gardener/dashboard/blob/master/docs/operations/webterminals.md#allow-list-for-hosts)
```

```breaking operator
Terminals: As gardener took over the responsibility of providing kube-apiserver endpoints with trusted certificates the `terminal-bootstrap` component was dropped. Clean up the created resources of the terminal bootstrapper. See PR description for more details.
```


frontend/src/lib/terminal.js
```diff
@@ -152,6 +154,27 @@ export class TerminalSession {
       if (this.cancelConnect) {
         return
       }
+
+      if (!this.workaroundOnce) {
```
holgerkoser, line 158: I would propose to move the workaround code into function `waitForPodRunning` by wrapping the the connectPromise code into a function that return an open websocket connection.
```js
async function waitForPodRunning (hostCluster, containerName, handleEvent, timeoutSeconds) {
  const connectTimeoutSeconds = 5

  const createConnection = () => {
    const protocols = ['garden'] // there must be at least one other subprotocol in addition to the bearer token
    addBearerToken(protocols, hostCluster.token)
    const url = watchPodUri(hostCluster)
    const ws = new WebSocket(url, protocols)

    const connectPromise = new Promise((resolve, reject) => {
      const openHandler = () => {
        ws.removeEventListener('open', openHandler)
        ws.removeEventListener('error', errorHandler)

        resolve(ws)
      }
      const errorHandler = ev => {
        ws.removeEventListener('open', openHandler)
        ws.removeEventListener('error', errorHandler)

        reject(new Error(`Failed to create connection to ${url}`))
      }

      ws.addEventListener('open', openHandler)
      ws.addEventListener('error', errorHandler)
    })

    return pTimeout(connectPromise, connectTimeoutSeconds * 1000, `Could not connect within ${connectTimeoutSeconds} seconds`)
  }

  let ws = null
  try {
    ws = await createConnection()
  } catch (err) {
    logger.info(err.message)
    ws = null
  }

  if (!ws) {
    // workaround for https://bugs.chromium.org/p/chromium/issues/detail?id=993907&q=ERR_SSL_CLIENT_AUTH_CERT_NEEDED&can=2
    logger.info('Trying connect workaround ...')
    try {
      // We want the tls handshake to be establised so that the request context has a client certificate selection for the endpoint.
      // It does not matter that the request will fail with 401.
      const options = {
        method: 'HEAD',
        mode: 'no-cors',
      }
      await fetch(`https://${hostCluster.kubeApiServer}`, options)
    } catch (err) { /* ignore errors */ }
    ws = await createConnection()
  }

  const podRunningPromise = new Promise((resolve, reject) => {
    ... 
  })

  try {
    await pTimeout(podRunningPromise, timeoutSeconds * 1000, `Timed out after ${timeoutSeconds}s`)
  } finally {
    closeWsIfNotClosed(ws)
  }
}
```

This first parameter of this function is now the hostCluster and not the websocket. This must be adapted. I would also move the finally handling into this function. We could also switch to long polling as a workaround. 
```js

  async waitUntilPodIsRunning (timeoutSeconds) {
    ...
    return waitForPodRunning(this.hostCluster, containerName, onPodStateChange, timeoutSeconds * 1000)
  }
```

petersutter: ✅

holgerkoser (approved):
/lgtm

PR #1562, opened 2023-09-15
Log error when when OIDC issuer discovery fails

petersutter:
Log error when when OIDC issuer discovery fails

Fixes #

**Release note**:

```other operator
An error will now be logged in case the OIDC issuer discovery fails
```


holgerkoser (approved):
/lgtm

PR #1566, opened 2023-09-18
Fixes loading in placeholder components

### LINKED ISSUE #1563: Navigate from a shoot to it's seed des not work.
**What happened**:
If the user is on the shoot details page and triggers navigation to the corresponding seed the subscription does not work and the UI behaves like the shoot is in creation state because there are no data.

**Environment**:

holgerkoser:
This PR fixes the loading Fixes in the placeholder components.

Fixes #1563 

**Release note**:

```bugfix user

```


petersutter (approved):
/lgtm

PR #1575, opened 2023-09-22
Fixed some shoot item navigation issues

holgerkoser:
**What this PR does / why we need it**:

This PR addresses inconsistencies in component initialization when using Vue Router's [_Fetching Before Navigation_](https://router.vuejs.org/guide/advanced/data-fetching.html#Fetching-Before-Navigation) approach. In development scenarios with Hot Module Replacement (HMR), the `beforeRouteEnter` and `beforeRouteUpdate` guards are not consistently invoked. However, Vue's [Component Lifecycle Hooks](https://vuejs.org/guide/essentials/lifecycle.html#lifecycle-diagram) are still triggered.

Our placeholder components rely on Vue's built-in [Meta Component](https://vuejs.org/api/built-in-special-elements.html#component) to dynamically render child components based on navigation or loading states.
```vue
<template>
  <component
    :is="component"
    v-bind="componentProperties"
  />
</template>
```
During this process, the component displays a loading indicator, error components for failures, or `<router-view/>` when navigation completes successfully.

##### Key Changes:

1. **Single Initialization**: 
   Ensures that component initialization only occurs once, either upon route change or initial page load.
   
2. **State Management**: 
  Manages the release of previous states (unsubscriptions and data resets) when navigating to new routes.

3. **Conditional Rendering**: 
  Ensures that `<router-view/>` is only displayed once navigation is complete to prevent rendering issues, such as brief flashes of old route components that get immediately deleted.

By implementing these changes, we aim to provide a more consistent and error-free navigation experience. This is crucial because any inconsistencies could lead to a range of issues, from rendering problems to violation of component property constraints.
 

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Release note**:

```bugfix user

```

PR #1577, opened 2023-09-22
Remove unnecessary component name

petersutter:
This PR removes the unnecessary component name
GShootItem is not touched as the name will be removed with PR https://github.com/gardener/dashboard/pull/1575

Fixes #

**Release note**:

```bugfix user

```

### DIFF
--- a/frontend/src/components/floating/PopperWrapper.vue
+++ b/frontend/src/components/floating/PopperWrapper.vue
@@ -87,7 +87,6 @@ import {
 const referenceProps = ['class']
 
 export default {
-  name: 'VPopperWrapper',
   components: {
     Popper: Popper(),
     PopperContent,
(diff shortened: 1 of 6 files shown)


frontend/src/components/floating/PopperWrapper.vue
```diff
@@ -87,7 +87,6 @@ import {
 const referenceProps = ['class']
 
 export default {
-  name: 'VPopperWrapper',
```
holgerkoser, line 90: This case is different since we create a variant of this component [VPopperWrapper](https://github.com/Akryum/floating-vue/blob/main/packages/floating-vue/src/components/PopperWrapper.vue#L61). I would keep the original name `VPopperWrapper`.

petersutter: ✅

holgerkoser (approved):
/lgtm

PR #1579, opened 2023-09-25
Use text-medium-emphasis

petersutter:
https://vuetifyjs.com/en/styles/text-and-typography/#opacity
> Opacity helper classes allow you to easily adjust the emphasis of text. text-high-emphasis has the same opacity as default text. text-medium-emphasis is used for hints and helper text. De-emphasize text with text-disabled.

On dark mode, the texts with text-secondary are too dark, even darker than `text-disabled` and thus look disabled in comparison.

left window: `text-secondary`
right window: `text-medium-emphasis`

Screenshot 2023-09-25 at 16 47 32
Screenshot 2023-09-25 at 16 48 10
Screenshot 2023-09-25 at 16 48 39

Fixes #

**Release note**:

```bugfix user

```

### DIFF
--- a/frontend/src/components/ShootAccessRestrictions/GAccessRestrictions.vue
+++ b/frontend/src/components/ShootAccessRestrictions/GAccessRestrictions.vue
@@ -151,7 +151,7 @@ export default {
     },
     textClass (definition) {
       return this.enabled(definition)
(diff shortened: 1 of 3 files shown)


holgerkoser (approved):
/lgtm

PR #1588, opened 2023-09-26
Trim user input

### LINKED ISSUE #1586: [master] Confirm dialog: input does not match
**What happened**:
When you copy the cluster name from the title in order to paste it in the input field the following error is displayed
Screenshot 2023-09-26 at 11 11 26
The reason for the input validation to fail is that the `&ZeroWidthSpace;` was copied in addition and thus the validation fails.

We should at least trim the input.

**Environment**:

petersutter:
This PR trims the user input and trims zero width space from the user input

Fixes #1586

**Release note**:

```bugfix user

```


grolu (approved):
/lgtm

PR #1598, opened 2023-10-04
Add support for projected volume kubeconfig

timuthy:
/area security
/kind enhancement

This PR adds the option to configure a kubeconfig projected volume for the dashboard deployment. It is for example needed, if operators generate their kubeconfigs for the virtual garden cluster via the [TokenRequestor](https://github.com/gardener/gardener/blob/master/docs/concepts/resource-manager.md#tokeninvalidator-controller).

**Release note**:

```feature operator
The `dashboard` chart allows to optionally configure a projected volume based kubeconfig.
```


charts/gardener-dashboard/values.yaml
```diff
@@ -29,6 +29,11 @@ global:
       expirationSeconds: 43200 # 12 hours
       audience: ''
 
+  # projectedKubeconfig:
```
petersutter, line 32: ideally, also add a comment here describing this option

petersutter (approved):
just a nit regarding adding an additional comment in the values.yaml file to describe the option, otherwise: /lgtm

holgerkoser (approved):
/lgtm

grolu (approved):
/lgtm

PR #1599, opened 2023-10-04
Fixed some issues with workerless cluster creation

grolu:
- Reset Workerless flag in resetNewShootResource navigation guard
- Reset infrastructure specific template in shoot resource when toggling workerless flag

Fixes #

**Release note**:

```bugfix user
NONE
```

### DIFF
--- a/frontend/src/store/shoot/index.js
+++ b/frontend/src/store/shoot/index.js
@@ -31,6 +31,7 @@ import { useSecretStore } from '../secret'
 import { useSocketStore } from '../socket'
 import { useTicketStore } from '../ticket'
 import { useLocalStorageStore } from '../localStorage'
+import { useShootStagingStore } from '../shootStaging'
 
 import {
   uriPattern,
@@ -72,6 +73,7 @@ export const useShootStore = defineStore('shoot', () => {
   const socketStore = useSocketStore()
   const projectStore = useProjectStore()
   const localStorageStore = useLocalStorageStore()
+  const shootStagingStore = useShootStagingStore()
 
   const context = {
     api,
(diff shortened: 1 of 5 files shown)


frontend/src/store/shoot/index.js
```diff
@@ -441,6 +443,7 @@ export const useShootStore = defineStore('shoot', () => {
 
     state.newShootResource = value
     state.initialNewShootResource = cloneDeep(value)
+    shootStagingStore.workerless = false
```
petersutter, line 446: maybe introduce a reset function on shootStagingStore?

grolu: As discussed, we could do this. Right now the shootStagingStore is only used by components that set the state when they are initialized. `workerless` is the only value that needs to be reset as it is not represented by a shoot spec attribute. It is used to back the state of the checkbox only.
However, I'm not against introducing a reset method.

frontend/src/views/GNewShoot.vue
```diff
@@ -297,6 +298,18 @@ export default {
       this.setClusterConfiguration(this.newShootResource)
     }
   },
+  watch: {
+    async workerless(value) {
```
holgerkoser, line 302:
```suggestion
    workerless(value) {
```

frontend/src/views/GNewShoot.vue
```diff
@@ -297,6 +298,18 @@ export default {
...
+      if(!value) {
+        const shootResource = cloneDeep(this.newShootResource)
+        const cloudProfileName = shootResource.spec.cloudProfileName
+        const infrastructureKind = shootResource.spec.provider.type
+        const defaultNodesCIDR = this.getDefaultNodesCIDR({ cloudProfileName })
+        assign(shootResource.spec, getSpecTemplate(infrastructureKind, defaultNodesCIDR))
```
holgerkoser, line 308:
```suggestion
        Object.assign(shootResource.spec, getSpecTemplate(infrastructureKind, defaultNodesCIDR))
```

holgerkoser (approved):
/lgtm

PR #1613, opened 2023-10-19
[hotfix 1.70] fixes memory leaks in `GTimeString` and `localStorage` store

holgerkoser:
This PR fixes memory leak in GTimeString and localStorage store. This issue was causing an unnecessary allocation of memory, which, over time, led to degraded UI performanceand eventually leading to Out of Memory crashes.

Fixes #

**Release note**:

```bugfix user
Fixed two memory leaks in `GTimeString` and `localStorage` store. This issue was causing an unnecessary allocation of memory, which, over time, led to degraded UI performance and eventually leading to Out of Memory crashes.
```

### DIFF
--- a/frontend/src/components/GTimeString.vue
+++ b/frontend/src/components/GTimeString.vue
@@ -101,6 +101,9 @@ export default {
     }
   },
   computed: {
+    nextClockTimeValue () {
+      return toValue(this.nextClockTime)
+    },
     relDateTimeString () {
       if (this.mode === 'future' && !isValidTerminationDate(this.dateTime)) {
         return this.expiredText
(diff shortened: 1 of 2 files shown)


grolu (approved):
/lgtm

frontend/src/components/GTimeString.vue
```diff
@@ -34,154 +32,145 @@ import {
...
+  watch: {
+    dateTime (value) {
+      if (value) {
+        this.updateClockInstance(value)
+      }
+    },
+    nextClockTime (value) {
+      if (value) {
+        this.updateClockInstance(this.dateTime)
+      }
```
petersutter, line 139: `nextClockTime` watch does not work

petersutter (approved):
/lgtm

PR #1624, opened 2023-10-24
Keep the current URL path if the user is automatically logged out

### LINKED ISSUE #1619: During navigation the URL path is lost if a session cookie is present but expired
**What happened**:
When a user who has not accessed the dashboard for an extended period attempts to open a specific shoot detail page, they are redirected to the login page due to session expiration. During this process, the original redirectPath is lost.

After successfully logging in, the user should be directed to the originally requested shoot detail page.

1. Make sure you have previously logged into the dashboard and that your session has since expired.
2. Navigate directly to a specific shoot cluster page, e.g., /namespace/garden-core/shoots/<shoot-name>.
3. Note that you are redirected to the login page.
4. Click the login button.
5. Observe that you are redirected to /namespace/garden/shoots, not the specific shoot detail page at /namespace/garden-core/shoots/<shoot-name>.

The problem arises when the JWT token expires. We reset the user in authnStore.$reset even when the token has expired. Because all the information is in the store, this creates a "chicken-and-egg" issue ([code reference](https://github.com/gardener/dashboard/blob/master/frontend/src/router/guards.js#L52-L54)).

socketStore has a watch on authnStore.user. When the user is reset, it attempts to establish a socket connection ([code reference](https://github.com/gardener/dashboard/blob/master/frontend/src/store/socket/index.js#L125-L131)). This attempt fails, resulting in a signout without preserving the redirectPath.

**Environment**:

holgerkoser:
Users will now be redirected back to the last visited URL after automatic logout due to session timeout or authentication errors.

Fixes #1619 

**Release note**:

```bugfix user
Users will now be redirected back to the last visited URL after automatic logout due to session timeout or authentication errors.
```

### DIFF
--- a/frontend/src/router/guards.js
+++ b/frontend/src/router/guards.js
@@ -58,7 +58,7 @@ export function createGlobalBeforeGuards () {
         : 'Session has expired'
       logger.info('%s --> Redirecting to login page', message)
 
-      const query = redirectPath
+      const query = redirectPath && redirectPath !== '/'
         ? { redirectPath }
         : undefined
       return {
--- a/frontend/src/router/guards.js
+++ b/frontend/src/router/guards.js
@@ -36,8 +36,6 @@ export function createGlobalBeforeGuards () {
   const shootStore = useShootStore()
   const terminalStore = useTerminalStore()
 
-  let sessionTimeoutId
-
   function ensureUserAuthenticatedForNonPublicRoutes () {
     return to => {
       const {
@@ -51,17 +49,7 @@ export function createGlobalBeforeGuards () {
 
       authnStore.$reset()
 
-      const expiresIn = authnStore.sessionExpiresAt - Date.now()
-      if (expiresIn > 0) {
-        clearTimeout(sessionTimeoutId)
-        sessionTimeoutId = setTimeout(() => {
-          if (!authnStore.isExpired()) {
-            logger.info('User session updated by other browser process --> Skipping signout')
-            return
-          }
-          logger.info('User session is expired --> Initiating signout')
-          authnStore.signout(null, redirectPath)
-        }, expiresIn)
+      if (!authnStore.isExpired()) {
         return true
       }
 
--- a/frontend/src/router/guards.js
+++ b/frontend/src/router/guards.js
@@ -55,6 +55,10 @@ export function createGlobalBeforeGuards () {
       if (expiresIn > 0) {
         clearTimeout(sessionTimeoutId)
         sessionTimeoutId = setTimeout(() => {
+          if (!authnStore.isExpired()) {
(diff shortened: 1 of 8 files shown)


frontend/src/store/authn/helper.js
```diff
@@ -61,13 +60,17 @@ function isExpired (user) {
...
     logger = useLogger(),
   } = options ?? {}
 
   let refreshTokenPromise
+  let signoutInProgress = false
 
-  const origin = unref(location).origin
+  useEventListener(window, 'beforeunload', () => {
+    signoutInProgress = true
+  })
```
petersutter, line 71: for which case is this relevant?

frontend/src/router/guards.js
```diff
@@ -55,6 +55,10 @@ export function createGlobalBeforeGuards () {
       if (expiresIn > 0) {
         clearTimeout(sessionTimeoutId)
         sessionTimeoutId = setTimeout(() => {
+          if (!authnStore.isExpired()) {
+            logger.info('User session updated by other browser process --> Skipping signout')
+            return
+          }
```
petersutter, line 61: shouldn't the session timeout be restarted?

frontend/src/store/authn/helper.js
holgerkoser: This is not strictly necessary, especially if the page is going to be completely unloaded during the sign-out process. Using `beforeunload` is more about keeping the state clean in real-time, which might not offer any practical benefit in this case.

frontend/src/router/guards.js
holgerkoser: Removed the sessionTimeout timer

frontend/src/store/socket/helper.js
```diff
@@ -118,7 +118,11 @@ export function createSocket (state, context) {
     switch (reason) {
       case 'io server disconnect': {
         logger.debug('socket was forcefully disconnected by the server')
-        reconnect()
+        if (authnStore.isExpired()) {
+          authnStore.signout()
```
petersutter, line 122: for me it's not clear why we do the signout only in case when `io server disconnect` and why not in any case if it is expired

holgerkoser: The important thing here is that we do not try to reconnect if the session is expired. The `io server disconnect` is the case when the API token needs to be refreshed. This should happen if the absolute lifetime of the session is over. We counld also remove the signout here or move it outside of the case. But the other caes will not occur I assume.

petersutter (approved):
/lgtm

frontend/src/store/socket/helper.js
holgerkoser: moved the signout at the bottom of the handler.

PR #1629, opened 2023-10-26
Fix: scroll alert into view

### LINKED ISSUE #1606: Error message on create cluster page only visible at the bottom
**What happened**:
When creating a shoot cluster fails an error message is displayed on the create cluster page. However it is only visible when you scroll to the bottom of the page

Select a k8s version >= 1.27 and press the create button.

**Environment**:

petersutter:
Fixed issue on create cluster page: On error, alert should scroll into view

Fixes #1606

**Release note**:

```bugfix user
Fixed issue on create cluster page: On error, alert should scroll into view
```


holgerkoser (approved):
/lgtm

PR #1633, opened 2023-10-30
Enhance Error Message Handling

### LINKED ISSUE #1618: Worker validation not properly triggered
**What happened**:
On azure, go to the yaml tab and add another zone. Go back to the overview tab. The save button is greyed out because of validation errors, but no error is shown. If you manually touch the zone config the validation error will appear (autoscaler max. muster be greater or equal to the number of zones configured for the respective pool).

**Environment**:

grolu:
This PR enhances several things related to vuelidate
- Added error messages to validators instead of having them in a separate data structure
- This also removes the requirement for the somewhat complicated `getVuelidateErrors` logic
- Removed all specific error messages like `Project name is required` if the default error message for a validator is sufficient for understanding the issue. This helps to align the error messages and keeps the code cleaner
- Added `field name` param to validators so that we can show what caused an error message, if the message is showed somewhere else (not directly at the input)
- Cleaned up some small validation issues (see commits for details)
- Use vuelidate for shoot version update dialog
- If `$invalid` is `true` the submit button is no longer greyed out. A click on the button triggers `$vuelidate()` to make sure that all error message are shown at the inputs. The errors are also shown in an alert the same way we show server side error messages

Screenshot 2023-12-01 at 15 56 26
Screenshot 2023-12-01 at 15 54 53
Screenshot 2023-12-01 at 15 54 02

Fixes #1618

Commits should be reviewed individually

**Release note**:

```feature user
Enhanced Client-Side Validation:
Previously, if there was a validation error, the `submit` buttons on Dialogs and the Create Cluster page would be greyed out. Now, these buttons remain active even if there is a validation error.
Clicking on the `submit` button in the presence of a validation error will not execute the submit logic. Instead, it will trigger the validation process to ensure that all validation error messages are clearly visible at the respective input fields.
In addition, we've made it easier to view all error messages. They are now collected and displayed in an alert, similar to how server-side error messages are shown. This enhancement ensures that users are fully aware of any errors that prevent to proceed with the action.
```


frontend/src/utils/validators.js
```diff
@@ -52,62 +49,76 @@ const unique = key => withParams(
...
-)
+const nilUnless = key => withMessage(`Must not be provided if '${key}' is not set`,
+  withParams(
+    { type: 'nilUnless', key },
+    function nilUnless (value) {
+      return !this[key] ? !req(value) : true
+    },
+  ))
+
+const allWithCauserParam = (causer, validators) => {
```
holgerkoser, line 102: I think `causer` is not a good name. Maybe we should use collection validation and `$registerAs` to have meaningful `$propertyPath` as in this test https://github.com/vuelidate/vuelidate/blob/bdb672d334456461bcf9058dc551a21f0152d3d9/packages/test-project/src/components/CollectionValidations.vue#L9. And than map the propertyPath to a label?

frontend/src/components/GVuelidateTooltip.vue
```diff
@@ -0,0 +1,54 @@
+<!--
```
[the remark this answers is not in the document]
grolu: We now show these errors also as alert as you can see in the last screenshot

frontend/src/utils/validators.js
grolu: renamed to `withFieldName`

frontend/src/components/GPurpose.vue
```diff
@@ -16,7 +16,7 @@ SPDX-License-Identifier: Apache-2.0
       item-title="purpose"
       item-value="purpose"
       persistent-hint
-      :error-messages="getErrorMessages('internalPurpose')"
+      :error-messages="errors.internalPurpose"
```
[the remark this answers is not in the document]
grolu: Refactored to reuse function as described above

frontend/src/components/dialogs/GDialog.vue
```diff
@@ -86,30 +86,34 @@ SPDX-License-Identifier: Apache-2.0
...
         <v-tooltip
           location="top"
-          :disabled="valid"
+          :disabled="!notConfirmed || hasVisibleErrors || !valid"
         >
           <template #activator="{ props }">
             <div v-bind="props">
-              <v-btn
+              <g-vuelidate-button
+                :v="v$"
```
holgerkoser, line 94: I don't think this is a good idea to pass the full vuelidate object down into another component.

frontend/src/components/GVuelidateButton.vue
```diff
@@ -0,0 +1,84 @@
+<!--
```
holgerkoser, line 1: I have never seen or used a UI where error messages are shown as a tooltip for button. I would strongly recommend to solve the problem in a different way.

frontend/src/components/GTimeTextField.vue
```diff
@@ -55,15 +56,10 @@ const textFieldType = computed(() => {
...
-    timeSyntax: (value) => {
-      return !value?.length || /^([01][0-9]|2[0-3]):([0-5][0-9])$/.test(value)
-    },
-  },
-}
-
-const validationErrors = {
-  timeValue: {
-    timeSyntax: 'Input value is not valid 24-hour time format (HH:MM)',
+    timeSyntax: withMessage('Input value is not valid 24-hour time format (HH:MM)',
```
holgerkoser, line 59: no nesting (see below)

frontend/src/components/GTimeTextField.vue
```diff
@@ -76,16 +72,11 @@ const inputHint = computed(() => {
 })
 
 const errorMessages = computed(() => {
-  return [
-    ...getErrorMessages('timeValue'),
-    ...props.errorMessages,
-  ]
+  const errors = getErrorMessages(v$.value.timeValue)
```
holgerkoser, line 75: Does this function return errors (error objects) or error messages. If it returns errorMessages the constant should have the name errorMessages.

frontend/src/components/NewShoot/GNewShootDetails.vue
```diff
@@ -170,7 +157,22 @@ export default {
     }
   },
   validations () {
-    return this.validators
+    return {
```
holgerkoser, line 160: avoid nesting

frontend/src/components/NewShoot/GNewShootInfrastructureDetails.vue
```diff
@@ -272,91 +276,55 @@ export default {
...
+    return {
+      region: withFieldName('Region', {
+        required,
+      }),
+      networkingType: withFieldName('Networking Type', {
+        required: requiredIf(!this.workerless),
+      }),
+      loadBalancerProviderName: withFieldName('Cluster Name', {
+        required: requiredIf(function () {
+          return this.infrastructureKind === 'openstack'
```
holgerkoser, line 288: We use function not arrow function. Is this still needed? 
Can't we have local helper functions:
```js
const requiresInfrastructure = infrastructureKind => {
  return requiredIf(() => this.infrastructureKind === infrastructureKind)
}
// usage
const rules = {
        required: requiresInfrastructure('openstack')
}
```
This will also removes one level of nesting

frontend/src/components/Secrets/GSecretDialogAlicloud.vue
```diff
@@ -8,7 +8,7 @@ SPDX-License-Identifier: Apache-2.0
   <g-secret-dialog
     v-model="visible"
     :data="secretData"
-    :data-valid="valid"
+    :secret-validations="v$"
```
holgerkoser, line 11: I have already said that I don't like the this pattern to pass the full vuelidate object to a subcomponent. I don't understand why this is needed and this is the onlt way to solve the issue.

holgerkoser (commented):
I have not fully reviewed the PR. I think we should first talk about these comment before I continue.

frontend/src/views/GNewShoot.vue
```diff
@@ -151,16 +151,17 @@ SPDX-License-Identifier: Apache-2.0
     <v-divider />
     <div class="d-flex align-center justify-end toolbar">
       <v-divider vertical />
-      <v-btn
+      <g-vuelidate-button
```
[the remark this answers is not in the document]
grolu: removed component as discussed

frontend/src/views/GNewShoot.vue
```diff
@@ -151,16 +151,17 @@ SPDX-License-Identifier: Apache-2.0
...
+        :v="v$"
         variant="text"
-        :disabled="v$.$invalid"
         color="primary"
         @click.stop="createClicked()"
+        @error-messages-updated="showVuelidateErrors"
```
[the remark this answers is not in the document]
grolu: event obsolete

frontend/src/components/dialogs/GDialog.vue
```diff
@@ -272,6 +284,10 @@ export default {
       cardContentRef.scrollTop = scrollTopVal + 10
       cardContentRef.scrollTop = scrollTopVal - 10
     },
+    showVuelidateErrors (messages) {
```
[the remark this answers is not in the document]
grolu: method removed

frontend/src/components/dialogs/GDialog.vue
grolu: removed component as discussed

frontend/src/components/Secrets/GSecretDialogAlicloud.vue
grolu: removed component

frontend/src/components/GWildcardSelect.vue
```diff
@@ -199,9 +174,24 @@ export default {
 
       this.onInput()
     },
+    getErrorMessages,
   },
   validations () {
-    return this.validators
+    const rules = {}
+    const wildcardSelectedValueRules = {
+      required: withMessage(`${this.wildcardSelectLabel} is required`, required),
```
holgerkoser, line 182: Here static but below it is reactive with getter.

frontend/src/components/GWildcardSelect.vue
```diff
@@ -199,9 +174,24 @@ export default {
...
+      prefixRequired: withMessage('Prefix is required', () => {
```
holgerkoser, line 183: Do we still want to override the default message for required?

frontend/src/components/GTimeTextField.vue
```diff
@@ -72,9 +72,7 @@ const inputHint = computed(() => {
 })
 
 const errorMessages = computed(() => {
-  const errors = getErrorMessages(v$.value.timeValue)
-  const errorMessages = [...props.errorMessages, ...errors]
-  return errorMessages
+  return [...props.errorMessages, ...getErrorMessages(v$.value.timeValue)]
```
holgerkoser, line 75:
```suggestion
  return [
    ...props.errorMessages, 
    ...getErrorMessages(v$.value.timeValue),
  ]
```

frontend/src/components/Members/GMemberDialog.vue
```diff
@@ -228,8 +230,47 @@ export default {
...
+        unique: withMessage(() => `User '${this.internalName}' is already member of this project.`, unique('projectUsernames')),
+        isNoServiceAccount: withMessage('Please use add service account to add service accounts', value => !isServiceAccountUsername(value)),
+      }
+      rules.internalName = withFieldName('User Name', internalNameRules)
+    } else if (this.isServiceDialog) {
+      const serviceAccountKeyFunc = (value) => {
+        return isServiceAccountUsername(value)
+          ? 'serviceAccountUsernames'
+          : 'serviceAccountNames'
+      }
```
holgerkoser, line 261:
```suggestion
      const serviceAccountKeyFunc = value => isServiceAccountUsername(value)
          ? 'serviceAccountUsernames'
          : 'serviceAccountNames'
```

frontend/src/components/GWildcardSelect.vue
holgerkoser, line 182:
```suggestion
      required: withMessage(() => `${this.wildcardSelectLabel} is required`, required),
```

frontend/src/components/NewShoot/GNewShootDetails.vue
```diff
@@ -170,7 +157,27 @@ export default {
...
+    const nameRules = {
+      required,
+      maxLength: maxLength(this.maxShootNameLength),
+      noConsecutiveHyphen,
+      noStartEndHyphen,
+      lowerCaseAlphaNumHyphen,
+      unique: withMessage('A cluster with this name already exists in this project',
+        (value) => {
+          return this.shootByNamespaceAndName({ namespace: this.namespace, name: value }) === undefined
+        }),
```
holgerkoser, line 171:
```suggestion
        value => !this.shootByNamespaceAndName({ namespace: this.namespace, name: value })
      ),
```

frontend/src/components/Secrets/GSecretDialogOpenstack.vue
```diff
@@ -250,13 +223,76 @@ export default {
...
+
+    rules.tenantName = withFieldName('Project / Tenant Name', {
+      required,
+    })
+
+    const usernameRules = {
+      required: withMessage(requiredUserMessage,
+        requiredIf(function () {
+          return this.authenticationMethod === 'USER'
+        }),
```
holgerkoser, line 247:
```suggestion
        requiredIf(() => this.authenticationMethod === 'USER'),
```

frontend/src/components/Secrets/GSecretDialogOpenstack.vue
```diff
@@ -250,13 +223,76 @@ export default {
...
+      ),
+    }
+    rules.username = withFieldName('Technical User', usernameRules)
+
+    const passwordRules = {
+      required: withMessage(requiredUserMessage,
+        requiredIf(function () {
```
holgerkoser, line 254: Always or never use arrow function inside of the validations method.

frontend/src/components/Secrets/GSelectSecret.vue
```diff
@@ -127,38 +131,28 @@ export default {
...
-        },
-      },
     }
   },
   validations () {
-    return this.validators
+    const projectName = get(this.secret, 'metadata.projectName')
+    const isSecretInProject = this.projectName === projectName
+
+    const requiresCostObjectIfEnabledMessage = isSecretInProject
```
holgerkoser, line 140: If the message contains any getter it should be always an arrow function

frontend/src/components/ShootWorkers/GWorkerInputGeneric.vue
```diff
@@ -286,64 +261,70 @@ export default {
...
+    rules.worker.name = withFieldName(() => `${this.workerGroupName} Name`, nameRules)
+
+    rules.worker.minimum = withFieldName(() => `${this.workerGroupName} Autoscaler Min.`, {
+      minValue: minValue(0),
+    })
+
+    const maximumRules = {
+      minValue: minValue(0),
+      systemComponents: withMessage('Value must be greater or equal to the number of zones configured for this pool',
+        (value) => {
```
holgerkoser, line 287: I think we should install the `@stylistic/eslint-plugin-js` plugin and enforce a consistent code style via eslint.

frontend/src/components/ShootWorkers/GWorkerInputGeneric.vue
```diff
@@ -286,64 +261,70 @@ export default {
...
   data () {
     return {
-      validationErrors,
       immutableZones: undefined,
       volumeSize: undefined,
     }
   },
   validations () {
-    return this.validators
+    const rules = { worker: {} }
```
holgerkoser, line 269: Why do we have this additional layer `worker` and not just return the rules?

frontend/src/components/Secrets/GSecretDialog.vue
```diff
@@ -332,7 +324,13 @@ export default {
...
     async submit () {
-      this.v$.$touch()
+      if (this.secretValidations.$invalid) {
+        await this.secretValidations.$validate()
+        const message = messageFromErrors(this.secretValidations.$errors)
+        this.errorMessage = 'There are input errors that you need to resolve'
+        this.detailedErrorMessage = message
+        return
+      }
       if (this.valid) {
```
petersutter, line 334: valid does not exist anymore. This if has to be removed. Currently it is not possible to create any secrets

frontend/src/components/ShootWorkers/GWorkerInputGeneric.vue
grolu: yes, we should do this in a separate PR

holgerkoser (approved):
/lgtm

PR #1668, opened 2023-12-21
Drop support for basic auth

### LINKED ISSUE #1659: Take serverUrl from shoot yaml
**What happened**:
instead of taking the server url from the (now deprecated) static token kubeconfig https://github.com/gardener/dashboard/blob/700aa44477f5de9557907808b94b3954bb2641d8/backend/lib/services/shoots.js#L355 it should be taken from `Shoot.status.advertisedAddresses`.

Because of this, the `dashboardUrl` on the `shootInfo` is not set for shoot clusters that do not have the static token kubeconfig enabled.
https://github.com/gardener/dashboard/blob/700aa44477f5de9557907808b94b3954bb2641d8/frontend/src/components/GShootListRow.vue#L309

**Environment**:

petersutter:
Basic auth is not supported by gardener since long and thus it does not make sense to still support it in the dashboard. See also https://github.com/gardener/gardener/pull/7534

Fixes #1659

**Release note**:

```bugfix user

```

### DIFF
--- a/backend/lib/services/shoots.js
+++ b/backend/lib/services/shoots.js
@@ -345,14 +345,13 @@ exports.info = async function ({ user, namespace, name }) {
     _
       .chain(secret)
       .get('data')
-      .pick('kubeconfig', 'username', 'password', 'token')
+      .pick('kubeconfig', 'token')
       .forEach((value, key) => {
         value = decodeBase64(value)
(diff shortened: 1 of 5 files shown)


grolu (approved):
/lgtm

holgerkoser (approved):
/lgtm

PR #1671, opened 2023-12-29
Fixing Popover Attachment in the Shoot List

### LINKED ISSUE #1650: Bind popper in shoot list to element
For performance and memory consumption reasons the table rows do no longer have an `:key` base on the uid. But the popper popover should be bound to element instance if the row changes. Find out how this binding could be done in a performant way.

**Why is this needed**:

holgerkoser:
This PR addresses an issue encountered in the Shoot List. Previously, when a popover was opened and the order of the list entries changed, the popover remained tied to the row index rather than the specific element it was associated with. This behavior led to inconsistencies and a confusing user experience.
With this update, we have modified the logic to store the unique identifier (UID) of the opened popover centrally in the Shoot list context. This approach ensures that the popover is consistently displayed for the correct element, even if the order of the entries changes. The popover will now correctly associate with the intended row, enhancing the interface's usability and reliability.

Fixes #1650

The solution to the problem is not optimal, as it leads to somewhat jerky UI behavior. Therefore, the animation delay has been set to 0. The popover also disappears when the cluster is no longer visible on the first page. When you flip to the second page, the popover reappears. This is not necessarily what one would expect. A better solution would be to dynamically bind the popover to an HTML element and always store the reference HTML element. However, this solution would mean a complete overhaul of the popover. Currently, the reference element is part of the popover component.

**Release note**:

```bugfix user

```


petersutter (approved):
/lgtm

grolu (approved):
/lgtm

PR #1675, opened 2024-01-04
Do not show hidden github comments

### LINKED ISSUE #1640: Please, do not show hidden cluster ticket comments
When tickets are shown for a cluster, please do not show hidden comments. There were (probably always) good reasons to hide a comment in GitHub, so they shouldn't be shown in the Dashboard either.

**Why is this needed**:
Show only appropriate content. Respect decision made for GitHub also in Dashboard.

holgerkoser:
This PR introduces functionality to retrieve GitHub comments using the GitHub [GraphQL API](https://docs.github.com/en/graphql/reference/objects#issuecomment). It leverages the `isMinimized` property, unique to the GraphQL API's IssueComment object, to filter out minimized comments. As a result, only non-minimized comments will be displayed on the dashboard.

Fixes #1640

**Release note**:

```feature user
Hidden GitHub comments will no longer be displayed on the Cluster Details page.
```


backend/lib/github/index.js
```diff
@@ -48,13 +48,53 @@ function closeIssue ({ number }) {
...
-    owner,
-    repo,
-    issue_number: number
-  })
-  return octokit.paginate(options)
+async function getComments ({ number }) {
+  const query =
+  `query paginate($cursor: String) {
+    repository(owner: "${owner}", name: "${repo}") {
+      issue(number: ${number}) {
```
holgerkoser, line 55: Escape all params

holgerkoser: ✔️

grolu (approved):
/lgtm

petersutter (approved):
/lgtm

PR #1681, opened 2024-01-11
use variant=tonal instead of outlined

grolu:

Fixes #

**Release note**:

```other user
Introduced tonal variations to our chips and alert designs, previously defined with outline styles
```


petersutter (approved):
/lgtm

PR #1683, opened 2024-01-11
Kubernetes Version Update Improvements

### LINKED ISSUE #1481: Rework kubernetes upgrade dialog
Check the current implementation. Do we really want to show impossible upgrade paths?

**Why is this needed**:
>Why do we list all those non-allowed upgrade versions at all, I mean if it's not allowed to skip a K8s minor version, why have some dynamic list that lists all non-allowed options (instead of e.g. just one single note mentioning that minor skipping is not allowed)

grolu:
Clusters can only be upgraded one minor kubernetes version at a time.
We used to show all versions, and disabled those that a cluster could not be upgraded to.

Now, these versions are filtered from the selection, only a hint indicates that there are more versions available.
Moreover, we now only promote supported versions on the version chip.
Also, show version chip with warning color in case current version is deprecated

Screenshot 2024-01-11 at 16 34 47
Screenshot 2024-01-11 at 16 33 34

Fixes #1481

**Release note**:

```feature user
Streamlined Kubernetes cluster upgrades for enhanced user experience. We've removed non-eligible versions from the upgrade selection, replacing them with a hint to indicate the existence of more versions. Additionally, only supported versions are now highlighted on the version chip. To aid in version management, chips for deprecated versions will now display in a warning color, alerting users to the need for an upgrade
```


frontend/src/components/ShootVersion/GShootVersion.vue
```diff
@@ -150,11 +147,11 @@ export default {
...
-    k8sPatchAvailable () {
-      if (get(this.availableK8sUpdates, 'patch')) {
-        return true
-      }
-      return false
+    supportedPatchAvailable () {
+      return find(this.availableK8sUpdates?.patch, 'isSupported')
+    },
+    supportedUpgradeAvailable () {
+      return find(this.availableK8sUpdates?.minor, 'isSupported')
```
holgerkoser, line 154:
```suggestion
      return !!find(this.availableK8sUpdates?.minor, 'isSupported')
```

frontend/src/components/ShootVersion/GShootVersion.vue
```diff
@@ -166,14 +163,21 @@ export default {
       return this.availableKubernetesUpdatesForShoot(this.shootK8sVersion, this.shootCloudProfileName)
     },
     tooltipText () {
-      if (this.k8sPatchAvailable) {
+      if (this.kubernetesVersion.isDeprecated) {
```
holgerkoser, line 166: This function should be:
```js
tooltipText () {
      if (this.kubernetesVersion.isDeprecated) {
        return this.shootActionToolTip('Kubernetes version is deprecated')
      }
      if (this.supportedPatchAvailable) {
        return this.shootActionToolTip('Kubernetes patch available')
      } 
      if (this.supportedUpgradeAvailable) {
        return this.shootActionToolTip('Kubernetes upgrade available')
      } 
      if (this.availableK8sUpdates) {
        return this.shootActionToolTip('Updates available')
      }
      return 'Kubernetes version up to date`
    },
```

frontend/src/components/ShootVersion/GShootVersionUpdate.vue
```diff
@@ -183,7 +183,7 @@ export default {
       return filter(this.items, item => !item.notNextMinor)
     },
     hasMoreSupportedVersions () {
-      return filter(this.items, item => item.notNextMinor && item.issu).length
+      return filter(this.items, item => item.notNextMinor && item.isSupported).length
```
holgerkoser, line 186:
```suggestion
      return filter(this.items, item => item.notNextMinor && item.isSupported).length > 0
```
It should return a boolean value

petersutter (approved):
/lgtm with holgers suggestions

holgerkoser (approved):
/lgtm

PR #1686, opened 2024-01-11
Fixed member validation

grolu:
We need to ensure that there is at least an empty validation defined as the fields are visible but disabled in case of `update` dialog.
We also cannot just remove the return as we must not use the `add`  validations for `update` dialog. In this case, the dialog would always return errors upon save like 'user foo already exists'

Fixes #

**Release note**:

```bugfix user
Fixed an issue on the member management page. Update members and service account dialog did not render correctly because of an issue with the input validation
```

### DIFF
--- a/frontend/src/components/Members/GMemberDialog.vue
+++ b/frontend/src/components/Members/GMemberDialog.vue
@@ -240,6 +240,7 @@ export default {
     rules.internalRoles = withFieldName('Member Roles', internalRolesRules)
 
     if (this.isUpdateDialog) {
+      rules.internalName = {}
       return rules
     }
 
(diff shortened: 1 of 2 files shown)


frontend/src/components/Members/GMemberDialog.vue
```diff
@@ -240,6 +240,7 @@ export default {
     rules.internalRoles = withFieldName('Member Roles', internalRolesRules)
 
     if (this.isUpdateDialog) {
+      rules.internalName = {}
```
petersutter, line 243: instead, initialize it already here

```js
const rules = {
  internalName: {}
}
```

holgerkoser: I still do not understand why not just remove the `if (this.isUpdateDialog) {`

grolu: Added comment to make it clear

holgerkoser (approved):
/lgtm

petersutter (approved):
/lgtm

PR #1689, opened 2024-01-12
The author of a github comment can be `null`

holgerkoser:
This pull request addresses a specific behavior observed in the GitHub GraphQL API, where the author field is set to null for comments with an author_association value of 'NONE'. To maintain consistency with the GitHub REST API, which defaults to a user login of 'ghost' under similar circumstances, this change implements a similar fallback mechanism in the GraphQL API. Notably, this update focuses solely on the user login aspect and does not include a fallback for the avatar_url, as our dashboard does not utilize this information.

Fixes #

**Release note**:

```bugfix user
NONE
```


grolu (approved):
/lgtm

PR #1697, opened 2024-01-17
Streamlined Action Dialogs

grolu:
This PR addresses serveral issues with the Action Dialogs
Main Problem was that we had two different types of Action Button Dialogs one combined and one split in two parts. I think we mainly did this as a workaround because the v-menu content for the actions on the cluster list gets cleaned up when an action is triggered. I solved this by using the `eager` prop which forces vuetify to always render the content (including the dialogs) which is required as the dialog is visible after the menu closed. I measured the performance and this change has no impact - which makes sense, as we anyway rendered the dialogs but outside of the menu....
So I managed to get rid of this second Action Button Dialog type and all components involved.

Moreover I fixed several other things, this is the whole list:
- Added text button capabilities to `GActionButton`
- Make `GActionButtonDialog` use `GActionButton`
- Removed `GShootActionButton` (in favor of `GActionButtonDialog`)
- Removed `GShootActionDialog` (in favor of `GActionButtonDialog`)
- Make shoot list action v-menu eager (to avoid having to split up components)
- Always use shootActionToolTip for Action Dialogs (fixes missing tooltip descriptions) for some disabled cluster action buttons
- Fixed tooltip line breaks for some button tooltips (removed wrong class)
- Fixed notify/alert message for trigger maintenance
- Fixed maintenance precondition constraint message was always shown on maintenance start dialog
- moved to configure maintenance dialog (as we have it for manage hibernation schedule), fixed so that it is only shown in case of precondition not satisfied
- fixed wrong maintenance precondition constraint message title (probably c/p error from hibernation message)
- disable start maintenance button in case precondition not satisfied
- Removed unused/duplicated `GReconcileStart` component
- Fixed copy icon layout (rm `small` as it is no longer required)
- Show update k8s button only if user can update this also hides it in case of missing permission
- Cleaned up `GDialog` an `GActionButtonDialog` slots. Removed unnecessary conditional rendering and renamed slots to make usage clearer. Removed unused error message slot

Fixes #

**Release note**:

```bugfix user
Action dialogs fixes:
- We've addressed an issue where tooltips were missing for some disabled cluster action buttons. Users will now see helpful tooltips providing more context on why these actions are currently unavailable
- Fixed issues with line breaks in the tooltips for certain buttons, ensuring that the tooltip text is now displayed correctly and is easier to read
- Resolved problems related to the display of precondition constraint failure messages during maintenance operations
- Fixed an issue where some action buttons were visible to users who did not have permission to perform those actions. This update ensures that the interface correctly reflects available actions based on user rights
```


petersutter:
Screenshot 2024-01-24 at 16 43 25

not sure if this has changed with this PR but the button does not span the whole width as it is with the current dashboard release

petersutter (approved):
/lgtm

frontend/src/components/ShootMaintenance/GShootActionMaintenanceStart.vue
```diff
@@ -147,28 +108,20 @@ export default {
         return
       }
 
-      this.notify({
-        text: `Maintenance scheduled for ${this.shootName}`,
-        type: 'success',
-        position: 'bottom right',
-        duration: 5000,
+      this.setSuccess({
```
holgerkoser, line 111: Set success message as string
```js
 this.setSuccess(`Maintenance scheduled for ${this.shootName}`)
```
and remove the injected `notify` method!

frontend/src/components/ShootHibernation/GShootActionChangeHibernation.vue
```diff
@@ -5,49 +5,44 @@ SPDX-License-Identifier: Apache-2.0
...
+import GActionButtonDialog from '@/components/dialogs/GActionButtonDialog.vue'
 
 import { shootItem } from '@/mixins/shootItem'
 import { errorDetailsFromError } from '@/utils/error'
 
 export default {
   components: {
-    GShootActionButton,
-    GShootActionDialog,
+    GActionButtonDialog,
```
holgerkoser, line 45: Here we still use `notify`. In an other component we have replaced it with `setSuccess`. What is the way to go?

frontend/src/components/GShootActionRotateCredentials.vue
```diff
@@ -438,7 +403,7 @@ export default {
         return
       }
 
-      this.setAlert({
+      this.setSuccess({
```
holgerkoser, line 406: Set success message as string directly.

frontend/src/components/GShootActionReconcileStart.vue
```diff
@@ -141,28 +106,19 @@ export default {
         return
       }
 
-      this.setAlert({
+      this.setSuccess({
```
holgerkoser, line 109: Set success as string directly

frontend/src/components/GActionButton.vue
```diff
@@ -88,4 +92,32 @@ const { disabled, icon, text, color, size, tooltip, to } = toRefs(props)
...
+  return props.density
+    ? props.density
+    : defaultDensity
+})
+
 </script>
+
+<style>
+.g-action-button {
+  width: fit-content !important;
```
holgerkoser, line 117: We do not need important here because the div has no other styles.

frontend/src/components/dialogs/GDialog.vue
```diff
@@ -30,34 +30,27 @@ SPDX-License-Identifier: Apache-2.0
         </v-toolbar-title>
       </v-toolbar>
       <div>
-        <slot name="top" />
+        <slot name="header" />
```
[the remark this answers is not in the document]
grolu: so you suggest re-adding conditional rendering?

frontend/src/components/ShootHibernation/GShootActionChangeHibernation.vue
grolu: we use setSuccess / setAlert everywhere so I changed the remaining component to also use setSuccess

frontend/src/components/dialogs/GDialog.vue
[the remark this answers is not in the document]
holgerkoser: yes

PR #1702, opened 2024-01-30
Spinner: ensure interval is started only once

petersutter:
There is a case where the spinner is started twice and one interval leaks https://github.com/gardener/dashboard/blob/5c69e731ea0d7f7f3b41dfbc7e9496e21a12b128/frontend/src/lib/terminal.js#L219
This PR ensures that only one spinner interval runs at a time.

Fixes #

**Release note**:

```bugfix user
Terminal: fixed an issue where the loading spinner would not disappear, even though the terminal pod is `running` and the terminal connection shows as `connected`
```


grolu (approved):
/lgtm

grolu (approved):
/lgtm

holgerkoser (approved):
/lgtm

PR #1665, opened 2023-12-19
Force Delete Cluster

grolu:
**What this PR does / why we need it**:

Screenshot 2024-02-01 at 12 48 35

Screenshot 2024-02-02 at 18 39 58

Screenshot 2024-02-01 at 13 41 43

**Which issue(s) this PR fixes**:
Fixes #1607

**Special notes for your reviewer**:

**Release note**:

```feature user
Implemented Force Deletion of Clusters
If a cluster is stuck in deletion, users can force delete clusters via the dashboard. This is only possible when the `ShootForceDeletion` featuregate in the apiserver is turned on
```

```bugfix user
Fixed an issue that caused `dark` theme color to be applied to some alert messages
```

```feature user
Improved cluster messages
- Show icons always next to title on cluster details page
- Fixed overall severity sometimes not correct
```

PR #1713, opened 2024-02-08
Removed force delete feature flag from dashboard configuration

grolu:
Removed force delete feature flag from dashboard configuration.
It is not required in most cases, as the feature will be enabled in most landscapes and the feature gate will be removed in future gardener versions. Moreover, it is acceptable to run into an error in those rare situations where the feature is disabled as this feature is only available in edge error scenarios anyway

Fixes #

No release note as PR feature has not yet been released
Adjusted release note in original PR: https://github.com/gardener/dashboard/pull/1665

**Release note**:

```bugfix user
NONE
```


petersutter (approved):
/lgtm

holgerkoser (approved):
/lgtm

PR #1719, opened 2024-02-16
parse numerical abbreviations

### LINKED ISSUE #1716: Dashboard cannot display quotas that are numerical abbreviations
**What happened**:
Quotas using numerical abbreviations e.g. 1k lead to NaN errors in Gardener Dashboard usage statistics.

**Environment**:

grolu:

Fixes #1716

**Release note**:

```bugfix user
Fixed an issue that caused quotas using numerical abbreviations (e.g., '1k') to not be displayed correctly in the Gardener Dashboard
```

### DIFF
--- a/frontend/src/utils/index.js
+++ b/frontend/src/utils/index.js
@@ -644,25 +644,14 @@ export function omitKeysWithSuffix (obj, suffix) {
   return omit(obj, keys)
 }
 
-export function parseNumericalAbbreviation (abbreviatedNumber) {
-  const regex = /(\d+(?:\.\d+)?)(k|M|B)?/i
-  const match = abbreviatedNumber.match(regex)
-
-  if (!match) {
+export function parseNumberWithMagnitudeSuffix (abbreviatedNumber) {
+  const [, number, suffix] = /^(\d+(?:\.\d*)?)([kmbt]?)$/i.exec(abbreviatedNumber) ?? []
+  if (!number) {
+    logger.error(`Could not parse number with suffix ${abbreviatedNumber} as it does not match regex ^(\\d+(?:\\.\\d*)?)([kmbt]?)$`)
     return null
   }
 
-  const number = parseFloat(match[1])
-  const suffix = match[2]
-
-  switch (suffix?.toLowerCase()) {
-    case 'k':
-      return number * 1e3
-    case 'm':
-      return number * 1e6
-    case 'b':
-      return number * 1e9
-    default:
-      return number
-  }
+  const suffixFactors = { k: 1e3, m: 1e6, b: 1e9, t: 1e12 }
+  const factor = suffixFactors[suffix?.toLowerCase()] ?? 1
+  return Number(number) * factor
 }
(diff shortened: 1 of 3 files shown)


frontend/src/utils/index.js
```diff
@@ -643,3 +643,26 @@ export function omitKeysWithSuffix (obj, suffix) {
   const keys = Object.keys(obj).filter(key => key.endsWith(suffix))
   return omit(obj, keys)
 }
+
+export function parseNumericalAbbreviation (abbreviatedNumber) {
```
holgerkoser, line 647: Rename to `parseAbbreviatedNumber` or `parseNumberWithMagnitudeSuffix`.

frontend/src/utils/index.js
```diff
@@ -643,3 +643,26 @@ export function omitKeysWithSuffix (obj, suffix) {
...
+  const regex = /(\d+(?:\.\d+)?)(k|M|B)?/i
+  const match = abbreviatedNumber.match(regex)
+
+  if (!match) {
+    return null
+  }
+
+  const number = parseFloat(match[1])
+  const suffix = match[2]
```
holgerkoser, line 656: If the input is `x1` we return 1.
If the input is `1.` we return null
if the input is `1km` we return 1000 
Wouldn't it be better to return NaN if parsing fails. The reason why we return `null` is that we want to use it directly in `min` and `max` functions. That only works with null. 
```suggestion
  const [, number, suffix] = /^(\d+(?:\.\d*)?)([kmbt]?)$/i.exec(abbreviatedNumber) ?? []
  if (!number) {
    return null
  }
```

frontend/src/utils/index.js
```diff
@@ -643,3 +643,26 @@ export function omitKeysWithSuffix (obj, suffix) {
...
+  switch (suffix?.toLowerCase()) {
+    case 'k':
+      return number * 1e3
+    case 'm':
+      return number * 1e6
+    case 'b':
+      return number * 1e9
+    default:
+      return number
+  }
```
holgerkoser, line 667: The regex already makes sure the number has correct format.
```suggestion
  const suffixFactors = { k: 1e3, m: 1e6, b: 1e9, t: 1e12 }
  const factor = suffixFactors [suffix?.toLowerCase()] ?? 1
  return Number(number) * factor
```

frontend/src/utils/index.js
```diff
@@ -644,25 +644,14 @@ export function omitKeysWithSuffix (obj, suffix) {
...
 
-export function parseNumericalAbbreviation (abbreviatedNumber) {
-  const regex = /(\d+(?:\.\d+)?)(k|M|B)?/i
-  const match = abbreviatedNumber.match(regex)
-
-  if (!match) {
+export function parseNumberWithMagnitudeSuffix (abbreviatedNumber) {
+  const [, number, suffix] = /^(\d+(?:\.\d*)?)([kmbt]?)$/i.exec(abbreviatedNumber) ?? []
+  if (!number) {
+    logger.error(`Could not parse number with suffix ${abbreviatedNumber} as it does not match regex ^(\\d+(?:\\.\\d*)?)([kmbt]?)$`)
```
holgerkoser, line 650:
```suggestion
    logger.error(`Failed to parse ${abbreviatedNumber} because it doesn't follow the required format: a number optionally with a decimal, followed by an optional magnitude suffix ('k', 'm', 'b', 't').`)
```

holgerkoser (approved):
/lgtm
Eventuell noch die error message ohne regex

petersutter (approved):
/lgtm

PR #1728, opened 2024-03-05
Do not double quote string in yaml config example

petersutter:
This PR addresses an issue with the handling of YAML string config values. Previously, these values were double quoted, which caused problems when using Windows paths. The backslash character in the Windows path needed to be escaped due to the double quotes, which was not intuitive for users.

To resolve this, we have updated the example config value for the `kubeconfig` key so that it is no longer double quoted. This change simplifies the use of Windows paths and eliminates the need for users to manually escape the backslash character.

Fixes #

**Release note**:

```bugfix user

```


grolu (approved):
/lgtm

PR #1823, opened 2024-04-18
Create `Lease` object during runtime

### LINKED ISSUE #1806: Remove `Lease` from Helm chart
https://github.com/gardener/dashboard/blob/master/charts/gardener-dashboard/charts/application/templates/dashboard/lease-github-webhook.yaml should be removed, and the https://github.com/gardener/dashboard/blob/c70643750c24506092fdec4f587913f60e864db4/backend/lib/github/webhook/handler.js#L17-L31 code should create the `Lease` when it is not found.

**Why is this needed**:
Deploying the Helm chart multiple times always resets the `Lease` spec which is undesired.

petersutter:
The `Lease` object is no longer included in the Helm chart. Instead, it is now created dynamically during runtime if it does not already exis

Fixes #1806

**Release note**:

```other developer
The `Lease` object is no longer included in the Helm chart. Instead, it is now created dynamically during runtime if it does not already exist
```


grolu (approved):
/lgtm

holgerkoser (approved):
/lgtm

PR #1838, opened 2024-04-23
Migrate Inter-Component Communication to Centralized State Management

### LINKED ISSUE #975: Improve Frontend Component Quality
Currently, most of the frontend components do not fulfil our quality standards due to several reasons
- Fuzzy interface (properties)
- (deep) usage of storage dependencies
- interaction through events and busses
- ...

The reasons above make components hard to test and to maintain. We need to work through our components and continuously improve. Maybe it makes also sense to collect components here and flag if they already have been reworked to track progress of this issue.

### LINKED ISSUE #1621: Store Refactoring / Cleanup
We need to refactor the store so that we do not have methods with lots of parameters (e.g. cloudprofile store).
We should set the current context of the store and all those methods should be getters.

**Why is this needed**:

holgerkoser:
This PR does a lot of code cleanup:
- Removal of `userInteractionBus` based eventing between components and use a central storage for the shoot modification context
- Removal of all mixins and replace them by composables
- Encapsulate the shoot editor state in a composable
- Removal of `asyncRef` composable and inter component communication. Use the store or injected state instead.
- Remove async yaml plugin and direclty use `js-yaml`
- Fixed Tabstrip selection on navigation abort https://github.com/gardener/dashboard/pull/1838/commits/31c51c883ba27ac1235a3a16e610fe5674d77546
- Do not use `v-tooltip` activator slot inside of `v-for` loops https://github.com/gardener/dashboard/pull/1838/commits/525f7228136e7d6b58959a0fe59b197a4cded11b
- Fixed resize splitpanes warning https://github.com/gardener/dashboard/pull/1838/commits/5fd6cc8db5d96b97043822cb2b209ec28768ea7d
... 

#### Shoot manifest dependency graph
```mermaid
flowchart LR
    root((root)) -->  dns & maintenanceAutoUpdate & maintenanceTimeWindow & providerType & workerless
    workerless --->  specTemplate
    providerType --> specTemplate & enableStaticKubeconfig & cloudProfileName
    cloudProfileName --> networkingType & kubernetesVersion & secretBindingName & region & loadBalancerClasses & projectID & firewallImage
    secretBindingName --> purpose  
    region --> workers & loadBalancerProviderName & partitionID & floatingPoolName 
    purpose --> hibernationShedules
    partitionID --> firewallSize & firewallNetworks
    workers --> addons

subgraph specTemplate
    direction LR
    infrastructureConfig
    controlPlaneConfig 
    networking 
    kubernetes
 end 
```
Fixes #1621 Fixes #975

**Release note**:

```other developer
Refactored store to use context and getters, and improved frontend components by eliminating fuzzy interfaces and direct store data mapping for better maintainability and quality.
```


frontend/src/components/ShootAccessRestrictions/GAccessRestrictions.vue
```diff
@@ -6,195 +6,115 @@ SPDX-License-Identifier: Apache-2.0
 
 <template>
   <div
-    v-if="definitions"
+    v-if="accessRestrictionDefinitions"
```
grolu, line 9:
```suggestion
    v-if="accessRestrictionDefinitions?.length"
```
Array is empty in case region without AR support gets selected. v-else div does not show up in this case

frontend/src/composables/useShootContext/index.js
```diff
@@ -0,0 +1,1261 @@
...
+  }
+
+  /* region */
+  const region = computed({
+    get () {
+      return get(manifest.value, 'spec.region')
+    },
+    set (value) {
+      set(manifest.value, 'spec.region', value)
+      resetProviderWorkers()
```
grolu, line 256:
```suggestion
      if (!workerless.value) {
        resetProviderWorkers()
      }
```
We should not reset the workers in case workerless is enabled, as this would make the cluster have workers again.
However, if we don't do it here we need to ensure it is done somewhere further up in the call hierarchy, eg. by adding `resetProviderWorkers()` to the `cloudProfileName` computed property setter.
Not sure if this is the right way to do it though

frontend/src/components/ShootWorkers/GWorkerInputGeneric.vue
```diff
@@ -512,22 +476,13 @@ export default {
...
     onInputZones () {
       this.v$.selectedZones.$touch()
       this.v$.worker.maximum.$touch()
     },
     resetWorkerMachine () {
-      this.worker.machine.type = get(head(this.machineTypes), 'name')
+      const machineType = head(this.machineTypes)
+      this.worker.machine.type = get(machineType, 'name')
       const machineImage = head(this.machineImages)
       this.worker.machine.image = pick(machineImage, ['name', 'version'])
```
grolu, line 487:
```suggestion
      this.worker.machine.type = machineType?.name
      const machineImage = this.defaultMachineImageForCloudProfileNameAndMachineType(
        this.cloudProfileName,
        machineType,
      )
```
This is not a new bug, also broken in ucrrent master. We need to use the store function to get the correct machine image. Simply using the first one may select a preview / deprecated version

frontend/src/composables/useShootContext/index.js
holgerkoser: I think we need to have an explicit property for the `workerless` state and not derive it from the existing numberOfWorkers. I think I understood the problem.

frontend/src/components/ShootAccessRestrictions/GAccessRestrictions.vue
holgerkoser: ✔️

petersutter (approved):
/lgtm

grolu (approved):
/lgtm

PR #1856, opened 2024-05-07
Support previous session secret, enabling seamless session secret rotation

### LINKED ISSUE #1790: Support multiple session secrets to allow rotation
Support for configuring multiple secrets for signing and encrypting user sessions in the Gardener dashboard.

The dashboard backend should accept an array of secrets instead of a single secret. When verifying session cookies, each secret should be tried in order until a valid signature is found. New session cookies should always be signed with the first secret in the list.

Update the dashboard Helm chart and documentation with information on how to configure multiple secrets. 

The secrets should be passed as environment variables to the dashboard deployment. For example:

```yaml
env:
- name: SESSION_SECRET_1
  valueFrom:
    secretKeyRef:
      name: dashboard-session-secret-1
      key: keyValue
- name: SESSION_SECRET_2  
  valueFrom:
    secretKeyRef:
      name: dashboard-session-secret-2
      key: keyValue
```

**Why is this needed**:
Currently, the Gardener dashboard only supports a single secret (passed via the `SESSION_SECRET` environment variable) for signing and encrypting user sessions. This makes it difficult to rotate secrets without invalidating all active user sessions.

Supporting multiple secrets would allow for graceful secret rotation by:
1. Adding a new secret to the beginning of the list (e.g. `SESSION_SECRET_1`)
2. Waiting for old sessions using the previous secret to expire
3. Removing the old secret (e.g. `SESSION_SECRET_2`) from the end of the list

This improves security by enabling regular session secret rotation without disrupting active sessions. It aligns the dashboard session management with best practices.

By enhancing the dashboard to support this functionality, we can provide Gardener operators with a more robust and secure user session implementation that allows secrets to be cycled gracefully without impacting the user experience.

petersutter:
With this PR, the dashboard supports a previous session secret. It allows for a seamless rotation of the session secret by supporting both the current and previous secrets.

Example values.yaml file: 
```yaml
global:
  dashboard:
    sessionSecret: <secret>
    sessionSecretPrevious: <previous-secret>
```

If you are not using the provided Helm chart to deploy the Gardener Dashboard (e.g. gardener operator), ensure secrets are mounted to the filesystem at the appropriate path:

- Previous session secret: `/etc/gardener-dashboard/secrets/session/sessionSecretPrevious`

Fixes #1790 

**Release note**:

```feature operator
The dashboard supports a previous session secret. It allows for a seamless rotation of the session secret by supporting both the current and previous secrets. When installed using the helm chart, provide `Values.global.dashboard.sessionSecretPrevious`. Set this value to the previous sessionSecret during secret rotation, and leave it empty otherwise.
```


grolu (approved):
/lgtm

backend/lib/security/jose.js
```diff
@@ -46,8 +46,13 @@ function decodeState (state) {
   }
 }
 
-module.exports = sessionSecret => {
-  const symetricKey = importSymmetricKey(sessionSecret)
+module.exports = sessionSecrets => {
+  if (!sessionSecrets || !sessionSecrets.length) {
+    throw new Error('No session secrets provided')
+  }
```
holgerkoser, line 52:
```suggestion
  if (!sessionSecrets?.length) {
    throw new Error('No session secrets provided')
  }
```

backend/lib/config/gardener.js
```diff
@@ -205,6 +210,10 @@ module.exports = {
       assert.ok(_.get(config, path), `Configuration value '${path}' is required`)
     })
 
+    _.set(config, 'sessionSecrets', [
+      config.sessionSecret,
+      ...(config.sessionSecretPrevious ? [config.sessionSecretPrevious] : [])
+    ])
```
holgerkoser, line 216:
```suggestion
    const sessionSecrets = [config.sessionSecret]
    if (config.sessionSecretPrevious) {
      sessionSecrets.push(config.sessionSecretPrevious)
    }
    _.set(config, 'sessionSecrets', sessionSecrets)
```

backend/lib/security/jose.js
```diff
@@ -69,8 +74,16 @@ module.exports = sessionSecret => {
...
+    async verify (token, options) {
+      let lastError
+      for (const sessionSecret of sessionSecrets) {
+        try {
+          return await jwtVerify(token, sessionSecret, options)
+        } catch (error) {
+          lastError = error
+        }
+      }
+      throw lastError
```
holgerkoser, line 86: It could be missleading to return the last error. Assume both secrets are in general invalid I think the first error is more significant. I would return a new Error.
```
async verify (token, options) {
  let firstError
  for (const sessionSecret of sessionSecrets) {
    try {
      return await jwtVerify(token, sessionSecret, options)
    } catch (err) {
      firstError ??= err
    }
  }
  throw new Error('Token cannot be verified with any of the session secrets')
}
```

backend/test/security.spec.js
```diff
@@ -9,22 +9,31 @@
 const _ = require('lodash')
 
 describe('security', function () {
+  const joseModule = require('../lib/security/jose')
```
holgerkoser, line 12: rename to `createJose`

backend/lib/security/jose.js
petersutter: ✅

backend/lib/config/gardener.js
petersutter: yeah I thought that it might make sense to have this code better readable, I will change it.. ✅

backend/test/security.spec.js
petersutter: ✅

holgerkoser (approved):
/lgtm

holgerkoser (approved):
/lgtm

PR #1877, opened 2024-05-23
Do not fail if tickets could not be fetched

### LINKED ISSUE #1876: Error Handling in `fetchShoots` Function: Promise.all Fails When GitHub API Request Fails
**What happened**:
https://github.com/gardener/dashboard/blob/77788079d465a677526df6b23fcaea87f7cc0143/frontend/src/store/shoot/index.js#L322-L334
The `fetchShoots` function uses `Promise.all` to concurrently fetch data from two different APIs: `api.getShoots` and `api.getIssues`. However, if the GitHub API request (`api.getIssues`) fails, the entire `Promise.all` call fails, causing the `fetchShoots` function to fail as well. This issue needs to be addressed to ensure that the function can handle errors gracefully and still return the data from the successful API call.

Use `Promise.allSettled` instead of `Promise.all`

**Environment**:

petersutter:
- Do not fail if tickets could not be fetched
- fix margins when displaying multiple tickets

Fixes #1876

**Release note**:

```bugfix user

```

### DIFF
--- a/frontend/src/components/GTicketsCard.vue
+++ b/frontend/src/components/GTicketsCard.vue
@@ -15,7 +15,7 @@ SPDX-License-Identifier: Apache-2.0
         :key="ticket.metadata.issueNumber"
         :ticket="ticket"
       />
-      <div class="d-flex align-center justify-center mt-4">
+      <div class="d-flex align-center justify-center">
         <v-btn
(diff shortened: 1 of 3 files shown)


holgerkoser (approved):
/lgtm

PR #1888, opened 2024-05-28
Configure extension DNS settings in the `spec.extensions` section

### LINKED ISSUE #1723: Adapt handling of DNS Providers
Add new separate dialogs for managing primary DNS provider and the additional DNS providers for the shoot-dns-service extension. These new dialogs should only be available, if the condition for using the new behaviour is fulfilled (see below).
With the new behaviour, the section `.spec.dns.providers` will have either `0` or `1` provider. If there is a provider, it must be the `primary` one. It will have no includes/excludes for domains and zones. All others will be cleaned up by the Gardenlet.
For existing clusters, the dashboard UI for this primary provider can only update type and secret values.
For new clusters, the dashboards UI could use a single dialog to create the primary provider and an provider for the shoot-dns-service with the same inputs. In this case, includes/excludes for domains and zones still make sense. After shoot creation, the UIs for primary provider and shoot-dns-service providers should be completely separate.

For managing the DNS providers of the shoot-dns-service, a dialog similar to the current one can be used. Only the sections used are changing to `.spec.extensions[@type="shoot-dns-service"].providerConfig` and `.spec.resources[]` for the secret.

Example for an additional DNS provider for the shoot-dns-service:
```yaml
spec:
  extensions:
  - type: shoot-dns-service
    providerConfig:
      apiVersion: service.dns.extensions.gardener.cloud/v1alpha1
      kind: DNSConfig
      providers:
        - domains:
            include:
              - wl-owndns.dnstest.dev.k8s.ondemand.com
          secretName: shoot-dns-service-my-aws-route53-secret
          type: aws-route53
  resources:
  	- name: shoot-dns-service-my-aws-route53-secret
  	  resourceRef:
  		kind: Secret
  		name: my-aws-route53-secret
  		apiVersion: v1
```

**Why is this needed**:
Gardener wants to cleanup some DNS legacy fields in the shoot manifest, see [[API] Align DNS provider handling](https://github.com/gardener/gardener/issues/9176). As a consequence, non-primary DNS providers for the shoot-dns-service must be managed in its extension providerConfig, i.e. `.spec.extensions[@type="shoot-dns-service"].providerConfig` and the DNS secrets are referenced resources, i.e. at `.spec.resources[]`.
Currently, the DNS providers are managed at `.spec.dns.providers` and copied to the extension providerConfig section by the mutating webhook of the shoot-dns-service. Both shoot-dns-service webhook and Dashboard UI must change their behaviour using the same condition.

The condition for using the new behaviour should be either `.spec.extensions[@type="shoot-dns-service"].syncProvidersFromShootSpecDNS==false` or Gardener Version (`.status.gardener.version`) >= `v1.XX`(concrete version not decided yet).

grolu:

Fixes #1723

**Release note**:

```feature user
Enhanced DNS configuration logic to configure extension DNS settings in the `spec.extensions` section
```
```breaking user
DNS configuration is now maintained under `spec.extensions`. When modifying the DNS configuration of an existing shoot, the settings will be migrated to this new location, and the `syncProvidersFromShootSpecDNS` flag will be set to false
```


frontend/src/composables/useShootContext/index.js
```diff
@@ -124,10 +133,29 @@ export function createShootContextComposable (options = {}) {
...
+
+    if (index === -1) {
+      extensions.push(extension)
+      return
+    }
+    if (extension) {
+      extensions[index] = extension
+    } else {
+      extensions.splice(index, 1)
+    }
```
petersutter, line 151: Somehow I overlooked the return statement. Either add a newline between the two if-cases or have it all in one if-else block
```go
    if (index === -1) {
      extensions.push(extension)
      return
    }

    if (extension) {
      extensions[index] = extension
    } else {
      extensions.splice(index, 1)
    }
```

```go
    if (index === -1) {
        extensions.push(extension);
    } else if (extension) {
        extensions[index] = extension;
    } else {
        extensions.splice(index, 1);
    }
```

frontend/src/composables/useShootContext/index.js
```diff
@@ -904,12 +970,9 @@ export function createShootContextComposable (options = {}) {
             include: includeZones = [],
           } = {},
         } = provider
-        if (primary) {
-          dnsPrimaryProviderId.value = id
-        }
         let readonly = false
         if (!isNewCluster.value) {
-          const secret = findDnsProviderSecret(type, secretName)
+          const secret = findExtensionDnsProviderSecret(type, utils.getResourceRefName(initialManifest.value.spec?.resources, secretName))
```
petersutter, line 975: extract `utils.getResourceRefName(initialManifest.value.spec?.resources, secretName)` to variable

frontend/src/composables/useShootContext/index.js
```diff
@@ -921,22 +984,49 @@ export function createShootContextComposable (options = {}) {
           id,
           type,
           secretName,
+          secretRefName: utils.getResourceRefName(initialManifest.value?.spec?.resources, secretName),
```
petersutter, line 987: inconsistent usage of `?` (`initialManifest.value?`), see also https://github.com/gardener/dashboard/blob/5b26d8b50377e0a0254c08d70726ffb714e60d95/frontend/src/composables/useShootContext/index.js#L975
https://github.com/gardener/dashboard/blob/5b26d8b50377e0a0254c08d70726ffb714e60d95/frontend/src/composables/useShootSpec.js#L135

frontend/src/composables/useShootSpec.js
```diff
@@ -119,11 +121,20 @@ export function useShootSpec (shootItem, options = {}) {
...
+    return !!shootDnsPrimaryProvider.value
   })
 
-  const shootDnsProviders = computed(() => {
-    return get(shootSpec.value, 'dns.providers')
+  const shootDnsPrimaryProvider = computed(() => {
+    return find(shootSpec.value.dns?.providers, { primary: true })
+  })
+  const shootExtensionDnsProviders = computed(() => {
+    const extensionDns = find(shootSpec.value.extensions, { type: 'shoot-dns-service' })
```
petersutter, line 131: we typically use the list-style for the lodash `find` function, instead of the object
however I must admit that the object style feels more natural (why do we use the list style again? :D)

frontend/src/composables/useShootSpec.js
```diff
@@ -119,11 +121,20 @@ export function useShootSpec (shootItem, options = {}) {
...
+    return map(extensionDns?.providerConfig.providers, provider => {
```
petersutter, line 132: is the `providerConfig` always set?

frontend/src/composables/useShootSpec.js
```diff
@@ -119,11 +121,20 @@ export function useShootSpec (shootItem, options = {}) {
...
-    return some(shootDnsProviders.value, 'primary')
+    return !!shootDnsPrimaryProvider.value
   })
 
-  const shootDnsProviders = computed(() => {
-    return get(shootSpec.value, 'dns.providers')
+  const shootDnsPrimaryProvider = computed(() => {
+    return find(shootSpec.value.dns?.providers, { primary: true })
+  })
+  const shootExtensionDnsProviders = computed(() => {
```
petersutter, line 130: should this computed property not just return the providers instead of also mapping the `secretRefName` into it? It's only used in one place, maybe it reduces complexity if we look up the `secretRefName` at the place where we need it?

frontend/src/sass/main.scss
```diff
@@ -128,7 +128,7 @@
 
 .large-input {
   max-width: 600px;
-  min-width: 300px;
+  min-width: 180px;
```
petersutter, line 131: what is this change about?

frontend/src/composables/useShootSpec.js
```diff
@@ -119,11 +121,20 @@ export function useShootSpec (shootItem, options = {}) {
   })
 
   const isCustomShootDomain = computed(() => {
-    return some(shootDnsProviders.value, 'primary')
+    return !!shootDnsPrimaryProvider.value
   })
 
-  const shootDnsProviders = computed(() => {
-    return get(shootSpec.value, 'dns.providers')
+  const shootDnsPrimaryProvider = computed(() => {
```
holgerkoser, line 127: We do you delete this computed property and not do something like this:
```js
 const shootDnsProviders = computed(() => {
    return get(shootSpec.value, 'dns.providers')
  })
  
  const shootDnsPrimaryProvider = computed(() => {
    return find(shootDnsProviders.value, 'primary')
  })
  
 const isCustomShootDomain = computed(() => {
    return !!shootDnsPrimaryProvider.value
  })
```

frontend/src/utils/index.js
```diff
@@ -685,6 +685,11 @@ export function normalizeVersion (version) {
   }
 }
 
+export function getResourceRefName (resources, resourceName) {
```
holgerkoser, line 688: I think this should not be something for general purpose utils. From my point of view it would be better nd more readable to reuse the list of resources and duplicate accessing the `resorceRef.name`.

frontend/src/store/gardenerExtension.js
```diff
@@ -43,23 +43,22 @@ export const useGardenerExtensionStore = defineStore('gardenerExtension', () =>
     list.value = response.data
   }
 
+  const isDnsServiceExtensionDeployed = computed(() => {
```
holgerkoser, line 46:
```suggestion
  const hasDnsServiceExtension = computed(() => {
```

frontend/src/components/ShootDns/GDnsConfiguration.vue
```diff
@@ -59,14 +61,17 @@ export default {
...
       setShootManifest,
+      dns,
+      extensionDns,
       componentKey,
     }
   },
   methods: {
     async onConfigurationDialogOpened () {
       this.componentKey = uuidv4() // force re-render
+      this.setShootManifest(this.shootItem)
```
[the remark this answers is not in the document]
holgerkoser: No. Must be removed!

frontend/src/components/ShootDns/GManageDns.vue
```diff
@@ -17,87 +30,124 @@ SPDX-License-Identifier: Apache-2.0
...
-            label="Primary DNS Provider"
-            clearable
-            :disabled="!isNewCluster"
-            variant="underlined"
-            @blur="v$.dnsPrimaryProvider.$touch()"
-            @update:model-value="v$.dnsPrimaryProvider.$touch()"
+        <template v-if="primaryProviderVisible">
+          <v-col cols="3">
+            <v-select
+              v-model="primaryDnsProviderType"
```
holgerkoser, line 38:
```suggestion
              v-model="v$.primaryDnsProviderType.$model"
```

frontend/src/components/ShootDns/GManageDns.vue
```diff
@@ -17,87 +30,124 @@ SPDX-License-Identifier: Apache-2.0
...
+              color="primary"
+              :items="dnsProviderTypesWithPrimarySupport"
+              :error-messages="getErrorMessages(v$.primaryDnsProviderType)"
+              label="Dns Provider Type"
+              :disabled="!isNewCluster"
+              :persistent-hint="!isNewCluster"
+              :hint="typeHint"
+              variant="underlined"
+              @change="v$.primaryDnsProviderType.$touch()"
+              @blur="v$.primaryDnsProviderType.$touch()"
```
holgerkoser, line 48: Do we really need blur

frontend/src/components/ShootDns/GManageDns.vue
```diff
@@ -110,53 +160,142 @@ import {
...
       return this.isNewCluster
         ? 'External available domain of the cluster'
         : 'Domain cannot be changed after cluster creation'
     },
+    domainCheckboxHint () {
+      return this.isNewCluster
+        ? undefined
+        : this.customDomainEnabled
+          ? 'Custom domain cannot be disabled after cluster creation'
+          : 'Custom domain cannot be enabled after cluster creation'
```
holgerkoser, line 230:
```suggestion
      const enabledOrDisabled =  this.customDomainEnabled ? 'disabled' : 'enabled'
      return this.isNewCluster
        ? ''
        : `Custom domain cannot be ${enabledOrDisabled} after cluster creation`
```

frontend/src/components/ShootDns/GManageDns.vue
```diff
@@ -110,53 +160,142 @@ import {
...
+    },
+    typeHint () {
+      return this.isNewCluster
+        ? undefined
```
holgerkoser, line 234:
```suggestion
        ? ''
```

frontend/src/composables/useShootContext/index.js
```diff
@@ -124,10 +133,29 @@ export function createShootContextComposable (options = {}) {
     return normalizeManifest(object)
   })
 
+  function setExtension (shootManifest, type, extension) {
+    if (!Array.isArray(shootManifest?.spec?.extensions)) {
+      set(shootManifest, 'spec.extensions', [])
+    }
+    const extensions = shootManifest.spec.extensions
+    const index = findIndex(extensions, { type })
```
holgerkoser, line 141:
```suggestion
    const index = findIndex(extensions, ['type', type])
```
Why do we always mix usages of propery shorthands? The object shorthand results in https://lodash.com/docs/4.17.15#matches while the property shorthand results in https://lodash.com/docs/4.17.15#matchesProperty. The second has a much better performance and is from my personal point of view much better readable.

frontend/src/composables/useShootContext/index.js
```diff
@@ -124,10 +133,29 @@ export function createShootContextComposable (options = {}) {
     return normalizeManifest(object)
   })
 
+  function setExtension (shootManifest, type, extension) {
```
holgerkoser, line 136: The function should be declared outsine of the composable scope.

frontend/src/composables/useShootContext/index.js
holgerkoser: Yes it is inconsistent and we do the same thing twice. I would prefer to use lodash get if we are unsure about the fact if an object is nullish or not.

backend/lib/services/shoots.js
```diff
@@ -227,14 +227,64 @@ exports.replaceControlPlaneHighAvailability = async function ({ user, namespace,
...
+  } else {
+    patchOperations.push({
+      op: 'add',
+      path: '/spec/extensions/-',
+      value: extensionDns
+    })
   }
 
-  return client['core.gardener.cloud'].shoots.mergePatch(namespace, name, payload)
+  return client['core.gardener.cloud'].shoots.jsonPatch(namespace, name, patchOperations)
```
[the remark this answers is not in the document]
grolu: done

frontend/src/components/ShootDns/GDnsConfiguration.vue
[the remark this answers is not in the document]
grolu: ok, leftover, removed

frontend/src/components/ShootDns/GManageDns.vue
grolu: removed

frontend/src/utils/index.js
grolu: removed utils function

frontend/src/composables/useShootSpec.js
grolu: changed to property shorthand

frontend/src/composables/useShootSpec.js
grolu: It should be always set. However, if the user manually edits the yaml spec and deletes it.... Same problem as we have in other places, too. We could make it more resilient and check if providerConfig is set, too. WDYT?

frontend/src/composables/useShootSpec.js
grolu: Yes, tis is kind of a leftover. I learned about this resources secretRef thingy only after fist implementation. Moved it to GShootInfrastructureCard

frontend/src/sass/main.scss
grolu: Is only used for dns-provider and hibernation schedules days selection. Reduced the min with to make more space in case input fields are empty, DNS Provider row would take up two lines just to show empty input fields for include / exclude input fields.
max-width is unchanged so nothing changes in case content is larger than 300px (which is the case for default hibernation schedules with weekdays in selection)

frontend/src/composables/useShootSpec.js
grolu: we no longer need `shootDnsProviders` as we read additional DNS providers from `spec.extensions`. We only care about the primary provider. For historical reasons this is a list but in the future only one entry (primary provider) will be allowed.

frontend/src/composables/useShootContext/index.js
grolu: changed. I prefer the object shorthand. But for consistency it makes sense to use the property shorthand.

frontend/src/composables/useShootContext/index.js
grolu: moved outside of composable scope

frontend/src/composables/useShootSpec.js
petersutter: +1 to make it more resilient

frontend/src/components/ShootDns/GManageDns.vue
```diff
@@ -110,53 +160,142 @@ import {
...
+    primaryDnsProviderSecret: {
+      get () {
+        const dnsSecrets = this.dnsSecretsByProviderKind(this.primaryDnsProviderType)
+        return find(dnsSecrets, ['metadata.secretRef.name', this.dnsPrimaryProvider.secretName])
+      },
+      set (value) {
+        this.dnsPrimaryProvider.secretName = value?.metadata.secretRef.name
+      },
+    },
+    showCustomDomainRecommendation () {
```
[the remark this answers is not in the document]
grolu: ok

frontend/src/composables/useShootSpec.js
```diff
@@ -119,11 +118,16 @@ export function useShootSpec (shootItem, options = {}) {
...
+    return find(shootSpec.value.dns?.providers, { primary: true })
```
holgerkoser, line 125:
```suggestion
    return find(shootSpec.value.dns?.providers, 'primary')
```

frontend/src/composables/useShootSpec.js
```diff
@@ -119,11 +118,16 @@ export function useShootSpec (shootItem, options = {}) {
...
+  })
+
+  const shootExtensionDnsProviders = computed(() => {
```
holgerkoser, line 128: rename to `shootDnsServiceExtensionProviders`

frontend/src/composables/useShootContext/index.js
```diff
@@ -843,219 +868,165 @@ export function createShootContextComposable (options = {}) {
...
   }
 
-  function findDnsProviderSecret (type, secretName) {
-    const secrets = secretStore.dnsSecretsByProviderKind(type)
-    return find(secrets, ['metadata.secretRef.name', secretName])
+  function deleteExtensionDnsProviderResourceRef (resourceName) {
+    manifest.value.spec.resources = filter(resources.value, ({ name }) => name !== resourceName)
+  }
+
+  function secretNameFromResources (resourceName) {
```
holgerkoser, line 1027: The name should be something like `getResourceRefName` or `findResourceRefName`

frontend/src/composables/useShootContext/index.js
```diff
@@ -843,219 +868,165 @@ export function createShootContextComposable (options = {}) {
...
+    if (!extensionDns.value.providerConfig.providers.length) {
+      setExtension(manifest.value, 'shoot-dns-service', undefined)
     }
   }
 
-  function findDnsProviderSecret (type, secretName) {
-    const secrets = secretStore.dnsSecretsByProviderKind(type)
-    return find(secrets, ['metadata.secretRef.name', secretName])
+  function deleteExtensionDnsProviderResourceRef (resourceName) {
+    manifest.value.spec.resources = filter(resources.value, ({ name }) => name !== resourceName)
```
holgerkoser, line 1024: Create a writable computed property for `resources`

frontend/src/composables/useShootContext/index.js
```diff
@@ -843,219 +868,165 @@ export function createShootContextComposable (options = {}) {
...
+      resourceRef: {
+        kind: 'Secret',
+        name: secretName,
+        apiVersion: 'v1',
+      },
+    })
+  }
+
+  function deleteExtensionDnsProvider (index) {
+    deleteExtensionDnsProviderResourceRef(extensionDnsProviders.value[index].secretName) // provider.secretName is the resourceName
```
holgerkoser, line 1015: It should not fail if the index does not exist

frontend/src/components/Secrets/GSelectSecret.vue
```diff
@@ -120,6 +120,14 @@ export default {
     registerVuelidateAs: {
       type: String,
     },
+    filterSecretNames: {
```
holgerkoser, line 123: The name filterSecretName suggests that it is a function that filters a list of secretNames. In reality, however, it is an allowlist of possible or permitted secret names. The name should reflect that. I would use something like `allowedSecretNames` oder `permittedSecretNames`.

frontend/src/components/ShootDns/GDnsProviderRow.vue
```diff
@@ -176,58 +164,85 @@ export default {
...
+    },
+    includeZones: {
+      get () {
+        return get(this.dnsProvider, 'zones.include')
+      },
+      set (value) {
+        set(this.dnsProvider, 'zones.include', value)
+      },
+    },
+    filterSecretNames () {
```
holgerkoser, line 238: Same as above. This is property not a function. Do not use verbs for properties!

frontend/src/components/ShootDns/GDnsProviderRow.vue
```diff
@@ -176,58 +164,85 @@ export default {
...
       set (value) {
         this.dnsProvider.type = value
         const dnsSecrets = this.dnsSecretsByProviderKind(value)
         const defaultDnsSecret = head(dnsSecrets)
-        this.dnsProvider.secretName = get(defaultDnsSecret, 'metadata.secretRef.name', null)
+        this.extensionDnsProviderSecret = defaultDnsSecret
       },
     },
-    dnsProviderSecret: {
+    extensionDnsProviderSecret: {
```
holgerkoser, line 185: The name we used the the dns service provider is `dnsServiceExtensionProvider`. This property should have the name `dnsServiceExtensionProviderSecret` or shorter `dnsProviderSecret`

frontend/src/components/ShootDns/GDnsProviderRow.vue
```diff
@@ -176,58 +164,85 @@ export default {
...
+      },
+    },
+    filterSecretNames () {
+      return this.dnsServiceExtensionProviders.map(provider => {
+        const secretName = this.getResourceRefName(provider.secretName) // provider.secretName is the resource name
+        if (secretName !== this.extensionDnsProviderSecret?.metadata.name) {
+          return secretName
+        }
+        return undefined
+      })
```
holgerkoser, line 245:
```suggestion
      const dnsProviderSecretName = get(this.dnsProviderSecret, 'metadata.name') 
      return this.dnsServiceExtensionProviders.filter(provider => {
        return dnsProviderSecretName !== this.getResourceRefName(provider.secretName)) {
      })
```

frontend/src/components/ShootDns/GManageDns.vue
```diff
@@ -5,158 +5,314 @@ SPDX-License-Identifier: Apache-2.0
...
+        this.dnsPrimaryProviderType = type
+        const dnsSecrets = this.dnsSecretsByProviderKind(type)
+        this.primaryDnsProviderSecret = head(dnsSecrets)
+        this.v$.dnsDomain.$reset()
+      }
     },
   },
   mounted () {
-    this.v$.$touch()
+    if (this.customDomainEnabled) {
```
holgerkoser, line 308: Initialize in `setup`

frontend/src/components/Secrets/GSecretDialogWrapper.vue
```diff
@@ -24,7 +24,7 @@ import VsphereDialog from '@/components/Secrets/GSecretDialogVSphere'
 import CloudflareDialog from '@/components/Secrets/GSecretDialogCloudflare'
 import InfobloxDialog from '@/components/Secrets/GSecretDialogInfoblox'
 import NetlifyDialog from '@/components/Secrets/GSecretDialogNetlify'
-import DdnsDialog from '@/components/Secrets/GSecretDialogDDNS'
+import DDnsDialog from '@/components/Secrets/GSecretDialogDDNS'
```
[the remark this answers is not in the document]
grolu: This is because of the componentName logic. See other imports. As the cloud provider kind is now rfc2136, we can import using actual name

frontend/src/components/ShootDns/GManageDns.vue
```diff
@@ -5,158 +5,314 @@ SPDX-License-Identifier: Apache-2.0
...
+        </v-alert>
       </v-col>
     </v-row>
-  </div>
+    <template v-if="hasDnsServiceExtension">
+      <div class="text-subtitle-1 my-3">
+        DNS Providers for the shoot-dns-service Extension
+      </div>
+      <div class="text-caption">
+        Configure DNS providers for the shoot-dns-service-extension to expose cluster workloads using domains managed by the configured providers
```
petersutter, line 107:
```suggestion
        Configure DNS providers for the shoot-dns-service extension to automatically manage and synchronize DNS entries for cluster resources like services and ingresses
```
also, in the original code it was `shoot-dns-service-extension` instead of `shoot-dns-service extension`

frontend/src/components/ShootDns/GManageDns.vue
```diff
@@ -5,158 +5,314 @@ SPDX-License-Identifier: Apache-2.0
...
+          max-width="50%"
         >
-          <v-icon class="text-primary">
-            mdi-plus
-          </v-icon>
-          Add DNS Provider
-        </v-btn>
+          <div>
+            The cluster domain is used for the API server of this cluster.
+            To expose workload services with your cluster domain, add a DNS provider for the <code>shoot-dns-service</code> extension.
```
petersutter, line 88:
```suggestion
            To manage DNS entries for your cluster domain, add a DNS provider for the <code>shoot-dns-service</code> extension.<br>
```

frontend/src/components/ShootDetails/GShootInfrastructureCard.vue
```diff
@@ -400,9 +414,10 @@ export default {
       'cloudProfileByName',
       'floatingPoolsByCloudProfileNameAndRegionAndDomain',
     ]),
-    ...mapActions(useSecretStore, [
-      'getCloudProviderSecretByName',
-    ]),
+    secretNameFromShootResources (resourceName) {
```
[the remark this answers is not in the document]
grolu: The method in the composable operates on the shootcontext manifest. If we want to re-use this method we could pass the shoot resource as parameter. Or we would need to initialize the composable with the current shoot item. However, in this component we only show data, not modify. Do we really want to use methods from shootContext composable here?

frontend/src/components/ShootDns/GDnsConfiguration.vue
```diff
@@ -74,10 +76,21 @@ export default {
     },
     async updateConfiguration () {
       try {
+        // Remove migration logic when all dns configurations have been migrated by Gardener
+        this.forceMigrateSyncDNSProvidersToFalse()
```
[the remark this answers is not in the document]
grolu: ok

frontend/src/components/ShootDns/GDnsProviderRow.vue
grolu: changed to allowedSecretNames

frontend/src/components/Secrets/GSelectSecret.vue
grolu: changed to allowedSecretNames

frontend/src/components/ShootDns/GDnsProviderRow.vue
grolu: this will not work... same problem as mentioned above

holgerkoser: Same problem as mentioned above? I don't get it. Why do we map something to undefined?

frontend/src/components/ShootDetails/GShootInfrastructureCard.vue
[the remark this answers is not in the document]
holgerkoser: No. We do not want to the shootContext composable here but we want to use a composable the deals with shoot spec resources and may be extensions.

frontend/src/components/ShootDns/GDnsProviderRow.vue
holgerkoser: I sill think this name is not in line with the convention we use for other names for `shoot-dns-service` provider properties?

frontend/src/components/ShootDns/GDnsProviderRow.vue
grolu: Sorry, you are right. Not the same problem. The optimized code should look like this:
```
const dnsProviderSecretName = get(this.dnsServiceExtensionProviderSecret, 'metadata.name')
const secretNames = this.dnsServiceExtensionProviders.map(({ secretName }) => this.getResourceRefName(secretName))
const filtered = secretNames.filter(secretName => {
  return dnsProviderSecretName !== secretName
})
return filtered
```

petersutter (approved):
/lgtm

holgerkoser (approved):
/lgtm

PR #1905, opened 2024-06-07
[hotfix-1.74] Do not fail if tickets could not be fetched

petersutter:
- Do not fail if tickets could not be fetched
- fix margins when displaying multiple tickets

Fixes #

**Release note**:

```bugfix user
Fixed an issue where the shoot details page could not be loaded when GitHub comments could not be fetched
```


holgerkoser (approved):
/lgtm

PR #1926, opened 2024-06-19
Add Custom Field Configuration Editor for Shoot Clusters

petersutter:
**What this PR does / why we need it**:
This PR introduces an editor for custom shoot fields. The new feature allows users to configure custom fields directly from the dashboard without needing to use `kubectl`. 
custom-fields-5
custom-fields-6
custom-fields-7

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Release note**:

```feature user
You can now configure custom shoot fields directly from the dashboard using the new editor, eliminating the need to use `kubectl`.
```

PR #1943, opened 2024-06-27
Fix issue when no costObject is defined

petersutter:
Fixed issue introduced with #1926 and #1934 where the following error is displayed in the console logs when no cost object is defined
```
Error: [@vuelidate/validators]: First parameter to "withMessage" should be string or a function returning a string, provided undefined
```

Fixes #

**Release note**:

```bugfix user

```


holgerkoser (approved):
/lgtm

PR #1950, opened 2024-07-02
Avoid side effects using `trimObjectMetadata`

petersutter:
This PR ensures that the `list` function does not modify the original project objects to prevent any unintended side effects on the original project objects stored in the cache.

Fixes #

**Release note**:

```bugfix user

```


grolu (approved):
/lgtm

PR #1952, opened 2024-07-02
GShootList: Reflect search query in URL

petersutter:
This PR updates the `GShootList` component to reflect the search query in the URL. When the `shootSearch` changes, the URL is updated so users can bookmark the URL with the search query. 

Fixes #

**Release note**:

```feature user
Reflect search query on project clusters page in URL
```

### DIFF
--- a/frontend/src/views/GShootList.vue
+++ b/frontend/src/views/GShootList.vue
@@ -232,14 +232,13 @@ import {
   provide,
   toRef,
   watch,
-  onMounted,
 } from 'vue'
 import {
   mapState,
   mapWritableState,
   mapActions,
 } from 'pinia'
-import { useRoute } from 'vue-router'
+import { useUrlSearchParams } from '@vueuse/core'
 
 import { useAuthnStore } from '@/store/authn'
 import { useAuthzStore } from '@/store/authz'
@@ -305,7 +304,6 @@ export default {
   },
   setup () {
     const projectStore = useProjectStore()
-    const route = useRoute()
 
     const activePopoverKey = ref('')
     provide('activePopoverKey', activePopoverKey)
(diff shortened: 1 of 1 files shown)


grolu:
I see this warning in the logs:

Screenshot 2024-07-05 at 11 43 34

frontend/src/views/GShootList.vue
```diff
@@ -311,15 +315,39 @@ export default {
...
+
+    function setSearchFromUrl () {
+      const searchQuery = route.query.search || ''
+      shootSearch.value = searchQuery
+      debouncedShootSearch.value = searchQuery
+    }
+
+    onMounted(() => {
+      setSearchFromUrl()
+    })
```
holgerkoser, line 340: I would use `vueuse` composable in this case
```js
    const params = useUrlSearchParams('hash-params')
    const shootSearch = ref(params.q)
    const debouncedShootSearch = ref(shootSearch.value)

    watch(params, value => {
      if (value.q !== shootSearch.value) {
        debouncedShootSearch.value = shootSearch.value = value.q
      }
    })

    watch(debouncedShootSearch, value => {
      params.q = value
    })
```

petersutter: ✅

petersutter:
The warning is now resolved, thanks to @holgerkoser's suggestion

frontend/src/views/GShootList.vue
```diff
@@ -311,15 +313,29 @@ export default {
...
+    watch(params, value => {
+      if (value.q !== shootSearch.value) {
+        debouncedShootSearch.value = shootSearch.value = value.q
+      }
+    })
+
+    watch(shootSearch, value => {
+      params.q = value
+    })
+
```
holgerkoser, line 329: * If I change the hash value in the URL the `searchQuery` is not updated correctly due to the `resetSearchQuery` call in the navigation guards.

* To delete the parameter q in the URL we should set the value of `params.q` to `null`. I would not delete the property of the reactive object.
  
I would propose to change the code like this:
```js
    const params = useUrlSearchParams('hash-params')
    const shootSearch = ref(params.q)
    const debouncedShootSearch = ref(shootSearch.value)

    function setShootSearch (value) {
      shootSearch.value = value
      setDebouncedShootSearch()
    }

    function resetShootSearch () {
      debouncedShootSearch.value = shootSearch.value = params.q
    }

    const setDebouncedShootSearch = debounce(() => {
      debouncedShootSearch.value = shootSearch.value
    }, 500)

    watch(() => params.q, value => {
      if (shootSearch.value !== value) {
        setShootSearch(value)
      }
    })

    watch(shootSearch, value => {
      if (!value) {
        params.q = null
      } else if (params.q !== value) {
        params.q = value
      }
    })
```

petersutter: ✅

holgerkoser (approved):
/lgtm

grolu (approved):
/lgtm

PR #1959, opened 2024-07-04
[DRAFT] Activate Popover using activator prop

grolu:
This introduces an activator prop als an alternative way to activate popovers.

Fixes #

Currently the popover position is not correct. The arrow is not positioned in the middle of the activator element. It is placed at the left side.

**Release note**:

```bugfix user

```


grolu:
/close
As discussed this is currently out of scope.

PR #1969, opened 2024-07-10
Fixed remove last dns provider

grolu:

Fixes #

**Release note**:

```bugfix user

```

### DIFF
--- a/frontend/src/composables/useShootDns.js
+++ b/frontend/src/composables/useShootDns.js
@@ -45,6 +45,21 @@ export const useShootDns = (manifest, options) => {
     deleteExtension,
   } = useShootExtensions(manifest)
 
+  const setDnsServiceExtension = providers => {
+    const providerConfig = {
+      apiVersion: 'service.dns.extensions.gardener.cloud/v1alpha1',
+      kind: 'DNSConfig',
+      syncProvidersFromShootSpecDNS: false,
+    }
+    if (providers) {
+      providerConfig.providers = providers
+    }
+    setExtension({
+      type: 'shoot-dns-service',
+      providerConfig,
+    })
+  }
+
   const dnsServiceExtension = computed(() => {
     return find(extensions.value, ['type', 'shoot-dns-service'])
   })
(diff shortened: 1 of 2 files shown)


frontend/src/composables/useShootDns.js
```diff
@@ -45,6 +45,21 @@ export const useShootDns = (manifest, options) => {
...
+      syncProvidersFromShootSpecDNS: false,
+    }
+    if (providers) {
+      providerConfig.providers = providers
+    }
+    setExtension({
+      type: 'shoot-dns-service',
+      providerConfig,
+    })
+  }
```
holgerkoser, line 61:
```suggestion
  function setDnsServiceExtensionProviders (providers) {
    setExtension({
      type: 'shoot-dns-service',
      providerConfig: {
        apiVersion: 'service.dns.extensions.gardener.cloud/v1alpha1',
        kind: 'DNSConfig',
        syncProvidersFromShootSpecDNS: false,
        providers,
      },
    })
  }
```
It make no difference if the property does not exist or the value is `undefined`

frontend/src/composables/useShootDns.js
```diff
@@ -177,15 +192,7 @@ export const useShootDns = (manifest, options) => {
...
-      setExtension({
-        type: 'shoot-dns-service',
-        providerConfig: {
-          apiVersion: 'service.dns.extensions.gardener.cloud/v1alpha1',
-          kind: 'DNSConfig',
-          syncProvidersFromShootSpecDNS: false,
-          providers: [provider],
-        },
-      })
+      setDnsServiceExtension([provider])
```
holgerkoser, line 195:
```suggestion
      setDnsServiceExtensionProviders([provider])
```

frontend/src/composables/useShootDns.js
```diff
@@ -222,6 +229,9 @@ export const useShootDns = (manifest, options) => {
       // Migrate from old DNS configuration to new DNS configuration
       set(dnsServiceExtension.value, 'providerConfig.syncProvidersFromShootSpecDNS', false)
       dnsProviders.value = filter(dnsProviders.value, ['primary', true])
+    } else if (!dnsServiceExtension.value) {
+      // set default empty dns service extension where sync is set to off
+      setDnsServiceExtension()
```
holgerkoser, line 234:
```suggestion
      setDnsServiceExtensionProviders()
```

frontend/src/composables/useShootDns.js
grolu: why would you rename the function. I agree that it makes it clearer to include `providers` in the name as they are passed. But this function not only sets the providers but also the extension itself. I think that this information gets lost by removing it?

holgerkoser: The only information you pass to this function is the list of providers. The rest is hardcoded and can not be changed. So it does not update the extension but only the provider. As as side effect it populates the hardcoded rest of the extension properties. Both solutions are not optimal. But I wanted to keep the propsal simple.

holgerkoser, line 195:
```suggestion
      setDnsServiceExtension({ providers: [provider] })
```

holgerkoser (approved):
/lgtm

petersutter (approved):
/lgtm

PR #1972, opened 2024-07-11
[master] Fix issue with Ticket link

petersutter:
- Fixed a bug where navigating between cluster details pages caused new ticket titles to inherit the title from the previously viewed cluster. This issue occurred because the frontend configuration was mutated instead of creating a copy when constructing the ticket URL
- pinned braintree/sanitize-url to 7.0.1 as with ^7.0.2 urls params that where previously encoded would be decoded https://github.com/braintree/sanitize-url/issues/75

Fixes #

**Release note**:

```bugfix user
Fixed a bug where navigating between cluster details pages caused new ticket titles to inherit the title from the previously viewed cluster
```


holgerkoser (approved):
/lgtm

PR #2033, opened 2024-08-14
Show warnings returned in admission reviews

### LINKED ISSUE #1474: Show warnings returned in admission reviews when updating shoots
The dashboard provides the means to edit a shoot spec directly from the browser. If an admission webhook denies the request the returned error is displayed to the user, however if the request is allowed and returned with a warning that warning is not displayed to the user. It will be nice if such warnings are shown in the dashboard as well. Similarly if I edit a resource via `kubectl` and a warning is returned that is displayed in the terminal.

```bash
k -n garden-local edit shoot local
Warning: my cool warning message
shoot.core.gardener.cloud/local edited
```

**Why is this needed**:
This will improve user experience in a sense that it might bring attention to potential problems and misconfigurations.

grolu:
- Show kubernetes warnings as alert
- Make update errors persistent alerts
- Refactored app store to use notify composable

Screenshot 2024-10-08 at 17 38 36

Fixes #1474

- @holgerkoser Check changes to kubernetes client, Also currently only create / update methods support to return response
- Could also add logic for edit workes as there also the yaml can be edited WDYT?

**Release note**:

```feature user
Warnings returned in admission reviews are now displayed when creating or updating clusters using the cluster editor
```


packages/request/lib/Client.js
```diff
@@ -320,6 +320,11 @@ class Client {
         body
       })
     }
+
+    if (returnResponse) {
```
petersutter, line 324: as you only need the headers for this feature, how about making use of the afterResponse hook (line 314)?

grolu: I think the headers are not available in the hook but I'm not 100% sure. I remember that I tried to use the hook but it was not possible. @holgerkoser may has a better idea regarding the implementation

holgerkoser: I have added an option `onWarning` to the kube-client that give access to the warning header value without changing the method signatures.

frontend/src/store/app.js
```diff
@@ -51,9 +51,16 @@ export const useAppStore = defineStore('app', () => {
     } else {
       alert.message = ''
     }
+    if (persistent) {
+      alert.duration = -1
+    }
     setAlert(alert)
   }
 
+  function setPersistentError (value) {
```
holgerkoser, line 60: Do not add new methods. The existing api can handle this case.
```js
  function setAlertWithType (type, value) {
    const alert = {
      type,
    }
    if (typeof value === 'string') {
      alert.message = value
    } else if (value) {
      const { message = '', title, duration } = value
      alert.message = value.response
        ? errorDetailsFromError(value).detailedMessage
        : message
      if (title) {
        alert.title = title
      }
      if (duration) {
        alert.duration= duration
      }
    } else {
      alert.message = ''
    }
    setAlert(alert)
  }
```
Usage:
```js
  appStore.setError({
    message: `Failed to reset Service Account: ${err.message}`,
    duration: -1
  })
```

frontend/src/views/GNewShoot.vue
```diff
@@ -243,11 +244,15 @@ export default {
...
 
       try {
-        await this.api.createShoot({
+        const { headers } = await this.api.createShoot({
           namespace: this.shootNamespace,
           data: this.shootManifest,
         })
-        this.setSuccess('Cluster created')
+        if (headers.warning) {
+          this.setPersistentWarning({ message: headers.warning })
```
holgerkoser, line 252: The text is not renderd correctly. Long strings are cut off. We could use https://github.com/marcbachmann/warning-header-parser or create our own implementation.

grolu: Added parser

holgerkoser (approved):
/lgtm

PR #2058, opened 2024-08-30
Upgrade to CodeMirror 6

### LINKED ISSUE #1892: Upgrade to Codemirror v6
Sicne we still use the old Codemirror v5 library there are always three warnings int the console. 
image

Upgarde to the new version 6 by following the [migration guide](https://codemirror.net/docs/migration/)
**Why is this needed**:

grolu:
Upgrade CodeMirror 5 to CodeMirror 6

Fixes #1892

**Release note**:

```feature user
Upgraded the code editor from CodeMirror 5 to CodeMirror 6 to enhance performance, modernize the interface, and improve extensibility
```


frontend/src/composables/useShootEditor/index.js
```diff
@@ -68,36 +102,31 @@ export function useShootEditor (initialValue, options = {}) {
...
-      ...options.extraKeys,
-    }
-    const extraKeys = {}
-    for (const [key, value] of Object.entries(originalExtraKeys)) {
-      extraKeys[localStorageStore.editorShortcuts[key] ?? key] = value
+      ...(options.extraKeys ?? []),
+    ]
+    for (const extraKey of extraKeys) {
+      if (localStorageStore.editorShortcuts[extraKey.key]) {
+        extraKey.key = localStorageStore.editorShortcuts[extraKey.key]
```
petersutter, line 128: use lodash get instead of using bracket notation

```js
if (get(localStorageStore.editorShortcuts, [extraKey.key])) {
    extraKey.key = get(localStorageStore.editorShortcuts, [extraKey.key])
}
```

frontend/src/composables/useShootEditor/index.js
```diff
@@ -257,33 +312,24 @@ export function useShootEditor (initialValue, options = {}) {
...
-    if (cm.value) {
-      cm.value.doc.clearHistory()
-      generation.value = cm.value.doc.changeGeneration()
+    if (cmView.value) {
+      cmView.value.dispatch({
+        effects: historyCompartment.reconfigure([]),
+      })
+      cmView.value.dispatch({
+        effects: historyCompartment.reconfigure(history()),
+      })
```
petersutter, line 332: could you kindly explain this code?

frontend/src/composables/useShootEditor/index.js
```diff
@@ -295,40 +341,38 @@ export function useShootEditor (initialValue, options = {}) {
...
 
-  watch(cmTheme, value => {
-    if (cm.value) {
-      cm.value.setOption('theme', value)
-    }
+  watch(isDarkMode, isDarkMode => {
+    cmView.value.dispatch({
+      effects: themeCompartment.reconfigure([isDarkMode ? oneDark : [], syntaxHighlighting(isDarkMode ? oneDarkHighlightStyle : defaultHighlightStyle)]),
+    })
   })
```
petersutter, line 376: added additional line breaks for better readability
```suggestion
  watch(isDarkMode, isDarkMode => {
    cmView.value.dispatch({
      effects: themeCompartment.reconfigure([
        isDarkMode ? oneDark : [],
        syntaxHighlighting(isDarkMode ? oneDarkHighlightStyle : defaultHighlightStyle)
      ])
    })
  })
```

frontend/src/composables/useEditorLineHighlighter.js
```diff
@@ -9,94 +9,154 @@ import {
...
+      return valuesFromParams()
     },
     set (values = []) {
-      const [startLine, endLine = startLine] = values
+      const valuesCurrent = valuesFromParams()
+
+      let [startLine, endLine = startLine] = values
+      const [startLineCurrent, endLineCurrent] = valuesCurrent
+
+      if (endLine === startLine && startLine === startLineCurrent && !endLineCurrent) {
```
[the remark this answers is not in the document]
grolu: I know. The problem here is that I did not find a way to add a key listener to the gutter. It is possible to add a key listener to the editor itself and clear the highlight on escape. However, the editor needs to be focused, so you would need to click inside the editor to make this work.
So I checked how Github does the highlighting and it does not clear the selection on escape, it works exactly like I implemented it now.

frontend/src/composables/useEditorLineHighlighter.js
```diff
@@ -9,94 +9,154 @@ import {
...
+  }
+
+  const setHighlightEffect = StateEffect.define()
+  const highlightField = StateField.define({
+    create () {
+      return Decoration.none
+    },
+    update (highlights, tr) {
+      // need to re-add when editor updates
+      cmView.dom.classList.add('g-cm-line-highlighter')
```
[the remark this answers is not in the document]
grolu: This is because of the border in the g-highlighted css classes. Not sure why this did not affect cm5 lines... However, I changed the css to use box-shadow for the left border and pseudo-elements for top and bottom lines (box-shadow did not work here as the classes would overwrite each-other)

frontend/src/composables/useShootEditor/helper.js
```diff
@@ -150,106 +158,83 @@ export class EditorCompletions {
     }
 
     forIn(completions, (completion, propertyName) => {
-      let text = generateCompletionText(propertyName, completion.type, token.type)
+      const text = generateCompletionText(propertyName, completion.type, token.type)
       if (token.type === 'firstArrayItem') {
-        text = '- ' + text
+        // text = '- ' + text // TODO: Selection ccurently does not replace the bullet
```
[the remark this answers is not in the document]
grolu: can be removed

frontend/src/composables/useShootEditor/helper.js
```diff
@@ -150,106 +158,83 @@ export class EditorCompletions {
...
-      const type = generateTypeText(completion.type, completion.format)
+      const typeText = generateTypeText(completion.type, completion.format)
 
       completionArray.push({
-        text,
-        string,
+        label: string,
+        displayLabel: text,
+        detail: typeText,
+        type: 'keyword', // TODO check weather we want to show diffrent icons
```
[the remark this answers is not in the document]
grolu: https://codemirror.net/docs/ref/#autocomplete.Completion.type
It is also possible to omit the type property so that no icon is rendered. AFAIK it is not possible to provide a **custom** render function for the completion box (like we did for CM5), only for the content that is shown next to it, the _info_ (details) it is possible to provide a render function

frontend/src/composables/useShootEditor/index.js
grolu: This code resets the internal history plugin. AFAIK there is no better way to do this. ChatGPT tells you this:
>In CodeMirror 6, the history is managed internally by the history extension, and there isn't a public API to directly reset or clear the history state. Directly manipulating internal state fields or extensions is generally not recommended, as it can lead to unexpected behavior and may break with future updates.

However, simply reconfiguring with
```
cmView.value.dispatch({
  effects: historyCompartment.reconfigure(history()),
})
```

does not reset the history. Explicitly removing and re-adding the history plugin is the only way I got it working.

frontend/src/composables/useShootEditor/helper.js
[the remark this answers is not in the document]
petersutter: let's leave it like it is for now, I do not have a better suggestion

frontend/src/composables/useShootEditor/helper.js
```diff
@@ -83,9 +90,9 @@ export class EditorCompletions {
         end,
         string,
       }
-      if (token.start <= pos.ch && pos.ch <= token.end) {
+      if (token.start <= lineChar && lineChar <= token.end) {
         // Ensure that mouse pointer is on propety and not somewhere else on this line
```
petersutter, line 94:
```suggestion
        // Ensure that mouse pointer is on property and not somewhere else on this line
```

frontend/src/composables/useShootEditor/helper.js
```diff
@@ -83,9 +90,9 @@ export class EditorCompletions {
...
-        const completions = this._getYamlCompletions(token, pos, cm, true)
+        const completions = this._getYamlCompletions(token, { ch: lineChar, line }, cmView, true)
```
petersutter, line 95: extract to variable for better readability
```suggestion
        const cur = { ch: lineChar, line }
        const completions = this._getYamlCompletions(token, cur, cmView, true)
```

frontend/src/composables/useShootEditor/helper.js
```diff
@@ -315,35 +300,22 @@ export class EditorCompletions {
...
     token.type = 'property'
-    return token
-  }
-
-  _getTokenLine (cm, cur) {
-    const lineTokens = cm.getLineTokens(cur.line)
-    const lineString = join(map(lineTokens, 'string'), '')
 
-    return { lineTokens, lineString }
+    return token
```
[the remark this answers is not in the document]
petersutter: do you want to change it in this PR anyway?

petersutter (approved):
minor things, otherwise /lgtm

frontend/src/composables/useEditorLineHighlighter.js
```diff
@@ -9,94 +9,154 @@ import {
...
+
+      for (const effect of tr.effects) {
+        if (effect.is(setHighlightEffect)) {
+          return effect.value
+        }
+      }
+      return highlights
+    },
+    provide: f => EditorView.decorations.from(f),
```
holgerkoser, line 106:
```suggestion
    provide (f) {
      return EditorView.decorations.from(f)
    },
```

frontend/src/composables/useEditorWhitespace.js
```diff
@@ -0,0 +1,85 @@
...
+  MatchDecorator,
+  Decoration,
+  ViewPlugin,
+  WidgetType,
+} from '@codemirror/view'
+
+export function useEditorWhitespace () {
+  const matchDecorator = new MatchDecorator({
+    regexp: /[ \t]/g,
+    decoration: match => {
```
holgerkoser, line 17:
```suggestion
    decoration (match) {
```

frontend/src/composables/useEditorWhitespace.js
```diff
@@ -0,0 +1,85 @@
...
+      if (match[0] === '\t') {
+        return Decoration.mark({
+          attributes: { class: 'g-cm-highlightedTab' },
+        })
+      } else if (match[0] === ' ') {
```
holgerkoser, line 22:
```suggestion
      }
      if (match[0] === ' ') {
```

frontend/src/composables/useShootEditor/helper.js
holgerkoser: What is cur? Is it current? Or is it cursor? Use easy understable names.

frontend/src/composables/useShootEditor/helper.js
```diff
@@ -151,106 +159,80 @@ export class EditorCompletions {
...
+        return includes(completion.label, tokenString)
       })
     }
+
     return completionArray
   }
 
   // get token at cursor position in editor with yaml content
-  _getYamlToken (cm, cur) {
+  _getYamlToken (cmView, cur) {
```
holgerkoser, line 195: These methods starting with underscore should be private and the underscore removed.

frontend/src/sass/main.scss
```diff
@@ -154,3 +120,84 @@
...
+      background-image: url('data:image/svg+xml,<svg xmlns="http://www.w3.org/2000/svg" width="10" height="6"><path d="M0 3 L10 3" stroke="#{$whitespace-color-encoded}" stroke-width="1" /></svg>');
+      background-repeat: no-repeat;
+      background-position: center;
+      background-size: contain;
+    }
+
+    .g-cm-newline-marker {
+      color: $whitespace-color;
+    }
+  }
```
holgerkoser, line 190:
```suggestion
@mixin alternate-row-background($background-color) {
  .alternate-row-background>div:nth-of-type(even) {
    background-color: $background-color;
  }
}

@mixin g-highlighted($background-color, $border-color) {
  .g-highlighted {
    position: relative;
    background-color: $background-color;
    /* Left border */
    box-shadow: inset 1px 0 0 0 $border-color;

    /* Top border */
    &.g-highlighted--top::before {
      content: '';
      position: absolute;
      top: 0;
      left: 0;
      right: 0;
      height: 1px;
      /* Adjust border thickness as needed */
      background-color: $border-color;
    }

    /* Bottom border */
    &.g-highlighted--bottom::after {
      content: '';
      position: absolute;
      bottom: 0;
      left: 0;
      right: 0;
      height: 1px;
      /* Adjust border thickness as needed */
      background-color: $border-color;
    }
  }
}

@mixin g-cm-highlighted-whitespace($whitespace-color) {
  $whitespace-color-encoded: #{'%23' + str-slice(inspect($whitespace-color), 2)};

  .g-cm-highlightedSpace {
    background-image: url('data:image/svg+xml,<svg xmlns="http://www.w3.org/2000/svg" width="6" height="6"><circle cx="3" cy="3" r="1" fill="#{$whitespace-color-encoded}" /></svg>');
    background-repeat: no-repeat;
    background-position: center;
    background-size: contain;
  }

  .g-cm-highlightedTab {
    background-image: url('data:image/svg+xml,<svg xmlns="http://www.w3.org/2000/svg" width="10" height="6"><path d="M0 3 L10 3" stroke="#{$whitespace-color-encoded}" stroke-width="1" /></svg>');
    background-repeat: no-repeat;
    background-position: center;
    background-size: contain;
  }

  .g-cm-newline-marker {
    color: $whitespace-color;
  }
}

.v-theme-- {
```

frontend/src/views/GNewShootEditor.vue
```diff
@@ -134,7 +135,16 @@ onBeforeRouteLeave(async (to, from, next) => {
...
-  if (isShootDirty.value) {
+  if (!clean.value) {
+    if (!await confirmEditorNavigation()) {
+      focusEditor()
+      return next(false)
+    }
+  }
+  return next()
+})
+onBeforeRouteUpdate(async (to, from, next) => {
```
holgerkoser, line 146: What is the reason for this change. I think this has nothing to do with cm6. Isnt't this part of an ohter PR?

frontend/src/composables/useShootEditor/helper.js
grolu: I think this is a leftover convention from cm5 where they used `cur` for cursor in some places. Anyway, I changed all occurrences of `cur` to `cursor` in our code.

frontend/src/components/GShootEditor.vue
```diff
@@ -126,6 +125,20 @@ SPDX-License-Identifier: Apache-2.0
             <span>{{ showManagedFields ? 'Hide' : 'Show' }} managed fields</span>
           </v-tooltip>
         </div>
+        <div class="px-2">
+          <v-tooltip location="top">
+            <template #activator="slotProps">
+              <div v-bind="slotProps.props">
+                <g-action-button
+                  size="x-small"
+                  :icon="renderWhitespaces ? 'mdi-grid' : 'mdi-grid-off'"
```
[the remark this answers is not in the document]
grolu: @petersutter WDYT?
Also, do you think we should render line breaks at all?
Moreover, the setting is not stored anywhere currently. Should we store the state in local storage?

frontend/src/views/GNewShootEditor.vue
grolu: reverted

frontend/src/components/GShootEditor.vue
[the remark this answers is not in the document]
petersutter: the suggestion sounds good to me. 

> Also, do you think we should render line breaks at all?

I think we should not render the line breaks, it does not give much value

> Moreover, the setting is not stored anywhere currently. Should we store the state in local storage?

hm, yes probably this should be stored in local storage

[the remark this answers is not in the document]
grolu: Okay
- removed the new line symbols.
- included the proposed icons
- store whitespace setting in local storage

frontend/src/components/GShootEditor.vue
```diff
@@ -246,6 +255,27 @@ const showToolbar = computed(() => {
   return !isReadOnly.value && !props.hideToolbar
 })
 
+const whitespaceEye = computed(() => {
+  return `
+    <svg viewBox="0 0 24 24" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
```
holgerkoser, line 260: Import the svg images from the assets folder.

holgerkoser (approved):
/lgtm

frontend/src/components/GShootEditor.vue
```diff
@@ -126,6 +125,29 @@ SPDX-License-Identifier: Apache-2.0
...
+                  variant="text"
+                  size="x-small"
+                  flat
+                  icon
+                  @click="renderWhitespaces = !renderWhitespaces"
+                >
+                  <!-- eslint-disable vue/no-v-html -->
+                  <span
+                    style="width: 18px; height: 18px"
+                    v-html="renderWhitespaces ? whitespaceEye : whitespaceEyeOff"
```
petersutter, line 142:
```suggestion
                    v-html="renderWhitespaces ? whitespaceEyeOff : whitespaceEye"
```

grolu: Can we discuss this? I agree as of now it is different than we handle the icon for the managed fields. But my gut feeling is that it would be wrong to show the off icon when we render the whitespaces. How do other tools handle this kind of flag buttons? I need to check this

petersutter (approved):
/lgtm

PR #2076, opened 2024-09-12
Fix issues with hibernation schedule dialog

### LINKED ISSUE #2061: Hibernation Schedule "Wake up at" Time Not Saved in the Dialog
**What happened**:
When adding a "Wake up at" time to an existing Shoot's Hibernation Schedule (in the same row as a pre-existing "Hibernate at" time) and pressing the Save button, the dialog closes without any error message, but the "Wake up at" time is not saved.

However, when the same action is performed using the YAML editor, the "Wake up at" time is saved correctly and can be subsequently edited in the dialog. Unfortunately, once added, the "Wake up at" time cannot be removed through the dialog, which is another issue.

image

The "Wake up at" time should be saved when added through the dialog.
The "Wake up at" time should also be removable via the dialog.

1. Open an existing Shoot and go to the Hibernation Schedule.
2. Add a "Wake up at" time to a row that already has a "Hibernate at" time.
3. Press the Save button.
4. Observe that the dialog closes without errors, but the "Wake up at" time is not saved.
5. Perform the same action using the YAML editor and observe that the "Wake up at" time is saved.
6. Try to remove the "Wake up at" time using the dialog.

**Environment**:

petersutter:
This PR fixes two issues with the hibernation schedule dialog, where 
- start/end time could not be reset by clicking on the reset button
- start/end time not saved (#2061)

Fixes #2061

**Release note**:

```bugfix user
Fixed issues with hibernation schedule dialog: reset button and time saving
```

### DIFF
--- a/frontend/src/components/ShootHibernation/GHibernationScheduleEvent.vue
+++ b/frontend/src/components/ShootHibernation/GHibernationScheduleEvent.vue
@@ -215,10 +215,8 @@ export default {
       },
       set (value) {
         const time = parseTimeString(value)
-        if (time) {
-          set(this.scheduleEvent, 'end.minute', time.minute)
-          set(this.scheduleEvent, 'end.hour', time.hour)
-        }
+        set(this.scheduleEvent, 'end.minute', time?.minute)
+        set(this.scheduleEvent, 'end.hour', time?.hour)
       },
     },
     hibernateTime: {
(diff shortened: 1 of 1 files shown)


grolu (approved):
/lgtm

holgerkoser (approved):
/lgtm

PR #2141, opened 2024-10-17
Removed dashboard specific secretbinding labels

### LINKED ISSUE #781: `cloudprofile.garden.sapcloud.io/name` label should not be mandatory
Currently the dashboard does not show Secrets under `<host>/namespace/<namespace>/secrets/` when the SecretBinding does not have the `cloudprofile.garden.sapcloud.io/name` label.

From end user perspective it is not okay that you need to label the SecretBinding in order to be able to see it in the dashboard (and then edit it).
I would expect to create a SecretBinding with any means and to be able to see it in the dashboard without applying additional labels.
P.S. I also realise that the SecretBinding or Secret don't have a field that indicates the provider type which is required for the sorting by provider in the secrets page.

**Steps to reproduce**:

1. Create Secret and SecretBinding that does not have this label

```
cat <<EOF | kubectl apply -f -
apiVersion: v1
kind: Secret
metadata:
  name: crazy-botany
  namespace: garden-foo
type: Opaque
data:
  # omitted
---
kind: SecretBinding
metadata:
  name:  crazy-botany
  namespace: garden-foo
secretRef:
  name: crazy-botany
  namespace: garden-foo
EOF
```

2. Go to `<host>/namespace/<namespace>/secrets/` and ensure that only SecretBindings that have the label are shown

grolu:
cloudprofile.garden.sapcloud.io/name and gardener.cloud/dnsProviderName

Fixes #781

**Release note**:

```breaking user
The dashboard no longer relies on the specific labels `cloudprofile.garden.sapcloud.io/name` and `gardener.cloud/dnsProviderName` to filter when selecting secrets. Instead, it now solely relies on the newly introduced provider.type field. As a result, cloud profiles are no longer selectable during secret creation, and all cloud profiles for the current infrastructure are now available for selection
```


frontend/src/views/GSecrets.vue
```diff
@@ -407,11 +406,10 @@ export default {
         isOwnSecret: isOwnSecret(secret),
         secretNamespace: secret.metadata.secretRef.namespace,
         secretName: secret.metadata.secretRef.name,
-        infrastructureName: secret.metadata.cloudProviderKind,
+        cloudProviderKind: secret.metadata.provider.type,
```
holgerkoser, line 409: The name should be `cloudProviderType` or may be even short `providerType`

frontend/src/views/GSecrets.vue
```diff
@@ -525,8 +522,10 @@ export default {
     relatedShootCountDns (secret) {
       const shootsByDnsSecret = filter(this.shootList, shoot => {
         return some(shoot.spec.dns?.providers, {
-          type: secret.metadata.dnsProviderName,
           secretName: secret.metadata.name,
+        }) ||
+        some(shoot.spec.resources, ({ resourceRef }) => {
+          return resourceRef?.kind === 'Secret' && resourceRef?.name === secret.metadata.name
```
holgerkoser, line 528:
```suggestion
        some(shoot.spec.resources, {
          resourceRef: {
            kind: 'Secret', 
            name: secret.metadata.name,
          },
```

frontend/src/store/secret.js
```diff
@@ -145,7 +149,7 @@ export const useSecretStore = defineStore('secret', () => {
     dnsSecretList,
     dnsSecretsByProviderKind,
     getCloudProviderSecretByName,
-    infrastructureSecretsByCloudProfileName,
+    infrastructureSecretsByCloudProviderKind,
```
holgerkoser, line 152: same here. rename to ...ProviderType

backend/__fixtures__/secretbindings.js
```diff
@@ -37,7 +37,7 @@ const secretBindingList = [
   getSecretBinding({
     namespace: 'garden-foo',
     name: 'trial-infra1',
-    cloudProfileName: 'infra1-profileName',
+    type: 'infra1',
```
petersutter, line 40: rename to providerType

holgerkoser: yes

grolu: done

frontend/src/store/secret.js
grolu: I replaced all occurrences of providerKind with providerType. I wanted to doi this in a separate PR at first. But as I started renaming the things I touched in this PR it felt like it makes no sense stopping there.

frontend/src/components/GVendor.vue
```diff
@@ -71,10 +67,7 @@ export default {
       type: Array,
       default: () => [],
     },
-    cloudProviderKind: {
-      type: String,
-    },
-    cloudProfileName: {
```
petersutter, line 77: I would not drop cloudProfileName. We can take the cloud profile name info from `Shoot.spec.cloudProfileName`

frontend/src/components/Secrets/GSecretDialog.vue
```diff
@@ -358,17 +326,10 @@ export default {
             name: this.name,
             namespace: this.namespace,
           },
+          provider: {
+            type: this.vendor,
```
petersutter, line 330: isn't vendor not a another alias (like infrastructureKind etc) for providerType? If yes, we should also use providerType accordingly

frontend/src/store/secret.js
```diff
@@ -90,27 +94,25 @@ export const useSecretStore = defineStore('secret', () => {
...
   function getCloudProviderSecretByName ({ namespace, name }) {
     return find(list.value, eqlNameAndNamespace({ name, namespace }))
   }
 
-  function infrastructureSecretsByCloudProfileName (cloudProfileName) {
-    return filter(list.value, ['metadata.cloudProfileName', cloudProfileName])
-  }
-
-  function dnsSecretsByProviderKind (dnsProviderName) {
+  function secretsByProviderType (providerType) {
```
petersutter, line 111: Use a factory function to create computed properties based on `providerType`

```js
function useSecretsByProviderType(providerType) {
  return computed(() => {
    const isDnsSecret = ..... // rest of the code
  })
}
```

frontend/src/store/secret.js
```diff
@@ -90,27 +94,25 @@ export const useSecretStore = defineStore('secret', () => {
...
+    const isDnsSecret = gardenerExtensionStore.dnsProviderTypes.includes(providerType)
     return filter(list.value, secret => {
-      return secret.metadata.dnsProviderName === dnsProviderName && isOwnSecret(secret) // secret binding not supported
+      return secret.metadata.provider?.type === providerType &&
+        (!isDnsSecret || isOwnSecret(secret)) // secret binding not supported for dns secrets
```
petersutter, line 115: I don't understand this logic

frontend/src/components/Secrets/GSecretDialog.vue
grolu: Yes, I will replace vendor with providerType in all secret dialogs

frontend/src/store/secret.js
grolu: we currently don't use factory functions for the other functions why do we want to change it for this one?

frontend/src/views/GSecrets.vue
```diff
@@ -525,8 +515,13 @@ export default {
     relatedShootCountDns (secret) {
       const shootsByDnsSecret = filter(this.shootList, shoot => {
         return some(shoot.spec.dns?.providers, {
-          type: secret.metadata.dnsProviderName,
           secretName: secret.metadata.name,
+        }) ||
```
holgerkoser, line 519:
```js
    relatedShootCountDns (secret) {
      const hasDnsProvider = (providers, secretName) => some(providers, [
        ['secretName'], 
        secretName],
      )
      const hasSecretResource = (resources, secretName) => some(resources, {
        resourceRef: { 
           kind: 'Secret', 
           name: secretName,
        },
      })

      let count = 0
      for (const shoot of this.shootList) {
        const secretName = secret.metadata.name
        if (hasDnsProvider(shoot.spec.dns?.providers, secretName) || hasSecretResource(shoot.spec.resources, secretName)) {
          count++
        }
      }
      return count
    }
```
This function is called twice in the getter`dnsSecretItems`. It should be called only once. Same for `relatedShootCountInfra` which is called twice in `infrastructureSecretItems`.

frontend/src/store/secret.js
grolu: I created a dedicated composable `useSecretList`

frontend/src/composables/useSecretList.js
```diff
@@ -0,0 +1,39 @@
...
+export const useSecretList = (value, options = {}) => {
+  const {
+    secretStore = useSecretStore(),
+    gardenerExtensionStore = useGardenerExtensionStore(),
+  } = options
+
+  return computed(() => {
+    const providerType = unref(value)
+
+    const isDnsSecret = gardenerExtensionStore.dnsProviderTypes.includes(providerType)
```
petersutter, line 28:
```suggestion
export const useSecretList = (providerType, options = {}) => {
  const {
    secretStore = useSecretStore(),
    gardenerExtensionStore = useGardenerExtensionStore(),
  } = options

  return computed(() => {
    const isDnsSecret = gardenerExtensionStore.dnsProviderTypes.includes(providerType.value)
```

petersutter (approved):
/lgtm

frontend/src/views/GSecrets.vue
```diff
@@ -109,7 +109,7 @@ SPDX-License-Identifier: Apache-2.0
       >
         <template #item="{ item }">
           <g-secret-row-infra
-            :key="`${item.cloudProfileName}/${item.name}`"
+            :key="`${item.secretName}/${item.secretNamespace}`"
```
holgerkoser, line 112: the order should be `namespace/name`

holgerkoser (approved):
/lgtm

PR #2155, opened 2024-10-29
Use shared composable instead of event propagation to distribute state

holgerkoser:
Use shared composable instead of event propagation to distribute state

Fixes #

**Release note**:

```bugfix user

```


frontend/src/components/GShootListActions.vue
```diff
@@ -5,16 +5,45 @@ SPDX-License-Identifier: Apache-2.0
...
+        <code class="text-toolbar-title">
+          {{ shootName }}
+        </code>
+        <template #append>
+          <v-btn
+            variant="text"
+            density="comfortable"
+            icon="mdi-close"
+            color="toolbar-title"
+            @click.stop="accessDialog = false"
```
petersutter, line 26:
```suggestion
            @click="accessDialog = false"
```

petersutter (approved):
I only have a minor change suggestion, otherwise /lgtm

PR #2164, opened 2024-11-04
Remove fallback to read cluster ca Secret

petersutter:
This PR removes the fallback to read the cluster ca from the `Secret`.

Fixes #

**Release note**:

```breaking operator
The dashboard no longer falls back to reading the cluster CA from the deprecated `<shoot-name>.ca-cluster` `Secret`. This change requires Gardener `v1.89.0` or higher.
```


holgerkoser (approved):
/lgtm

PR #2171, opened 2024-11-07
Drop support of static token kubeconfig

### LINKED ISSUE #1595: Static token kubeconfigs for k8s version >= 1.27
The static token kubeconfig on the `Access` card for k8s version >= 1.27 should either be hidden or disabled, as it cannot be enabled anyhow.

**Why is this needed**:

petersutter:
With this PR, static token kubeconfig support is dropped from the dashboard. This feature became obsolete with Kubernetes versions `1.27` or higher, as the `enableStaticTokenKubeconfig` field is permanently set to `false` starting with Kubernetes `1.27`.

Fixes #1595

Related change in g/g https://github.com/gardener/gardener/pull/10664

**Release note**:

```breaking user
Static token kubeconfig support has been removed from the dashboard. This feature is no longer needed as of Kubernetes version `1.27`, where the `enableStaticTokenKubeconfig` field is permanently set to `false`.
```


frontend/src/composables/useShootStatusCredentialRotation.js
```diff
@@ -73,22 +66,11 @@ export const twoStepRotationTypes = Object.freeze(filter(rotationTypes, {
...
-    shootEnableStaticTokenKubeconfig,
-    shootCredentialsRotation,
-  } = mapValues(shootPropertyMappings, args => {
-    const [path, defaultValue] = Array.isArray(args) ? args : [args]
-    return computed(() => get(state.value, path, defaultValue))
-  })
-
   options.type ??= 'ALL_CREDENTIALS'
 
+  const shootCredentialsRotation = computed(() => get(state.value, 'status.credentials.rotation', {}))
```
holgerkoser, line 72: This should be a linter error? I will find out why not.
```suggestion
  const shootCredentialsRotation = computed(() => get(state.value, ['status', 'credentials', 'rotation'], {}))
```

holgerkoser (approved):
/lgtm
beside of the lodash array syntax issue

grolu (approved):
/lgtm

PR #2180, opened 2024-11-13
Fix bug on theme change

petersutter:
This PR fixes an unreleased bug where the code mirror instance is accessed when it is not yet initialized.

Fixes #

**Release note**:

```bugfix user

```


holgerkoser (approved):
/lgtm
but we should together look at the `useShootEditor` compoable. There are some parts that could be improved.

grolu (approved):
/lgtm

PR #2186, opened 2024-11-15
Fix deprecated Sass `@import` rules

petersutter:
With `sass` version `1.80.0`, the `@import` rule has been marked as deprecated. The last remaining usage in our codebase is found [here](https://github.com/gardener/dashboard/blob/79fc067adad4a324a41a6b538c5569cf057b9045/frontend/src/sass/materialdesignicons.scss#L7-L12). However, the `@mdi/font` package (v7.4.47) still relies on the deprecated `@import` rule. To avoid issues, we have temporarily pinned the `sass` version to `1.79` until either `@mdi/font` removes its dependency on `@import` or provides a release containing only the WOFF2 font format. Once the latter happens, we can switch to importing `materialdesignicons.css` directly, eliminating the need for the current custom `font-face` workaround.

Fixes #

**Release note**:

```bugfix user

```


holgerkoser (approved):
/lgtm

grolu (approved):
/lgtm

PR #2202, opened 2024-11-26
Use a polling watcher instead of chokidar

holgerkoser:
Use a polling watcher approach instead of chokidar for watching kube-config files

Fixes #

**Release note**:

```bugfix user

```


petersutter (approved):
/lgtm

PR #2206, opened 2024-11-27
Optimize CodeMirror Integration with Dynamic Imports

holgerkoser:
This PR refactors the CodeMirror integration to use dynamic imports for all CodeMirror modules. The primary goal is to reduce the initial bundle size by deferring the loading of CodeMirror modules until they are needed.  

```bash
cd frontend
yarn build
open build/stats.treemap.html
```

**before**
before
**after**
after

Fixes #

**Release note**:

```bugfix user

```


frontend/src/composables/useShootEditor.js
```diff
@@ -0,0 +1,254 @@
...
+          }
+        },
+        doc: yaml.dump(shootItem.value),
+        extraKeys: getExtraKeys(),
+        readOnly: isReadOnly.value,
+        darkMode: isDarkMode.value,
+      })
+    } catch (err) {
+      logger.error('Failed to create codemirror instance: %s', err.message)
+      logger.error(err.stack)
```
petersutter, line 139:
```suggestion
```

frontend/src/composables/useCodemirror/index.js
```diff
@@ -0,0 +1,290 @@
...
+  dispatchEffects(view, whitespacesCompartment.reconfigure(content))
+}
+
+export function useCodemirror (element, options) {
+  const {
+    api = useApi(),
+    logger = useLogger(),
+    onDocumentChanged,
+    onTooltip,
+    extraKeys,
```
petersutter, line 118: add default values

frontend/src/composables/useShootEditor.js
```diff
@@ -0,0 +1,254 @@
...
+  function getDocumentValue () {
+    return cm?.getDocValue()
+  }
+
+  function setDocumentValue (value) {
+    cm?.setDocValue(value)
+  }
+
+  function clearDocumentHistory () {
+    cm?.clearDocHistory()
```
petersutter, line 181: call dispatchHistoryEffect directly

frontend/src/composables/useShootEditor.js
holgerkoser: ✔️

frontend/src/composables/useCodemirror/index.js
holgerkoser: ✔️

frontend/src/composables/useShootEditor.js
holgerkoser: I decided to keep the method `cm.clearDocHistory` because it is used twice, and in this particular case, the following is necessary according to @grolu.
```
cm.dispatchHistoryEffect(false)
cm.dispatchHistoryEffect(true)
```
Encapsulating this logic in a function improves code readability, prevents duplication, and reduces the risk of errors if the same logic needs to be applied in multiple places.

petersutter (approved):
/lgtm

PR #2228, opened 2024-12-10
[OpenMFP] OpenFGA SelfSubjectRulesReview implementation

petersutter:
With this PR, OpenFGA is queried to map FGA relations to RBAC resource rules.

Fixes #

**Release note**:

```bugfix user

```

### DIFF
--- a/frontend/src/router/guards.js
+++ b/frontend/src/router/guards.js
@@ -133,8 +133,11 @@ export function createGlobalResolveGuards () {
       }
 
       try {
+        const openMFP = useOpenMFP()
+        const accountId = openMFP.accountId.value ?? to.query.accountId
+
         const namespace = to.params.namespace ?? to.query.namespace
-        await refreshRules(authzStore, namespace)
+        await refreshRules(authzStore, namespace, accountId)
 
         if (namespace && namespace !== '_all' && !projectStore.namespaces.includes(namespace)) {
           authzStore.$reset()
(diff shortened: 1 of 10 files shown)


frontend/src/router/guards.js
```diff
@@ -133,8 +133,11 @@ export function createGlobalResolveGuards () {
       }
 
       try {
+        const openMFP = useOpenMFP()
+        const accountId = openMFP.accountId.value ?? to.query.accountId
```
holgerkoser, line 137:
```suggestion
        const { accountId } = useOpenMFP()
```

frontend/src/router/guards.js
```diff
@@ -133,8 +133,11 @@ export function createGlobalResolveGuards () {
...
+
         const namespace = to.params.namespace ?? to.query.namespace
-        await refreshRules(authzStore, namespace)
+        await refreshRules(authzStore, namespace, accountId)
```
holgerkoser, line 140:
```suggestion
        await refreshRules(authzStore, namespace, accountId.value ?? to.query.accountId)
```

frontend/src/router/guards.js
petersutter: ✅

frontend/src/router/guards.js
petersutter: ✅

holgerkoser (approved):
/lgtm

PR #2274, opened 2025-01-23
Added tests for internal vuetify classes that we rely on for styling

### LINKED ISSUE #2270: Ensure that we have tests for all internal Vuetify CSS classes
In some cases, it is necessary to rely on internal Vuetify CSS class structures to apply custom styling. In these cases we should have unit tests which ensure that those classes remain stable in future Vuetify releases. However, we currently have this only for some classes.

**Why is this needed**:
Ensure that Vuetify upgrades do not break our css

grolu:

Fixes #2270

**Release note**:

```bugfix user
NONE
```


petersutter (approved):
/lgtm

PR #2280, opened 2025-01-29
fix setError caller

petersutter:
`setError` is not called properly and thus (detail) information is missing when a error notification is shown.

Fixes #

**Release note**:

```bugfix user
Fixed missing details on error notification
```
```feature user
Notifications will remain visible as long as the mouse hovers over them, rather than disappearing after 5 seconds.
```

### DIFF
--- a/frontend/src/layouts/GLogin.vue
+++ b/frontend/src/layouts/GLogin.vue
@@ -184,10 +184,7 @@ export default {
     if (this.error) {
       const err = this.error
       this.error = null
-      this.setError({
-        title: err.title,
-        text: err.message,
-      })
+      this.setError(err)
     }
     next()
   },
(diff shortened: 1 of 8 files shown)


frontend/src/layouts/GLogin.vue
```diff
@@ -184,7 +184,10 @@ export default {
     if (this.error) {
       const err = this.error
       this.error = null
-      this.setError(err)
+      this.setError({
```
grolu, line 187: while this fixes the issue, shouldn't we enhance the `setAlertWithType` in `appStore` to also support the `message` property as alternative to `text`. This ensures that the function works with error objects and prevents possible wrong usage in the future

petersutter: passing an error object is now supported

grolu (approved):
/lgtm

PR #2331, opened 2025-02-21
Fix generic "box icon" tooltip

### LINKED ISSUE #2261: improve "box icon" tooltip on certain SVGs
**What happened**:
When hovering over the connection status or over the certified kubernetes logo in the shoot list, the tooltip "box icon" is shown. 
https://github.com/gardener/dashboard/blob/1c7c437965aed84c7558bdb2e678e05a0d8d89d6/frontend/src/components/icons/GIconBase.vue#L16-L19
Image

Show proper tooltip. When the GIconBase component is used, also the `icon-name` property should be passed so that the default "box" is not used.
We should remove the title if no `icon-name` is passed to prevent showing an unnecessary tooltip.

**Environment**:

petersutter:
The following changes were introduced with this PR:
- Fixed the issue where the connection icon was missing in the subscription progress state
- Corrected the generic "box icon" tooltip
- A tooltip is now displayed for the subscription state icon

Fixes #2261

**Release note**:

```bugfix user
Minor icon fixes:
- Fixed the issue where the connection icon was missing in the subscription progress state
- Corrected the generic "box icon" tooltip
- A tooltip is now displayed for the subscription state icon
```

### DIFF
--- a/frontend/src/components/GShootSubscriptionStatus.vue
+++ b/frontend/src/components/GShootSubscriptionStatus.vue
@@ -34,6 +34,7 @@ SPDX-License-Identifier: Apache-2.0
           <g-icon-base
             :width="16"
             :icon-color="color"
+            :icon-name="iconName"
           >
             <component :is="resolveComponent(iconName)" />
           </g-icon-base>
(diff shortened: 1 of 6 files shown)


grolu (approved):
/lgtm

PR #2339, opened 2025-02-28
Fix missing Terminal Shortcuts for "shoot" and "cp" scopes.

Lappihuan:
This fixes a bug that broke all "shoot" and "cp" scope while refactoring the state management:
https://github.com/gardener/dashboard/commit/4b4c5107a9adbd328243788f908775c5ae184b49#diff-abeb7e9d3448e309c3fc4a7cc959611852d5242b0fcf58540b063f494b131b04L41

The prop `shootItem` used to be passed to `GTerminalShortcuts` component, this was removed but the component still relies on the object to be present for it to include the terminal shortcuts as seen here:
https://github.com/gardener/dashboard/blob/2b0f0aed793ea3fad80d6de6f7b3b682d8475b57/frontend/src/components/GTerminalShortcuts.vue#L137-L139

But `this.shootItem` was never set anymore in any parent component as prop:
https://github.com/search?q=repo%3Agardener%2Fdashboard+%22%3Cg-terminal-shortcuts%22&type=code

this PR get the `shootItem` via the `useShootItem` composable the same way as in the `GTerminalShortcut.vue`:
https://github.com/gardener/dashboard/commit/4b4c5107a9adbd328243788f908775c5ae184b49#diff-3925a06cd814d08ed6d969e9c3553128a5b66d2b5415c735374bd5b1fe2315efR132-R142

With this change all three scopes "garden", "cp" and "shoot" shortcuts are shown again correctly

**Release note**:
```bugfix user
fixes missing terminal shortcuts for "shoot" and "cp" scopes
```


gardener-robot-ci-2:

frontend/src/components/GTerminalShortcut.vue
```diff
@@ -129,9 +129,9 @@ export default {
   ],
   setup () {
     const {
-      shootItem,
+      shootItem = null,
```
petersutter, line 132:
```suggestion
      shootItem,
```

frontend/src/components/GTerminalShortcut.vue
```diff
@@ -129,9 +129,9 @@ export default {
...
       isShootStatusHibernated,
-    } = useShootItem()
+    } = useShootItem() || {}
```
petersutter, line 134:
```suggestion
    } = useShootItem()
```

frontend/src/components/GTerminalShortcuts.vue
```diff
@@ -121,16 +123,22 @@ export default {
...
       type: String,
     },
   },
   emits: [
     'addTerminalShortcut',
   ],
+  setup () {
+    const {
+      shootItem = null,
+    } = useShootItem() || {}
```
petersutter, line 136:
```suggestion
      shootItem,
    } = useShootItem()
```

frontend/src/components/GTerminalShortcut.vue
Lappihuan: this would break the context when coming from the button "GARDEN CLUSTER", since then `useShootItem` seems to return `null`.
```
GTerminalShortcuts.vue:135 Uncaught (in promise) TypeError: Cannot destructure property 'shootItem' of 'useShootItem(...)' as it is null.
```
this just tries to handle this gracefully so we stil lend up with `shootItem` as `null`.

there might be a better way to solve it in `useShootItem` to still return a result object instead of just `null`

frontend/src/components/GTerminalShortcut.vue
Lappihuan: same as above

frontend/src/components/GTerminalShortcuts.vue
Lappihuan: same as above

frontend/src/components/GTerminalShortcut.vue
Lappihuan: this seems to come from this default value here:
https://github.com/gardener/dashboard/blob/97c13571dc2199d32ca1dd3be31fe9a4867d21e8/frontend/src/composables/useShootItem.js#L264-L266

petersutter: ah I see, then how about adding a comment like the following to make it clear that the shoot-item can be absent in valid scenarios. WDYT?

```js
} = useShootItem() || {} // shoot-item is not provided in case of GardenTerminal
```

Lappihuan: definitly! i agree that its quite obscure and not super readable.
i will add the comments

petersutter (approved):
/lgtm, thank you!

PR #2353, opened 2025-03-11
oidc-use-pkce deprecation

benedikt-haug:
Currently, kubectl works fine, but complains about the following:
```
Flag --oidc-use-pkce has been deprecated, use --oidc-pkce-method instead. For the most providers, you don't need to set the flag.
```

None, should I open a separate one?

I didn't test this code, my intent is to get rid of the following line in the kubeconfigs, that are generated:
```
          - '--oidc-use-pkce'
```

**Release note**:

```bugfix user
Flag --oidc-use-pkce has been deprecated. In case your use-case requires it, use --oidc-pkce-method instead. For the most providers, you don't need to set the flag.
```

### DIFF
--- a/frontend/src/views/GAccount.vue
+++ b/frontend/src/views/GAccount.vue
@@ -378,9 +378,6 @@ export default {
           args.push('--oidc-extra-scope=' + scope)
         }
       }
-      if (oidc.usePKCE || !oidc.clientSecret) {
-        args.push('--oidc-use-pkce')
-      }
       if (oidc.certificateAuthorityData) {
         args.push('--certificate-authority-data=' + oidc.certificateAuthorityData)
       } else if (oidc.insecureSkipTlsVerify) {


gardener-robot-ci-2:
Thank you @benedikt-haug for your contribution. Before I can start building your PR, a member of the organization must set the required label(s) {'reviewed/ok-to-test'}. Once started, you can check the build status in the PR checks section below.

petersutter:
Hi @benedikt-haug, thanks for the PR. 
Why not always set/enforce the `--oidc-pkce-method` flag? I'd rather prefer this, similar to like it was previously. WDYT?

Lappihuan (changes requested):
i can't add a suggestion on deleted lines so i added it on the line above

benedikt-haug:
fair, kind of forgot about this merge to be honest ^^

petersutter (approved):
/lgtm, thanks

grolu (approved):
/lgtm

PR #2357, opened 2025-03-12
Fixed some issues caused by ProjectPlaceholder

grolu:
Fixed ProjectPlaceholder must not be rendered while navigating to All Projects Cluster Page -> Breaks ShootList initialization
Fixed useRoute() must not be called from async context -> Breaks terminal splitpane restore
Explanation:
> When a component is created, Vue sets up a “current instance” that composable functions (like useRoute()) use to access context (e.g., injections, reactivity). This instance is only available while executing the component’s setup. When you use await, the function execution pauses. By the time it resumes, the synchronous setup context may no longer be active (or might not be correctly re-established). This means calling useRoute() after an await can fail because it can’t find the current component instance.

Example: https://codesandbox.io/p/sandbox/autumn-surf-ykyl4l

Fixes #

**Release note**:

```bugfix user
NONE
```


petersutter (approved):
/lgtm

Issue #1417, opened 2023-02-07
Show `Seed`s and `Garden` in Dashboard


grolu:
**What would you like to be added**:
Have a page with a list of seeds as we have for the shoots. Maybe this list should only be visible to operators (even if users can read seed resources)
We already have the seeds in the dashboard, however we would need to watch them to make the seed status reactive.
Maybe it also makes sense to read the ManagedSeed resources to identify the associated shoot. Right now this is done via a name assumption which might not always work depending on the landscape setup.
If we have this information in the dashboard, the link to seed will also work from the cluster list.
The link will lead to the seed resource, but we should also offer a way to switch to shoot (if available) and back. Also add information that a shoot belongs to a managed seed to the shoot details page.

See also
- https://github.com/gardener/dashboard/issues/998
- https://github.com/gardener/dashboard/issues/1480

**Why is this needed**:
Get additional information about seeds, including their status
Some seeds are not managed, so they can not be accessed via their shoot resource

rfranzke:
We could also show the `operator.gardener.cloud/v1alpha1.Garden` resource. It is similar like a `Seed`/`Shoot` and has conditions that would be interesting for human operators.

PR #2358, opened 2025-03-17
Disable terminal scheduling on non-managed seeds

### LINKED ISSUE #2170: Disable terminal scheduling on non-managed seeds in the dashboard
The dashboard should disable or hide the option to schedule a terminal on a `Seed` if it is not a managed seed. The frontend should detect whether a `Seed` is managed and adjust the available options accordingly.

**Why is this needed**:
Currently, users may encounter the error "terminal cannot be hosted on non-managed seed" (introduced with https://github.com/gardener/dashboard/pull/2169) when trying to schedule a terminal on an unmanaged `Seed`. Disabling or hiding this option upfront would prevent this error and improve the user experience.

Lappihuan:
This disables the terminal on shoots that are not on a managed seed, using the suggestion from https://github.com/gardener/dashboard/issues/2170#issuecomment-2460045962

Fixes #2170

I'm unsure why there was a check on the garden namespace, but this excluded all shoots in the garden project.
https://github.com/gardener/dashboard/pull/2358/files#diff-6ed2ba0f6d68010cb296caa1dfe39da1cd2acf7dabd973d99fdf670d5849907fL307-R307
This might cause a bug elsewhere.

**Release note**:
```bugfix user
Disable terminal scheduling on non-managed seeds
```


gardener-robot-ci-1:

grolu (dismissed):
/lgtm

Removing `namespace !== 'garden'` should not break anything. If a seed is not managed, canLinkToSeed will be false. It was just an assumption that shoot resources in garden namespace will not have a seed that is managed but it is fair to remove it as this could be the case.
Also, using `canLinkToSeed` as an indicator for the terminal that the seed is managed is considered a workaround. It would be better to actually read the ManagedSeed resources in the backend and add a proper field to the shoot info. But we can target this in a separate PR.

Lappihuan:
> It would be better to actually read the ManagedSeed resources in the backend and add a proper field to the shoot info.
It seems this is the way forward then.

So it depends somewhat on #1417

grolu:
>It seems this is the way forward then.
> So it depends on https://github.com/gardener/dashboard/issues/1417

So yes, in the long run this is the preferred solution. @petersutter already started working on watching seeds in the dashboard frontend. However, I cannot give an estimation on the timeline as we are currently blocked with other things.
However, you could still try to build an intermediate solution using `canLinkToSeed`. The important thing is that you don't disable the button to open the dialog but rather disable the options to schedule a terminal on the seed. This needs to be done on the `GCreateTerminalSessionDialog` and a tooltip message should be provided informing the operator, why he can only schedule a terminal on the shoot cluster and not on the control plane (seed). Also, the Terminal Runtime needs to be switched to Cluster.
I also noticed a couple of small issues in this context. I will have a look at it.

Lappihuan:
@grolu is this now superseded by #2373 so i can close this PR?

grolu:
>@grolu is this now superseded by https://github.com/gardener/dashboard/pull/2373 so i can close this PR?

Yes, I hope that we can merge this PR. But there are still some thinks to clarify first :)

PR #2359, opened 2025-03-18
extend available data in url expansion for external tools

Lappihuan:
This allows for example to create urls within the cluster domain:
```yaml
- title: Example Tool
  icon: link
  url: https://myapp.{domain}
```
For example if shoots come predeployed with apps.

```bugfix user
extend available data in url expansion for external tools
```


gardener-robot-ci-1:

grolu:
Hi @Lappihuan 
Thank you for your contribution. We understand your use case to include the `shootDomain` in the template. This change is fine for us. However, can you explain the reason for exposing the other shoot values? Especially `availableK8sUpdates` and `isNewCluster` would expose internal state / structures that we then need to keep stable. Moreover, `isNewCluster` can be derived from the `creationTimeStamp` which is exposed via the `metadata` object.
Speaking of `metadata`: Right now, we officially only support `namespace` / `name` in the template (ref: https://github.com/gardener/dashboard/blob/master/charts/gardener-dashboard/values.yaml#L350). It would be great, if you could limit the available `metadata` properties by picking only namespace, name (or whatever else you need) from the object. This is more explicit and avoid polluting the template. Please also update the comment in the chart values (add the additional available values like `shootDomain`).
Thank you!

frontend/src/components/ShootDetails/GShootExternalToolsCard.vue
```diff
@@ -57,15 +57,26 @@ const configStore = useConfigStore()
...
 } = useShootItem()
 
 const items = computed(() => {
   return configStore.externalTools ?? []
 })
 
+const urlData = computed(() => {
+  return {
+    name: shootMetadata.value.name,
+    namespace: shootMetadata.value.namespace,
```
petersutter, line 71: use `shootNamespace` and `shootName` instead of  `shootMetadata`

Lappihuan: thanks, changed it

grolu (approved):
/lgtm

petersutter (approved):
/lgtm, thanks!

PR #2423, opened 2025-05-06
[master] No fallback secret details rendering for unknown provider types

grolu:

Fixes #

**Release note**:

```bugfix user

```

### DIFF
--- a/frontend/src/components/Credentials/GCredentialDetailsItemContent.vue
+++ b/frontend/src/components/Credentials/GCredentialDetailsItemContent.vue
@@ -71,7 +71,10 @@ export default {
         ]
       }
       if (this.credential?.kind === 'Secret') {
-        return this.getCredentialDetails(this.credential)
+        const details = this.getCredentialDetails(this.credential)
+        if (details) {
+          return details
+        }
       }
       if (this.credential?.kind === 'WorkloadIdentity') {
         return [
@@ -84,8 +87,9 @@ export default {
       }
       return [
         {
-          label: 'Unknown',
-          value: 'Information not available',
+          label: this.credential?.kind || 'Unknown',
+          value: `No details available for credentials of type ${this.providerType}`,
+          valueClass: 'font-weight-light text-disabled',
         },
       ]
     },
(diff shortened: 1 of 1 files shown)


frontend/src/components/Credentials/GCredentialDetailsItemContent.vue
```diff
@@ -84,8 +87,9 @@ export default {
       }
       return [
         {
-          label: 'Unknown',
-          value: 'Information not available',
+          label: this.credential?.kind || 'Unknown',
+          value: `No details available for credentials of type ${this.providerType}`,
```
petersutter, line 91: the following should be sufficient and also aligns with the other message that we display for shared credentials, which is "Details not available for shared credentials"
```suggestion
          value: 'Details not available',
```

frontend/src/components/Credentials/GCredentialDetailsItemContent.vue
```diff
@@ -161,7 +165,7 @@ export default {
             return [
               {
                 label: 'Hetzner Cloud Token',
-                value: decodeBase64(secretData.hcloudToken),
+                value: 'hidden',
```
petersutter, line 168: what is the benefit of having the value "hidden" (which is not displayed with `font-weight-light text-disabled` class), vs. letting it run into the default case and return undefined.

frontend/src/components/Credentials/GCredentialDetailsItemContent.vue
```diff
@@ -71,7 +71,10 @@ export default {
         ]
       }
       if (this.credential?.kind === 'Secret') {
-        return this.getCredentialDetails(this.credential)
+        const details = this.getCredentialDetails(this.credential)
+        if (details) {
+          return details
+        }
       }
       if (this.credential?.kind === 'WorkloadIdentity') {
```
petersutter, line 79: shouldn't this be rather an else-if / switch statement?
(I know this code was not touched, but I thought I gave this feedback already in the previous CredentialBinding support PR)?

petersutter (approved):
/lgtm

PR #2429, opened 2025-05-09
Fixed GSecretDialogGeneric [master]

grolu:

Fixes #

**Release note**:

```bugfix user
Fixed Generic Secret Dialog issues
- Dialog now correctly writes key-value pairs at the top level
- Resolved an issue where data was not available in update mode
```


frontend/src/components/Credentials/GSecretDialogGeneric.vue
```diff
@@ -116,39 +109,41 @@ export default {
...
-
-      this.data = ''
-      this.secretData = {}
-
-      if (!this.isCreateMode) {
-        setDelayedInputFocus(this, 'data')
+  watch: {
+    stringData (value) {
+      if (!this.touched) {
+        this.internalYaml = yaml.dump(value)
```
petersutter, line 142: adding a secret with 
```yaml
foo: ~
```
and pressing save will result in
```yaml
foo: 'null'
```

grolu: fixed

PR #2441, opened 2025-05-15
[master] Use static row height

grolu:
Use static row height as a workaround to fix lazy row rendering in shoot table

Fixes #

**Release note**:

```bugfix user
Lazy rendering was not functioning correctly in some scenarios, leading to performance issues with large cluster lists
```


petersutter (approved):
/lgtm

PR #2444, opened 2025-05-16
End to end reactivity for `Seed` resources

petersutter:
This PR makes the seed resource reactive so that we receive updates via websocket connection.
In the shoot list, you can now see the conditions of the corresponding seed cluster in the `Seed Readiness` column. For now, this is visible only for gardener operators.

Fixes #

**Release note**:

```feature operator
Real-time updates for seeds are now supported. In the shoot list, you can now see the conditions of the corresponding seed cluster in the `Seed Readiness` column. The `Seed` column has also been moved close to the newly introduced `Seed Readiness` column.
```


backend/lib/watches/seeds.js
```diff
@@ -0,0 +1,41 @@
...
+      }
+    }
+    for (const user of users.values()) {
+      const canListSeeds = get(user, ['profiles', 'canListSeeds'], false)
+      if (!canListSeeds) {
+        continue
+      }
+      const event = { uid, type }
+      const room = ioHelper.sha256(user.id)
+      nsp.to(room).emit('seeds', event)
```
petersutter: check if we want to have a separate seeds room

frontend/src/store/shoot/helper.js
```diff
@@ -394,7 +435,19 @@ export function sortItemsFn (state, context) {
...
-    return orderBy(items, [iteratee, 'metadata.name'], [order, 'asc'])
+
+    let sortKeys
+    let sortOrders
+    if (key === 'seedReadiness') {
+      // second sort key is the seed name
+      sortKeys = [iteratee, 'spec.seedName', 'metadata.name']
+      sortOrders = [order, 'asc', 'asc']
+    } else {
+      sortKeys = [iteratee, 'metadata.name']
```
petersutter: use array syntax for path

backend/lib/watches/seeds.js
petersutter: addressed with https://github.com/gardener/dashboard/pull/2444/commits/dd2de09fae697680412cf3078af48d006da5a14f

frontend/src/components/GSeedStatusTag.vue
grolu, line 1: As discussed, we should merge with existing components but not part of this PR, I created an issue for this: https://github.com/gardener/dashboard/issues/2534

frontend/src/views/GShootList.vue
```diff
@@ -517,6 +509,23 @@ export default {
...
+          align: 'start',
+          defaultSelected: false,
+          hidden: false,
+        },
+        {
+          title: 'SEED READINESS',
+          key: 'seedReadiness',
+          sortable: isSortable(true),
+          align: 'start',
+          defaultSelected: true,
```
grolu, line 525: shouldn't wee align `defaultSelected` Seed and Seed Readiness? Maybe we could do `defaultSelected: isAdmin` as we do for other columns. I know, seed readniness is only visible for admins (operators) but in this case I think it makes also sense to see the seed column next to it. wdyt?

petersutter: with https://github.com/gardener/dashboard/pull/2444/commits/4ca1dad3731a355f1dc01925c51336eff5f422ec the seed name is now visible in the status tag toolbar title. That's why the seed column can stay `defaultSelected: false`

grolu (approved):
/lgtm

backend/__fixtures__/seeds.cjs
```diff
@@ -41,6 +40,9 @@ function getSeed ({
       },
     },
   }
+  if (!isEmpty(labels)) {
```
klocke-io: for clarity would it make sense if we remove the [fallback value for labels in the function definition](https://github.com/gardener/dashboard/pull/2444/files#diff-69d3c31bb45b0dd4e038ee74f68cdf0971ab9ece2a950565a7fc028450be03fcL18) and then drop the check? 


```suggestion
```

backend/__tests__/acceptance/io.spec.cjs
```diff
@@ -128,7 +128,8 @@ describe('api', function () {
       let args
 
       beforeEach(async () => {
-        // authorization check for `canListProjects`
+        // authorization check for `canListProjects` and `canListSeeds`
+        mockRequest.mockImplementationOnce(fixtures.auth.mocks.reviewSelfSubjectAccess())
```
klocke-io: Consider chaining the `.mockImplementationOnce` call, since at least I thought that this line was accidentally duplicated 😁

```suggestion
        mockRequest
        	.mockImplementationOnce(fixtures.auth.mocks.reviewSelfSubjectAccess())
        	.mockImplementationOnce(fixtures.auth.mocks.reviewSelfSubjectAccess())
```

backend/lib/services/seeds.js
```diff
@@ -8,40 +8,19 @@ import _ from 'lodash-es'
...
   if (!allowed) {
     throw new Forbidden('You are not allowed to list seeds')
   }
 
-  const seeds = getSeeds()
-  return _.map(seeds, fromResource)
+  return _
+    .chain(getSeeds())
+    .map(_.cloneDeep)
+    .map(simplifySeed)
```
klocke-io, line 24: why not let the consumer decide if he wants `simplifySeed`'s or a normal ones? I also would not suspect to get simplified seeds by calling seed.list() instead the consumer could then simplify his list of seeds if he desires to do so.

frontend/src/components/GShootMessageDetails.vue
```diff
@@ -82,17 +82,19 @@ SPDX-License-Identifier: Apache-2.0
...
-                <h4 class="wrap-text font-weight-bold">
-                  This error is flagged as user error which indicates that no Gardener operator action is required.
-                  Please read the error message carefully and take action.
-                </h4>
+                <template v-if="userError">
+                  <h3>
+                    Your Action is required
+                  </h3>
+                  <h4 class="wrap-text font-weight-bold">
+                    This error is flagged as user error which indicates that no Gardener operator action is required.
```
klocke-io: I kinda find the messaging inconsistent. Your action is required? But then there is no action required by the Gardener operator?

frontend/src/store/shoot/helper.js
```diff
@@ -245,11 +245,37 @@ export function getRawVal (context, item, column) {
...
+function getReadinessSortVal ({
+  conditions,
+  lastOperationTime,
+  creationTime,
+  isErrorFn,
+  configStore,
+}) {
+  if (!conditions.length) {
+    // items without conditions have medium priority
+    const priority = '00000100'
```
klocke-io, line 257: maybe we could extract the priority values to a dedicated file as const to add meaning to them and ensure consistency.

klocke-io (approved):
I added a couple of comments, nothing blocking but maybe worth while considering

backend/lib/services/seeds.js
petersutter: agree, but it's not needed right now since the simplified seeds (and simplify project for the project-list case) cover our primary use cases efficiently without added complexity. We can revisit again once we have this requirement

frontend/src/store/shoot/helper.js
petersutter: this part of the code was refactored into a function, nothing newly introduced. I'd rather recommend to overhaul the complete getRawVal and getSortVal logic, but nothing that should be touched in this PR

backend/__tests__/acceptance/io.spec.cjs
petersutter: ✅

backend/__fixtures__/seeds.cjs
petersutter: ✅

frontend/src/components/GShootMessageDetails.vue
petersutter: I have improved the message. Please check if this clarifies it

PR #2448, opened 2025-05-19
Potential fix for code scanning alert no. 1: Loop bound injection

petersutter:
Potential fix for [https://github.com/gardener/dashboard/security/code-scanning/1](https://github.com/gardener/dashboard/security/code-scanning/1)

To fix the issue, we need to validate the `roles` array to ensure it is both an array and has a reasonable size limit before it is used in the loop. This can be achieved by:
1. Checking if `roles` is an array using `Array.isArray`.
2. Limiting the maximum allowed size of the `roles` array to a predefined constant (e.g., `MAX_ROLES`).
3. Throwing an error if the validation fails.

The validation should be added in the `setItemRoles` method in `MemberManager.js`, as this is where the `roles` parameter is directly used.

---

_Suggested fixes powered by Copilot Autofix. Review carefully before merging._

### DIFF
--- a/backend/lib/services/members/MemberManager.js
+++ b/backend/lib/services/members/MemberManager.js
@@ -179,6 +179,13 @@ class MemberManager {
   }
 
   setItemRoles (item, roles) {
+    if (!Array.isArray(roles)) {
+      throw new UnprocessableEntity('Roles must be an array')
+    }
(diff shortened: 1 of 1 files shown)


grolu (approved):
/lgtm

PR #2196, opened 2024-11-21
Handle removal of legacy access restriction

petersutter:
**What this PR does / why we need it**:
With [gardener/gardener#10883](https://github.com/gardener/gardener/pull/10883), Gardener no longer automatically removes legacy fields when `.spec.accessRestrictions` is deleted. This behavior specifically applies to the `eu-access-only` access restriction. To fully disable the `eu-access-only` restriction for a `Shoot`, the legacy fields must also be explicitly removed. This PR implements the necessary changes to handle this scenario.

Additionally, this PR updates the access restriction configuration of the dashboard at runtime by mapping `accessRestriction.items[@key="seed.gardener.cloud/eu-access"]` to `accessRestriction.items[@key="eu-access-only"]`. This mapping will be removed in a future version of the dashboard and operators are expected to adapt their config accordingly, if applicable.

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Release note**:

```other operator
**Action Required:** The automatic mapping from `accessRestriction.items[@key="seed.gardener.cloud/eu-access"]` to `accessRestriction.items[@key="eu-access-only"]` will be removed in a future dashboard version. Dashboard operators who use `seed.gardener.cloud/eu-access` in their access restriction configuration should update the key to `eu-access-only`. If your configuration does not include `seed.gardener.cloud/eu-access`, no action is required.
```

PR #2452, opened 2025-05-20
Cleanup legacy EU access-only sync mechanism and config mapping

petersutter:
This PR removes the temporary mapping that was introduced in a previous PR ([ref](https://github.com/gardener/dashboard/pull/2196)) where `accessRestriction.items[@key="seed.gardener.cloud/eu-access"]` was mapped to `accessRestriction.items[@key="eu-access-only"]`. As announced in the original PR, this mapping was meant to be temporary to give operators time to update their configurations. The grace period has now passed, and we are proceeding with the removal as planned.

The removal of this mapping completes the transition that was started with the previous changes, and operators should now be using the `eu-access-only` key directly in their access restriction configurations as previously advised.

Fixes #

This change follows through on the previously communicated plan to remove the temporary mapping. Operators were given advance notice in the previous PR that this change would be coming.

**Release note**:

```breaking operator
The temporary mapping from `accessRestriction.items[@key="seed.gardener.cloud/eu-access"]` to `accessRestriction.items[@key="eu-access-only"]` has been removed as previously announced ([ref](https://github.com/gardener/dashboard/pull/2196)).
```

### DIFF
--- a/frontend/src/composables/useShootAccessRestrictions/index.js
+++ b/frontend/src/composables/useShootAccessRestrictions/index.js
@@ -4,17 +4,13 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
-import {
-  computed,
-  ref,
-} from 'vue'
+import { computed } from 'vue'
 
 import { useCloudProfileStore } from '@/store/cloudProfile'
 
 import { NAND } from './helper'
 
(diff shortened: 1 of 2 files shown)


grolu (approved):
/lgtm

PR #2454, opened 2025-05-21
Show cloudprofiles without matching seed

grolu:

Fixes #

**Release note**:

```other operator
Removed the logic that filters cloud profiles without a matching seed in the backend. All cloud profiles will now show up in the frontend. If a cloud profile without a matching seed is selected for a new cluster, a hint will be shown to the user
```


frontend/src/components/GSelectCloudProfile.vue
```diff
@@ -108,5 +113,17 @@ const rules = {
     required,
   }),
 }
+
+const selectedCloudProfile = computed(() => {
+  return find(props.cloudProfiles, { metadata: { name: props.modelValue?.name } })
```
petersutter, line 118: use kind of modelValue to determine in which list to search for

frontend/src/components/NewShoot/GNewShootInfrastructureDetails.vue
```diff
@@ -8,7 +8,7 @@ SPDX-License-Identifier: Apache-2.0
   <v-container class="px-0 mx-0">
     <v-row>
       <v-col
-        v-if="cloudProfiles.length > 1"
+        v-if="cloudProfiles.length > 1 || !cloudProfileHasSeedNames"
```
petersutter, line 11: why do we need this change here?

grolu: If a cloudprovider has only one profile we want to show the dropdown anyway in case there is no seed as this is the only place where we show the hint about missing seeds (hint). If you have a better idea please let me know :)

grolu: ok, done

frontend/src/views/GNewShoot.vue
```diff
@@ -122,7 +122,7 @@ SPDX-License-Identifier: Apache-2.0
     class="ma-3"
     type="warning"
   >
-    There must be at least one cloud profile supported by the dashboard as well as a seed that matches it's requirements.
+    There must be at least one cloud profile that is supported by the dashboard and visible (not hidden by the dashboard vendor configuration).
```
petersutter: "(not hidden by the dashboard vendor configuration)" might be covered by "that is supported by the dashboard".
```suggestion
    There must be at least one cloud profile that is supported by the dashboard and visible.
```
Not sure if users understand what "and visible" means. Maybe marked / flagged as visible might help?

PR #2476, opened 2025-06-04
Extend the defaults accepted via configuration

benedikt-haug:
This PR is part of the Hackaton.
This PR extend the defaults gardener dashboard accepts via configuration.
Additionally, it moves the default declaration to the config store so they are declared in the same spot.
Lastly, it also adds the example folder describing the new additions made possible by the gardener-dashboard-frontend configmap.

-

/cc @marc1404 @klocke-io 

**Release note**:

```feature operator
New `shootDefaults` Configuration: A new frontend configuration namespace, `shootDefaults`, has been introduced. It allows predefined default values to be applied when users create new Shoot resources in the Gardener Dashboard.

The following existing configuration properties have been moved to the new namespace:

* `controlPlaneHighAvailabilityHelp` → `shootDefaults.controlPlaneHighAvailabilityHelp`
* `defaultHibernationSchedule` → `shootDefaults.hibernationSchedule`
* `defaultNodesCIDR` → `shootDefaults.nodesCIDR`

The previous configuration paths are now deprecated. Please update your deployments to use the corresponding properties under `shootDefaults`.
```

## Summary by CodeRabbit

* **New Features**
  * Added unified `shootDefaults` support for dashboard/frontend defaults (hibernation, maintenance, autoscaling, workerless, container runtime, purposes, networking, and help text), with legacy compatibility.
  * Updated dashboard ConfigMap templating to emit `shootDefaults` and validate `controlPlaneHighAvailabilityHelp.text`.
* **Bug Fixes**
  * Improved help-text sanitization to strip unsafe HTML/script content while preserving expected formatting/legacy markers.
  * Corrected provider-specific infrastructure rendering/validation so requirements apply only when workerless is off.
  * Refined wildcard/custom-wildcard handling and related emitted state.
* **Documentation**
  * Updated examples and local setup docs to use the new `shootDefaults` layout.
* **Tests**
  * Expanded unit test coverage for sanitization, default precedence, templating validation, and conditional UI/validation behavior.


grolu:
Maybe it would be better to group the default options under a key in the config, e.g., `shootDefaults`. This way, we can ensure the config remains clear and all shoot default settings are grouped together. These settings are also likely to grow, so this approach gives us better control.

Additionally, these values need to be added to the Helm charts. Right now, the Helm charts are the only documentation for the dashboard configuration. We should consider introducing a dedicated documentation page for the configuration — but that’s a separate task and not related to this PR.

benedikt-haug:
> Nice! Please also have a look at this PR: #2467 Especially the `defaultInfrastructures` configuration may cause conflicts.

Sure <3
On quick glance it looks like all it would take to allow user overrides for your new knownInfraVendors array would be to move the defaulting function to store/config.js, just like this PR does for all the other defaults.



> Maybe it would be better to group the default options under a key in the config, e.g., `shootDefaults`. This way, we can ensure the config remains clear and all shoot default settings are grouped together. These settings are also likely to grow, so this approach gives us better control.

Seems reasonable :) Incorporated that feedback. Also moved the "old" variables there, while including a fallback for the users of the old naming scheme so things don't suddenly break.

> Additionally, these values need to be added to the Helm charts. Right now, the Helm charts are the only documentation for the dashboard configuration. We should consider introducing a dedicated documentation page for the configuration — but that’s a separate task and not related to this PR.

Added an example like the other repos have it. Also tried to extend the helm charts to feature the new variable. Do you know how to test the charts locally, `helm template .` ain't working for me

frontend/src/store/config.js
```diff
@@ -230,19 +230,19 @@ export const useConfigStore = defineStore('config', () => {
   })
 
   const controlPlaneHighAvailabilityHelp = computed(() => {
-    return state.value?.shootDefaults.controlPlaneHighAvailabilityHelp ?? state.value?.controlPlaneHighAvailabilityHelp
+    return state.value?.shootDefaults?.controlPlaneHighAvailabilityHelp ?? state.value?.controlPlaneHighAvailabilityHelp
```
grolu: we should introduce a `shootDefaults` computed that defaults to `{} ` to reduce optional chaining.

petersutter: ping @benedikt-haug

benedikt-haug: sry for taking this long to respond.
tried incorporating the feedback

grolu: I saw that you added a default to the helm chart. However, the better approach would be to add a computed property like this:
```js
const shootDefaults = computed(() => {
  return state.value?.shootDefaults ?? {}
})
```
This way we have the defaulting also in case of operator deployment. Also, this shortens all your other computed logic as you can do like
```js
const defaultHibernationSchedule = computed(() => {
  return shootDefaults.hibernationSchedule ?? state.value?.defaultHibernationSchedule
})
```
for all the computed getters.

charts/gardener-dashboard/charts/runtime/templates/dashboard/configmap.yaml
```diff
@@ -297,6 +302,8 @@ data:
         {{- if .Values.global.dashboard.frontendConfig.shootDefaults.autoUpdateKubernetes}}
         autoUpdateKubernetes: {{ .Values.global.dashboard.frontendConfig.shootDefaults.autoUpdateKubernetes }}
         {{- end }}
+      {{- else }}
+      shootDefaults: {}
```
grolu:
```suggestion
```
See comment

frontend/src/store/config.js
```diff
@@ -230,19 +236,19 @@ export const useConfigStore = defineStore('config', () => {
   })
 
   const controlPlaneHighAvailabilityHelp = computed(() => {
-    return state.value?.shootDefaults?.controlPlaneHighAvailabilityHelp ?? state.value?.controlPlaneHighAvailabilityHelp
+    return state.value?.shootDefaults.controlPlaneHighAvailabilityHelp ?? state.value?.controlPlaneHighAvailabilityHelp
```
grolu:
```suggestion
    return shootDefaults.controlPlaneHighAvailabilityHelp ?? state.value?.controlPlaneHighAvailabilityHelp
```

Add a computed for shootDefaults. See also my comment. Adapt other computed properties accordingly

frontend/src/store/config.js
benedikt-haug: ah, now i get what you wanted to be changed. 

Incorporated your feedback, though my linter was persistent about using shootDefaults.value instead, please have a look.

grolu:
Hello @benedikt-haug 
I had a look at your changes, they look good except for a small error

`configStore` is missing in cloudprofile store. Please have a look.

Also, I noticed that you moved some settings like `defaultHibernationSchedule`, `controlPlaneHighAvailabilityHelp`, ... to the new `shootDefaults`. While this makes totally sense from my point of view, we should mention these changes as `breaking` in the release notes as operators need to update their configuration accordingly.

Moreover, I noticed that you also added defaults for `infrastructures` - did you have a look at this PR:  #2695 - it kind of conflicts with the new defaults setting. Please let me known if you still need the defaults for `infrastructures` after we merged the mentioned PR. Just to let you know. We currently do not support to hide certain infrastructures / image vendors / dns providers with PR #2695. Visible infrastructures and image vendors can be controlled using cloud profiles. For dns providers it is currently not possible to `hide` them in the dashboard. It is possible to extend but not to narrow the list. Is this a requirement from your side?

benedikt-haug (commented):
Hey @grolu,
thanks for having a look :)

I think I've successfully pulled off the rebase :D
Added the configStore to the cloudProfile.

You are correct about my changes being a breaking change. Is announcing this something I'd do in this merge request or something that is written somewhere else?

I personally don't need to change the infrastructure list. I just wanted to move all hardcoded defaults to configstore. I thought that maybe it would be fun to be able to change the sorting to prioritize specific infrastructures or something.

grolu:
Okay this looks good. I also like that you added configuration examples to the values.yaml.
I polished your release notes to provide some more details.
I will now do some final testing.

charts/gardener-dashboard/values.yaml
```diff
@@ -245,45 +245,59 @@ global:
...
-      # autoUpdateOS: true
-      # autoUpdateKubernetes: true
+        # hibernationSchedule:
+        #   evaluation:
+        #   - start: 00 17 * * 1,2,3,4,5
+        #   development:
+        #   - start: 00 17 * * 1,2,3,4,5
+        #     end: 00 08 * * 1,2,3,4,5
+        #   production: ~
+        # nodesCIDR: '10.250.0.0/16'
```
petersutter: uncomment here and comment out the deprecated values

frontend/src/components/NewShoot/GNewShootInfrastructureDetails.vue
```diff
@@ -270,21 +269,25 @@ export default {
...
         required: requiresInfrastructure('openstack'),
       }),
       loadBalancerClassNames: withFieldName('Load Balancer Class Names', {
         required: requiresInfrastructure('vsphere'),
-        includesKey: withMessage('Load Balancer Class \'default\' must be selected', includesIfAvailable('default', 'allLoadBalancerClassNames')),
+        includesKey: withMessage('Load Balancer Class \'default\' must be selected', selectedClassNames => {
+          const validationApplies = !this.workerless && this.providerType === 'vsphere'
+          const defaultClassIsAvailable = includes(this.allLoadBalancerClassNames, 'default')
+          return !validationApplies || !defaultClassIsAvailable || includes(selectedClassNames, 'default')
+        }),
```
petersutter: revert

petersutter (commented):
/lgtm

petersutter (commented):
/lgtm
/approve

PR #2478, opened 2025-06-05
Feature hide elements of NewShoot dialogue

benedikt-haug:
This PR is part of the Hackaton.
This PR allows the operator to hide elements of the NewShoot dialogue.

Without configuring anything, nothing should change. The end-result might be:
grafik

 /cc @marc1404 @klocke-io

**Release note**:

```feature operator
Allow hiding parts of the newShoot dialogue
```


charts/gardener-dashboard/values.yaml
```diff
@@ -159,6 +159,14 @@ global:
       - title: Issues
         icon: mdi-bug
         url: https://github.com/gardener/gardener/issues
+      # show: # hides new shoot creation items
```
petersutter: this is strange. I'd rather name the property `hide` and the default for each item is `false`.

petersutter (commented):
This PR makes sense together with PR https://github.com/gardener/dashboard/pull/2476, however, I do not get the complete story. Is the goal to avoid overwhelming users when creating shoots and to control certain defaults (some of which cannot be changed after shoot creation), while giving users the freedom to change other settings later on (like hibernation, DNS config, addons, auto update)?

If there are CloudProfiles from different providers, I can change the infrastructure in the YAML tab. But this makes it confusing. Also you can create credentials for the infrastructure provider but you cannot select / change the infrastructure provider on the creation dialog. Or is the idea/recommendation to hide the infrastructure option only in landscapes where there is no other provider type?

benedikt-haug:
> This PR makes sense together with PR #2476, however, I do not get the complete story. Is the goal to avoid overwhelming users when creating shoots and to control certain defaults (some of which cannot be changed after shoot creation), while giving users the freedom to change other settings later on (like hibernation, DNS config, addons, auto update)?
> If there are CloudProfiles from different providers, I can change the infrastructure in the YAML tab. But this makes it confusing. Also you can create credentials for the infrastructure provider but you cannot select / change the infrastructure provider on the creation dialog. Or is the idea/recommendation to hide the infrastructure option only in landscapes where there is no other provider type?

The goals of this PR is for the operator to be able to gain by configuration:
* free screen real estate if no choice is possible e.g. hide infrastructure section if only a single cloud provider exists or if choosing non-default isn't economically sensible
* nudge novice users e.g. if high-availability is the configured default (possible by the other PR), disabling this is generally not intended by the cluster operator, thus hiding this section reduces potential for user errors
* save explanation effort by hiding advanced usecases e.g. addons which aren't actually supported for production or workerless clusters which are mainly useful for development purposes

charts/gardener-dashboard/values.yaml
benedikt-haug: sorry for the slow reply :(
incorporated the feedback

PR #2489, opened 2025-06-13
Hide retry button as viewer

### LINKED ISSUE #2477: Retry button visible as viewer
**What happened**:
The retry button should not be visible to users with only viewer permission, as otherwise they run into a permission error when clicking on it.

**Environment**:

klocke-io:
Retry button was visible as viewer
Fixes #2477

**Release note**:

```bugfix user

```

### DIFF
--- a/frontend/src/components/GRetryOperation.vue
+++ b/frontend/src/components/GRetryOperation.vue
@@ -29,8 +29,10 @@ import {
   computed,
   inject,
 } from 'vue'
+import { storeToRefs } from 'pinia'
 
 import { useAppStore } from '@/store/app'
+import { useAuthzStore } from '@/store/authz.js'
 
 import { useShootItem } from '@/composables/useShootItem'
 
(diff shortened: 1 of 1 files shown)


grolu (approved):
Remove comment in if stmt
Otherwise /lgtm

frontend/src/components/GRetryOperation.vue
```diff
@@ -52,7 +54,14 @@ const appStore = useAppStore()
 
 const retryingOperation = ref(false)
 
+const authzStore = useAuthzStore()
+const {
+  canPatchShoots,
+} = storeToRefs(authzStore)
+
 const canRetry = computed(() => {
+  if (!canPatchShoots.value) return // No permission to patch shoots
```
petersutter: the comment is unnecessary, also we always add brackets (should have a linter for it). I also returned false to make it more explicit
```suggestion
  if (!canPatchShoots.value) {
    return false
  }
```

PR #2490, opened 2025-06-16
update autoscaler values on blur instead of input

### LINKED ISSUE #2468: "Autoscaler Min." of worker group wrongly changed to 1 when cluster owner typed 1 to increase "Autoscaler Max." to 10.
**What happened**:
"Autoscaler Min." of worker group wrongly changed to 1 when cluster owner typed 1 to increase "Autoscaler Max." to 10.

I think such UI auto behavior can be improved. And value check of "Autoscaler Min." and "Autoscaler Max." can be done after cluster owner confirmed adjustment completion in UI.

See video mentioned in 
https://github.tools.sap/kubernetes-live/issues-live/issues/7198

**Environment**:
Gardener 1.80.2

klocke-io:
Triggers the update of the min and max autoscaler only on blur. 
Fixes https://github.com/gardener/dashboard/issues/2468

**Release note**:

```bugfix user
update of dependent autoscaler value will only be triggered after the ui element is not focused anymore
```

### DIFF
--- a/frontend/src/components/ShootWorkers/GWorkerInputGeneric.vue
+++ b/frontend/src/components/ShootWorkers/GWorkerInputGeneric.vue
@@ -90,7 +90,7 @@ SPDX-License-Identifier: Apache-2.0
           label="Autoscaler Min."
           variant="underlined"
           @input="v$.worker.minimum.$touch()"
-          @blur="v$.worker.minimum.$touch()"
+          @blur="v$.worker.minimum.$touch(); overrideMax()"
         />
       </div>
       <div class="small-input">
@@ -103,7 +103,7 @@ SPDX-License-Identifier: Apache-2.0
           variant="underlined"
           :error-messages="getErrorMessages(v$.worker.maximum)"
           @input="v$.worker.maximum.$touch()"
-          @blur="v$.worker.maximum.$touch()"
+          @blur="v$.worker.maximum.$touch(); overrideMin()"
         />
       </div>
       <div class="small-input">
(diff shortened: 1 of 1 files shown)


grolu:
If you change to yaml tab and back to overview, now change the autoscaler values and switch to yaml tab again, the changes will get lost. This is not a new bug. I experienced the same behavior in master branch. Changes to other values seem to be preserved, but I did not check all of them.

frontend/src/components/ShootWorkers/GWorkerInputGeneric.vue
```diff
@@ -90,7 +90,7 @@ SPDX-License-Identifier: Apache-2.0
           label="Autoscaler Min."
           variant="underlined"
           @input="v$.worker.minimum.$touch()"
-          @blur="v$.worker.minimum.$touch()"
+          @blur="v$.worker.minimum.$touch(); overrideMax()"
```
grolu: please make a method for this. we usually do not assign multiple functions to an event

frontend/src/components/ShootWorkers/GWorkerInputGeneric.vue
```diff
@@ -103,7 +103,7 @@ SPDX-License-Identifier: Apache-2.0
           variant="underlined"
           :error-messages="getErrorMessages(v$.worker.maximum)"
           @input="v$.worker.maximum.$touch()"
-          @blur="v$.worker.maximum.$touch()"
+          @blur="v$.worker.maximum.$touch(); overrideMin()"
```
grolu: same here

klocke-io:
Fixed by moving the logic to dedicated methods `ensureValidAutoscalerMin()` and `ensureValidAutoscalerMax()`

frontend/src/components/ShootWorkers/GWorkerInputGeneric.vue
klocke-io: https://github.com/gardener/dashboard/pull/2490#issuecomment-3031419503

petersutter (approved):

PR #2501, opened 2025-07-03
Temporary disable flaky test

petersutter:
One test was temporarily disabled with this PR until we have migrated to the latest `pRetry` version and use an abort signal in the `afterEach` hook to ensure that retries are properly aborted.  
Currently, the test is skipped because the retry mechanism in `getConfiguration` repeatedly calls the `discovery` function on failure. Since retries are not aborted between tests, `discovery` may be called multiple times, which interferes with other tests that expect it to be called only once. This leads to test leakage, where ongoing retries from a failed test can affect subsequent tests and make the test suite unreliable.

Fixes #

**Release note**:

```bugfix user

```

### DIFF
--- a/backend/__tests__/acceptance/auth.spec.js
+++ b/backend/__tests__/acceptance/auth.spec.js
@@ -199,7 +199,8 @@ describe('auth', function () {
     expect(url.searchParams.get('state')).toEqual('state')
   })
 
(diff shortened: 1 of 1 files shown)


grolu (approved):
/lgtm

PR #2509, opened 2025-07-08
Refactor shoot issues watch

grolu:
## Summary
- stop tracking shoots with issues via Set in watches.shoots
- use `oldObject` from informer update events instead
- adjust unit tests

## Testing
- `yarn workspaces foreach --all --no-private run lint`
- `yarn workspaces foreach --all --no-private run test` *(fails: helm not found)*

------
https://chatgpt.com/codex/tasks/task_e_686d02442e008331a00be374d8116a87

### DIFF
--- a/backend/lib/watches/shoots.js
+++ b/backend/lib/watches/shoots.js
@@ -8,9 +8,8 @@
 
 const { shootHasIssue } = require('../utils')
 
-module.exports = (io, informer, options) => {
+module.exports = (io, informer) => {
   const nsp = io.of('/')
-  const { shootsWithIssues = new Set() } = options ?? {}
 
   const publishShoots = event => {
     const { type, object } = event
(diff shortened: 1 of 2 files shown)


backend/lib/watches/shoots.js
```diff
@@ -8,9 +8,8 @@
 
 const { shootHasIssue } = require('../utils')
 
-module.exports = (io, informer, options) => {
+module.exports = (io, informer) => {
```
petersutter, line 11: https://github.com/grolu/dashboard/blob/6819c39442a4ef009cd77c2ebbe00ac614210ade/backend/lib/hooks.js#L56

here a signal is passed as options, however I do not see how this was used in the previous code. We should double check why this is not used anymore

petersutter (approved):
/lgtm

PR #2517, opened 2025-07-17
Add workflow to prepare hotfix branch

petersutter:
This PR adds a workflow to prepare hotfix a branch. The workflow will create a PR for applying the changes to the newly created hotfix branch. The workflow accepts the tag, for which a hotfix branch should be created. 

Fixes #

**Release note**:

```bugfix user

```

### DIFF
--- a/.github/workflows/prepare-hotfix-branch.yaml
+++ b/.github/workflows/prepare-hotfix-branch.yaml
@@ -1,0 +1,84 @@
+name: Prepare Hotfix Branch
+
+on:
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: 'The tag for which the hotfix should be created'
+        required: true
+        type: string
+
+jobs:
+  prepare-hotfix:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      pull-requests: write
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          ref: ${{ inputs.tag }}
+      - name: Setup Git Identity
+        uses: gardener/cc-utils/.github/actions/setup-git-identity@master
+      - name: Compute and validate branch names
+        env:
+          TAG: ${{ inputs.tag }}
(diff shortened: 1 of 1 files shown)


.github/workflows/prepare-hotfix-branch.yaml
```diff
@@ -0,0 +1,84 @@
...
+      pull-requests: write
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          ref: ${{ inputs.tag }}
+      - name: Setup Git Identity
+        uses: gardener/cc-utils/.github/actions/setup-git-identity@master
+      - name: Compute and validate branch names
+        env:
+          TAG: ${{ inputs.tag }}
```
grolu: you could even move this env to step level and use it also in the checkout step

.github/workflows/prepare-hotfix-branch.yaml
```diff
@@ -0,0 +1,84 @@
...
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: 'The tag for which the hotfix should be created'
+        required: true
+        type: string
+
+jobs:
+  prepare-hotfix:
+    runs-on: ubuntu-latest
```
klocke-io, line 13: is there maybe a smaller and simpler image we could use like alpine?

klocke-io (approved):
/lgtm

.github/workflows/prepare-hotfix-branch.yaml
grolu: I guess we need tools like git to be available. Of course we could start from alpine and add only the tools we need but let's keep it simple

.github/workflows/prepare-hotfix-branch.yaml
petersutter: https://github.com/gardener/dashboard/pull/2517/commits/2f5a0caa9d7af9334673286fbb881c0ac29ecc87

PR #2526, opened 2025-07-24
Use GitHub App token in prepare-hotfix workflow

petersutter:
This PR updates the prepare-hotfix workflow to use a token from the Gardener GitHub Actions App instead of the default GITHUB_TOKEN. This is necessary because the workflow needs to push to protected branches, and the app provides the required permissions to bypass rulesets as an exception.

This aligns with the migration to GitHub Actions and the use of rulesets for branch protection, as documented here: https://github.com/gardener/cc-utils/blob/master/doc/github_actions.rst#branch-protection---rulesets.

Fixes #

**Release note**:

```bugfix user

```


grolu (approved):
/lgtm

PR #2572, opened 2025-08-12
Add icon picker for custom fields

petersutter:
This PR adds an icon picker for custom fields.
Generate `virtual:g-mdi-meta` module with icon names and aliases at build time

Screenshot 2025-08-12 at 17 53 22

Fixes #

**Release note**:

```feature user
Added icon picker for custom fields
```

### DIFF
--- a/frontend/src/components/GIconPicker.vue
+++ b/frontend/src/components/GIconPicker.vue
@@ -10,13 +10,13 @@ SPDX-License-Identifier: Apache-2.0
     :items="items"
     :label="label"
     variant="underlined"
-    clearable
+    :clearable="!!model"
     :error-messages="errorMessages"
     :hint="hint"
     persistent-hint
     :custom-filter="filter"
   >
-    <template #append>
+    <template #prepend-inner>
       <v-icon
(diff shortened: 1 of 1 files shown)


grolu:
When I mouseover the icon input field while it is initial (empty) a clear icon appears (not desired). After input and deletion of the input the clear icon disappears (desired). Something is wrong with that icon and the initial state

grolu:
Is it possible to show the selected icon in the `prepend` slot? Right now it is shown after the select input. But in the dropdown it is set as prepend-item or something. Wouldn't it be more consistent to also show it like that in the selection?

.dockerignore
```diff
@@ -60,6 +60,7 @@
 !frontend/index.html
 !frontend/public
 !frontend/src
+!frontend/vite
```
klocke-io: Are we sure that we always want to include the hole `vite` folder? 


```suggestion
!frontend/vite/mdilcons.js
```

petersutter: good point

grolu (approved):
/lgtm

PR #2588, opened 2025-08-25
Enforced WebSocket-only Socket.IO with an allowRequest configuration

grolu:
- Enforced WebSocket-only Socket.IO with an allowRequest configuration
- Added io.allowedOrigins configuration support and wired it into Helm charts and documentation
- Extended backend tests to cover allowed, disallowed, and default-origin scenarios

Fixes #

**Release note**:

```breaking operator
The backend now requires an `io.AllowedOrigin` configuration to protect the Socket.IO server from cross-site requests

* To allow all origins, you may set the value to `*` (not recommended)
* When deploying with the **Gardener operator**, this setting is automatically configured based on the public ingress hosts
* For manual deployments, you must provide the correct values in the Helm chart configuration
```


backend/lib/io/index.js
```diff
@@ -9,11 +9,40 @@ import logger from '../logger/index.js'
...
+import config from '../config/index.js'
 
 function init (httpServer, cache) {
+  const allowedOrigins = config.websocketAllowedOrigins
+  if (!allowedOrigins?.length) {
+    logger.warn('websocketAllowedOrigins configuration is required')
+  }
+  allowedOrigins.forEach(allowedOrigin => {
+    logger.info('WebSocket allowed origin: %s', allowedOrigin)
+  })
```
petersutter:
```suggestion
  if (!allowedOrigins?.length) {
    logger.warn('websocketAllowedOrigins configuration is required')
  } else {
    logger.info('WebSocket allowed origins: %s', allowedOrigins.join(', '))
  }
```

backend/lib/io/index.js
```diff
@@ -9,11 +9,40 @@ import logger from '../logger/index.js'
...
+
+  const allowAll = allowedOrigins.includes('*')
+  if (allowAll) {
+    logger.warn('WebSocket allowing all origins (*) - this seeting is not recommended for production environments')
```
petersutter: in case we want to have different messages depending on the isProd config. Feel free to take over or discard this suggestion

```suggestion
    const msg = config.isProd
      ? 'WebSocket allowing all origins (*) — this is unsafe in production. Restrict allowedOrigins.'
      : 'WebSocket allowing all origins (*) — OK for local/dev, but do not use in production.'
    logger.warn(msg)
```

charts/__tests__/gardener-dashboard/runtime/dashboard/configmap.spec.js
```diff
@@ -889,5 +889,20 @@ describe('gardener-dashboard', function () {
         expect(pick(config, ['frontend.experimental'])).toMatchSnapshot()
       })
     })
+
+    describe('socket io', function () {
+      it('should render the template with allowed origins', async function () {
```
petersutter: add test for default * case

charts/gardener-dashboard/values.yaml
```diff
@@ -84,6 +84,12 @@ global:
     # Example: nodeOptions: [--optimize-for-size, --max-old-space-size=920, --gc-interval=100]
     nodeOptions: [--max-old-space-size=920]
 
+    io:
+      # list of allowed origins for the socket.io events endpoint
+      # use ["*"] to allow all origins
```
petersutter: mention that * is not recommended for production use cases

backend/lib/io/index.js
```diff
@@ -9,11 +9,40 @@ import logger from '../logger/index.js'
 import helper from './helper.js'
 import dispatcher from './dispatcher.js'
 import _ from 'lodash-es'
+import config from '../config/index.js'
 
 function init (httpServer, cache) {
+  const allowedOrigins = config.websocketAllowedOrigins
+  if (!allowedOrigins?.length) {
+    logger.warn('websocketAllowedOrigins configuration is required')
+  }
```
petersutter, line 20:
```suggestion
```

rm, we don't reach this code as websocketAllowedOrigins is a required property

backend/__tests__/config.spec.cjs
```diff
@@ -167,14 +122,44 @@ describe('config', function () {
...
         expect(config.oidc.issuer).toBe('https://kubernetes:32001')
         expect(config.oidc.client_id).toBe('client_id_from_file')
         expect(config.oidc.client_secret).toBe('client_secret_from_file')
         expect(config.gitHub.authentication.appId).toBe(12345)
         expect(config.gitHub.authentication.installationId).toBe(67890)
+        expect(config.websocketAllowedOrigins).toEqual(['*'])
+        expect(config.logLevel).toBe('info')
+      })
+
+      it('should return the config overridden by environment variables', function () {
```
petersutter, line 137: add test for empty websocketAllowedOrigins

grolu: I upgraded the array env parsing logic and added a bunch of tests that ensure required config early exits

backend/lib/io/index.js
grolu: You can specify an empty array via configuration. The env paring logic discards empty arrays but it is still possible via config yaml like this

`websocketAllowedOrigins: []`

Instead of adding a logger warn here I suggest to throw an error in case of empty array. This has basically the same effect as the error is catched upstream and logged as warning but it makes the code cleaner in my pov.

backend/lib/io/index.js
grolu: lgtm

backend/__tests__/config.spec.cjs
```diff
@@ -85,96 +86,129 @@ describe('config', function () {
...
 
-        const filename = '/etc/gardener/3/config.yaml'
-        const config = gardener.loadConfig(filename, { env })
-        expect(gardener.readConfig).toHaveBeenCalledTimes(1)
-        expect(gardener.readConfig.mock.calls[0]).toEqual([filename])
+        // global env
+        expect(config.apiServerUrl).toBe(env.API_SERVER_URL)
         expect(config.sessionSecret).toBe(env.SESSION_SECRET)
-        expect(config.logLevel).toBe('debug')
+        expect(config.websocketAllowedOrigins).toEqual(env.WEBSOCKET_ALLOWED_ORIGINS.split(','))
```
petersutter: write it out what the expected value is

backend/__tests__/config.spec.cjs
petersutter: I'm still missing a test where config.websocketAllowedOrigins = [] or is not set, both of which should result in an error.

backend/lib/config/gardener.js
```diff
@@ -115,15 +115,29 @@ const configMappings = [
...
+    configPath: 'websocketAllowedOrigins',
+    type: 'Array',
+  },
 ]
 
 function parseConfigValue (value, type) {
+  const parseArray = value => {
+    if (typeof value !== 'string' || !value.length > 0) {
+      return undefined
+    }
```
petersutter: did you mean to write this odd `!value.length > 0`?

I think it's better to be more explicit
```suggestion
    if (value == null || typeof value !== 'string' || value.length === 0) {
      return undefined;
    }
```

backend/lib/io/index.js
petersutter: an empty array should should not be allowed in the config, at least for websocketAllowedOrigins

backend/lib/io/index.js
```diff
@@ -14,16 +14,17 @@ import config from '../config/index.js'
...
+    logger.info('WebSocket allowed origins: %s', allowedOrigins.join(', '))
   }
-  allowedOrigins.forEach(allowedOrigin => {
-    logger.info('WebSocket allowed origin: %s', allowedOrigin)
-  })
+
+  const msg = config.isProd
+    ? 'WebSocket allowing all origins (*) — this is unsafe in production. Restrict allowedOrigins.'
+    : 'WebSocket allowing all origins (*) — OK for local/dev, but do not use in production.'
+  logger.warn(msg)
```
petersutter: this should be inside 
```js
if (allowAll) {
  // log warning here
}
```

charts/__tests__/gardener-dashboard/runtime/dashboard/configmap.spec.js
```diff
@@ -903,6 +903,19 @@ describe('gardener-dashboard', function () {
         const config = yaml.load(configMap.data['config.yaml'])
         expect(pick(config, ['io'])).toMatchSnapshot()
       })
+
+      it('should render the template with default allowed origin', async function () {
+        const values = {
+          global: {
+            websocketAllowedOrigins: ['*'],
```
petersutter: unset `websocketAllowedOrigins` so that it defaults to * because of https://github.com/gardener/dashboard/blob/cca93ef7ace9305bc28f260c1730061a72796871/charts/gardener-dashboard/charts/runtime/templates/dashboard/configmap.yaml#L40

I want to test the case ` | default (list "*")`

backend/__tests__/config.spec.cjs
grolu: I removed the env variables completely from the test.
The empty array case is tested in `io.cors.spec.cjs` *should reject connections if no origins are configure*

charts/__tests__/gardener-dashboard/runtime/dashboard/configmap.spec.js
grolu: Changed the test to unset the values data. Now the helm defaulting is tested.
Also fixed the keys and helm structure for websocketAllowedOrigins

Issue #2497, opened 2025-06-26
Dashboard does not allow to create shoots on tainted seeds, even with project toleration


chrisbal-b:
**What happened**:
Gardener dashboard prevents:

- adding Infrastructure credentials
- creation of a shoot

in case of a tainted seed (see screenshot)
![Image](https://github.com/user-attachments/assets/68e8ab35-c6a3-468f-93fe-6ef5c3340f56)

If the Gardener project has a default configuration added:
``` yaml     
tolerations:
    defaults:
    -key: seed.gardener.cloud/protected
    whitelist:
    - key: seed.gardener.cloud/protected
 ``` 
it is possible to apply shoot-manifests (without toleration) to the virtual garden and have them deployed.

**What you expected to happen**:
Gardener Dashboard should consider project-wide tolerations, when checking, if proper Seed and CloudProfile is available

**How to reproduce it (as minimally and precisely as possible)**:
Gardener setup with one Seed that has taint:
``` yaml       
taints:
  - key: seed.gardener.cloud/protected
 ```
Project, with tolerations defined. Let me know, if you need an environment to observe this issue.

**Anything else we need to know**:
[This PR](https://github.com/gardener/dashboard/pull/2454) lowers the impact of the issue, as it seems to only show a warning for non-matching seeds and does not block the creation of shoots anymore

**Environment**:
GCP hosted Gardener environment

PR #2589, opened 2025-08-25
Considers project tolerations for cloudprofile selection in shoot creation dialogue

klocke-io:
Feature: Considers project tolerations 
Refactoring: moves seed matching for cloud profiles to frontend
Fixes https://github.com/gardener/dashboard/issues/2497

**Release note**:

```feature user
Considers project tolerations for cloudprofile selection in shoot creation dialogue
```

### DIFF
--- a/backend/lib/cache/index.js
+++ b/backend/lib/cache/index.js
@@ -86,15 +86,6 @@ export default {
   getSeedByUid (uid) {
     return cache.get('seeds').find(['metadata.uid', uid])
   },
-  getVisibleAndNotProtectedSeeds () {
-    const predicate = item => {
-      const taints = _.get(item, ['spec', 'taints'])
-      const unprotected = !_.find(taints, ['key', 'seed.gardener.cloud/protected'])
-      const visible = _.get(item, ['spec', 'settings', 'scheduling', 'visible'])
-      return unprotected && visible
-    }
-    return _.filter(cache.getSeeds(), predicate)
-  },
   getProject (name) {
     const project = cache.get('projects').find(['metadata.name', name])
     if (!project) {
--- a/frontend/__tests__/stores/cloudProfile.spec.js
+++ b/frontend/__tests__/stores/cloudProfile.spec.js
@@ -26,7 +26,7 @@ describe('stores', () => {
 
     let cloudProfileRef
 
-    function setData (spec) {
+    function setSpec (spec) {
       cloudProfileStore.setCloudProfiles([{
         metadata: {
           name: 'foo',
@@ -36,13 +36,13 @@ describe('stores', () => {
     }
 
     function setMachineImages (machineImages) {
-      setData({
+      setSpec({
         machineImages,
       })
     }
 
     function setKubernetesVersions (kubernetesVersions) {
-      setData({
+      setSpec({
         kubernetes: {
           versions: kubernetesVersions,
         },
@@ -651,7 +651,7 @@ describe('stores', () => {
       ]
 
       beforeEach(() => {
-        setData({
+        setSpec({
           machineTypes,
           volumeTypes,
           regions,
@@ -749,7 +749,7 @@ describe('stores', () => {
       ]
 
       beforeEach(() => {
-        setData({
+        setSpec({
           providerConfig: {
             constraints: {
               floatingPools,
@@ -833,7 +833,7 @@ describe('stores', () => {
       })
 
       it('should return default node cidr from cloud profile', () => {
-        setData({
+        setSpec({
           providerConfig: {
             defaultNodesCIDR: '1.2.3.4/16',
           },
--- a/frontend/__tests__/stores/cloudProfile.spec.js
+++ b/frontend/__tests__/stores/cloudProfile.spec.js
@@ -26,12 +26,12 @@ describe('stores', () => {
 
     let cloudProfileRef
 
-    function setData (data) {
+    function setData (spec) {
       cloudProfileStore.setCloudProfiles([{
         metadata: {
           name: 'foo',
         },
-        data,
+        spec,
       }])
     }
 
--- a/frontend/__tests__/stores/seed.spec.js
+++ b/frontend/__tests__/stores/seed.spec.js
@@ -29,6 +29,12 @@ describe('stores', () => {
             type: 'aws',
           },
           region: 'eu-west-1',
+          taints: [], // No taints = not protected
+          settings: {
+            scheduling: {
+              visible: true,
+            },
+          },
         },
       },
       {
@@ -44,6 +50,12 @@ describe('stores', () => {
             type: 'aws',
           },
           region: 'us-east-1',
+          taints: [], // No taints = not protected
+          settings: {
+            scheduling: {
+              visible: true,
+            },
+          },
         },
       },
       {
@@ -59,6 +71,12 @@ describe('stores', () => {
             type: 'gcp',
           },
           region: 'europe-west1',
+          taints: [], // No taints = not protected
+          settings: {
+            scheduling: {
+              visible: true,
+            },
+          },
         },
       },
       {
@@ -74,6 +92,12 @@ describe('stores', () => {
             type: 'azure',
           },
           region: 'westeurope',
+          taints: [], // No taints = not protected
+          settings: {
+            scheduling: {
+              visible: true,
+            },
+          },
         },
       },
     ]
@@ -92,7 +116,7 @@ describe('stores', () => {
             name: 'aws-profile',
             providerType: 'aws',
           },
-          data: {
+          spec: {
             type: 'aws',
           },
         }
@@ -109,7 +133,7 @@ describe('stores', () => {
             name: 'aws-production-profile',
             providerType: 'aws',
           },
-          data: {
+          spec: {
             type: 'aws',
             seedSelector: {
               matchLabels: {
@@ -130,7 +154,7 @@ describe('stores', () => {
             name: 'multi-provider-profile',
             providerType: 'aws',
           },
-          data: {
+          spec: {
             type: 'aws',
             seedSelector: {
               providerTypes: ['aws', 'gcp'],
@@ -151,7 +175,7 @@ describe('stores', () => {
             name: 'wildcard-profile',
             providerType: 'aws',
           },
-          data: {
+          spec: {
             type: 'aws',
             seedSelector: {
               providerTypes: ['*'],
@@ -169,7 +193,7 @@ describe('stores', () => {
             name: 'complex-selector-profile',
             providerType: 'aws',
           },
-          data: {
+          spec: {
             type: 'aws',
             seedSelector: {
               providerTypes: ['*'],
@@ -193,7 +217,7 @@ describe('stores', () => {
             name: 'no-match-profile',
             providerType: 'openstack',
           },
-          data: {
+          spec: {
             type: 'openstack',
           },
         }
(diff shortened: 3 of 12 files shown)


backend/lib/cache/index.js
```diff
@@ -86,15 +86,6 @@ export default {
   getSeedByUid (uid) {
     return cache.get('seeds').find(['metadata.uid', uid])
   },
-  getVisibleAndNotProtectedSeeds () {
-    const predicate = item => {
-      const taints = _.get(item, ['spec', 'taints'])
-      const unprotected = !_.find(taints, ['key', 'seed.gardener.cloud/protected'])
-      const visible = _.get(item, ['spec', 'settings', 'scheduling', 'visible'])
-      return unprotected && visible
```
petersutter, line 94: did this filter get lost?

frontend/__tests__/stores/cloudProfile.spec.js
```diff
@@ -26,12 +26,12 @@ describe('stores', () => {
 
     let cloudProfileRef
 
-    function setData (data) {
+    function setData (spec) {
```
petersutter: should probably also be renamed to setSpec

frontend/__tests__/stores/seed.spec.js
```diff
@@ -0,0 +1,304 @@
...
+    })
+
+    describe('#seedsForCloudProfile', () => {
+      it('should return seeds matching cloud profile provider type', () => {
+        const cloudProfile = {
+          metadata: {
+            name: 'aws-profile',
+            providerType: 'aws',
+          },
+          data: {
```
petersutter: the cloud profile structure should not have a `data` field but a `spec` instead.

frontend/__tests__/stores/cloudProfile.spec.js
klocke-io: addressed in https://github.com/gardener/dashboard/pull/2589/commits/355e3bc0990544c9bc1aa2ad303247a6039c2e5e

frontend/__tests__/stores/seed.spec.js
klocke-io: addressed in cd1a74d6bafcd29c71da145e74bb3c3c5fb41541

backend/lib/cache/index.js
klocke-io: addressed in cd1a74d6bafcd29c71da145e74bb3c3c5fb41541

frontend/src/components/GSelectCloudProfile.vue
```diff
@@ -39,49 +45,49 @@ const props = defineProps({
...
-    required: true,
-  },
-  namespacedCloudProfiles: {
-    type: Array,
-    required: false,
-    default: () => [],
-  },
 })
 
+const { seedsByCloudProfileRef, cloudProfileList } = useCloudProfileStore()
```
petersutter: use storeToRefs for cloudProfileList (you can check in other places in the codebase for reference), otherwise you loose reactivity

frontend/src/utils/index.js
```diff
@@ -149,6 +149,15 @@ export function displayName (username) {
   return username
 }
```
grolu, line 151: it is currently not used. It is also not included in the example cloudprofile in g/g repo. I quickly checked dev landscape and found only one cp using it. Also it includes the old sapcloud.io domain. So let's drop it for now. We could re-introduce it with new label if required.

frontend/src/components/GSelectCloudProfile.vue
```diff
@@ -39,49 +45,49 @@ const props = defineProps({
...
+const projectStore = useProjectStore()
+
+const seeds = computed(() => {
+  return seedsByCloudProfileRef(props.modelValue, projectStore.project)
```
petersutter: assign props.modelValue to variable for better readability

frontend/src/utils/index.js
petersutter: not sure if anyone else is using it in the community. Did we announce this as feature in the past?
We could also introduce a new gardener.cloud annotation and use the sapcloud.io as fallback for a few releases. In case we don't drop it, this should rather be moved to a cloudprofile composable

frontend/src/composables/useShootHelper.js
```diff
@@ -37,6 +38,7 @@ const shootPropertyMappings = Object.freeze({
   kubernetesVersion: ['spec', 'kubernetes', 'version'],
   providerType: ['spec', 'provider', 'type'],
   addons: ['spec', 'addons'],
+  projectName: ['metadata', 'namespace'],
```
petersutter: the namespace is not equivalent to the projectName. 
You can use `projectNameByNamespace` on the project store, to get a project name for a given namespace
Or you fetch the project by namespace, similar to how it is done here

https://github.com/gardener/dashboard/blob/372f302f5f21f0bb1ea26b2204b35b0ae0fa83ef/frontend/src/store/project.js#L88-L90

frontend/src/store/seed.js
```diff
@@ -44,17 +46,78 @@ export const useSeedStore = defineStore('seed', () => {
...
+    const seedTaints = get(seed, ['spec', 'taints'])
+
+    if (!seedTaints || seedTaints.length === 0) {
+      return true
+    }
+
+    const projectTolerations = get(project, ['spec', 'tolerations', 'defaults'], [])
+    if (!projectTolerations || projectTolerations.length === 0) {
+      return false
+    }
```
petersutter: we could also do the following, but no strong opinion. however I'd at least have the newline removed in line 51. Also there is inconsistent defaulting in the get function call. was this by intention? 
```suggestion
    const seedTaints = get(seed, ['spec', 'taints'])  ?? []
    if (seedTaints.length === 0) {
      return true
    }

    const projectTolerations = get(project, ['spec', 'tolerations', 'defaults']) ?? []
    if (projectTolerations.length === 0) {
      return false
    }
```

frontend/src/store/seed.js
```diff
@@ -44,17 +46,78 @@ export const useSeedStore = defineStore('seed', () => {
...
+      }
+
+      if (matchLabels && !isEmpty(matchLabels)) {
+        const seedLabels = get(seed, ['metadata', 'labels'], {})
+        const labelMatcher = matches(matchLabels)
+        return labelMatcher(seedLabels)
+      }
+
+      return true
+    }
```
petersutter: nit:  build label matcher once, also `matchLabels` check can be omitted when using isEmpty (line 98)

```suggestion
    const providerTypes = get(cloudProfile, ['spec', 'seedSelector', 'providerTypes'], [providerType])

    const labelMatcher = !isEmpty(matchLabels)
      ? matches(matchLabels)
      : null

    return function matchSeed (seed) {
      const seedProviderType = get(seed, ['spec', 'provider', 'type'])
      const providerTypeMatches = providerTypes.some(type => [seedProviderType, '*'].includes(type))

      if (!providerTypeMatches) {
        return false
      }

      if (matchLabels && !isEmpty(matchLabels)) {
        const seedLabels = get(seed, ['metadata', 'labels'], {})
        return labelMatcher(seedLabels)
      }

      return true
    }
```

frontend/src/store/seed.js
```diff
@@ -44,17 +46,78 @@ export const useSeedStore = defineStore('seed', () => {
...
+  }
+
+  function seedsForCloudProfileByProject (cloudProfile, project) {
+    if (!cloudProfile || !project) {
+      return []
+    }
```
petersutter: why would we return an empty list if project is not set?

frontend/src/store/seed.js
```diff
@@ -44,17 +46,78 @@ export const useSeedStore = defineStore('seed', () => {
...
+
+    const seeds = getVisibleAndToleratedSeeds(project)
+
+    if (!seeds.length > 0) {
+      return []
+    }
```
petersutter:
```suggestion
    const seeds = getVisibleAndToleratedSeeds(project)
    if (seeds.length === 0) {
      return []
    }
```

frontend/src/store/seed.js
```diff
@@ -44,17 +46,78 @@ export const useSeedStore = defineStore('seed', () => {
...
+      return () => false
+    }
+
+    const providerType = get(cloudProfile, ['spec', 'type'])
+    const matchLabels = get(cloudProfile, ['spec', 'seedSelector', 'matchLabels'])
+    const providerTypes = get(cloudProfile, ['spec', 'seedSelector', 'providerTypes'], [providerType])
+
+    return function matchSeed (seed) {
+      const seedProviderType = get(seed, ['spec', 'provider', 'type'])
+      const providerTypeMatches = providerTypes.some(type => [seedProviderType, '*'].includes(type))
```
petersutter: I know this code was not new (from backend code) but still it could be improved
```suggestion
      const providerTypeMatches = providerTypes.includes(seedProviderType) || providerTypes.includes('*')
```

frontend/src/composables/useShootHelper.js
klocke-io: @petersutter I would fetch the current `projectName` directly from the store instead of adding it to the shoot. like so: 

```js
diff --git a/frontend/src/composables/useShootHelper.js b/frontend/src/composables/useShootHelper.js
index 8231ccd6..b909f24c 100644
--- a/frontend/src/composables/useShootHelper.js
+++ b/frontend/src/composables/useShootHelper.js
@@ -9,6 +9,7 @@ import {
   inject,
   provide,
 } from 'vue'
+import { storeToRefs } from 'pinia'
 
 import { useCloudProfileStore } from '@/store/cloudProfile'
 import { useConfigStore } from '@/store/config'
@@ -38,7 +39,6 @@ const shootPropertyMappings = Object.freeze({
   kubernetesVersion: ['spec', 'kubernetes', 'version'],
   providerType: ['spec', 'provider', 'type'],
   addons: ['spec', 'addons'],
-  projectName: ['metadata', 'namespace'],
 })
 
 export function createShootHelperComposable (shootItem, options = {}) {
@@ -60,11 +60,12 @@ export function createShootHelperComposable (shootItem, options = {}) {
     kubernetesVersion,
     providerType,
     addons,
-    projectName,
   } = mapValues(shootPropertyMappings, path => {
     return computed(() => get(shootItem.value, path))
   })
 
+  const { projectName } = storeToRefs(projectStore)
+
   const project = computed(() => {
     return find(projectStore.projectList, ['metadata.name', projectName.value])
   })

```

frontend/src/store/seed.js
klocke-io: I see your point, is there a project standard in how we handle defaults? 
But I will definitely remove the newline.

frontend/src/store/seed.js
klocke-io: well I renamed the function accordingly :D, but jokes aside, I see your point, I will look again at this function, if it should also be consumable outside of a project context, and then maybe return just normal cloudprofiles.

frontend/src/utils/index.js
klocke-io: I created a dedicated issue which I would address afterwards. 

https://github.com/gardener/dashboard/issues/2611

frontend/src/components/GSelectCloudProfile.vue
```diff
@@ -47,11 +48,15 @@ const props = defineProps({
   },
 })
 
-const { seedsByCloudProfileRef, cloudProfileList } = useCloudProfileStore()
-const projectStore = useProjectStore()
+const cloudProfileRef = props.modelValue
```
petersutter: toRef

petersutter (approved):
/lgtm

PR #2610, opened 2025-08-29
Add support for `STACKIT` infrastructure

maboehm:
Add STACKIT to the list of supported providers. 

STACKIT used to give customers access to the Openstack API, but that will be replaced by its own [infrastructure API](https://docs.stackit.cloud/stackit/en/release-notes-23101442.html#ReleaseNotes-29072025InfrastructureAPIDeprecationNotice) . We are also in the process of writing and later open-sourcing the required `cloud-provider-stackit` and `gardener-extension-provider-stackit`.

Once the providers are ready (and the API stable), we will add full support in the dashboard.

**Release note**:

```feature user
Add support for STACKIT infrastructure
```

### DIFF
--- a/frontend/src/assets/stackit.svg
+++ b/frontend/src/assets/stackit.svg
@@ -1,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg id="Ebene_1" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 536.44 395.08">
+  <defs>
+    <style>
+      .cls-1 {
+        fill: #0a1e2d;
+      }
+    </style>
+  </defs>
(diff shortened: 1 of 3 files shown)


grolu:
Hi @maboehm 
Thank you for your contribution. Your PR looks good overall. However, the icon is not very visible in dark mode. I am aware that the AWS icon also has limited visibility in this mode, but you might consider adding a subtle border or similar adjustment to improve contrast. If this is not possible or not aligned with your guidelines, it is fine to leave it as it is. Unfortunately, providing a dedicated dark mode icon is currently not possible.

maboehm:
@grolu I'd prefer if we can just get this merged now. In the context of the wider open-sourcing of provider-stackit I will try to get a better logo for the dashboard.

Issue #1707, opened 2024-02-07
Add ability to brand infrastructures / cloud profiles


dhague:
**What would you like to be added**:
Gardener Dashboard currently has a built-in set of vendors, names and logos that cannot be customized. To help users identify the correct infrastructure and for emphasize (internal) brand names, it would be good to be able to configure the appearance in the Gardener Dashboard.

**Example use case**:
Instead of using an OpenStack logo for SAP Converged Cloud clusters, please use this logo:
![CCloud_Logo_dark](https://github.com/gardener/dashboard/assets/562527/3ab7c2b5-8a43-41c6-ab1a-6c003eb0e690)

Users may be confused when using SAP Converged Cloud to see this labeled as "OpenStack".
For SAP-internal demos this is especially confusing.

grolu:
Thank you @dhague for raising this issue. As this is an open source project, I slightly adjusted your issue to make it more generic.

We have already implemented several branding options making it possible to configure theme colors, logos, brand name and event the login page. Regarding vendors (cloud providers, machine image providers, etc.) we currently lack the possibility to configure them. However, I recently opened a PR (still in draft) that will make it possible to overwrite vendor names and icons via Dashboard configuration, see here: https://github.com/gardener/dashboard/pull/1553

However, this only solves part of your problem. When the PR mentioned above has been merged, it will be possible to rebrand `OpenStack` to `Converged Cloud`. But if some other `OpenStack` installation needs to be added to the landscape in the future, this ist the wrong abstraction layer. It would be better to differentiate on a cloud profile level.

We could also add the option to configure icons for individual cloud profiles. Also, clustering cloud profiles would be an option, so that you have an additional selection layer after selecting `OpenStack` with the `Converged Cloud` icon where all `converged cloud` profiles are grouped together. This would also help the user understand why the provider type in the shoot spec is still `openstack`.

PR #2695, opened 2025-12-01
feat(frontend) Support vendor branding

grolu:
- Moved vendor assets from src/assets to public/static/assets to make them replaceable with custom assets
- GVendorIcon component serves icons from this directory using filenames from vendor information
- Centralized dictionary for vendors (infra/dns/image) in config store
- Added defaults for all known vendors (displayname, icon, weight)
- All values can be extended or overridden via configuration
- Removed need to provide createTitle and updateTile in GSecretDialog (now kebap-case generated from displayName)
- Use vendorDisplayName everywhere to ensure we show configured name for vendors instead of technical name

Fixes https://github.com/gardener/dashboard/issues/1707 Fixes https://github.com/gardener/dashboard/issues/207 (partly)

**Release note**:

```feature operator
Vendor Branding
- Define order, name and icon of cloud providers in dashboard configuration
- Define order, name and icon of machine image vendors in dashboard configuration
- Extend built-in providers by providing additional configuration
```

### DIFF
--- a/docs/operations/customization.md
+++ b/docs/operations/customization.md
@@ -26,7 +26,10 @@ It is possible to change the branding of the Gardener Dashboard when using the [
 
 ## Vendor Branding
 
-It is possible to customize the appearance of vendors shown in the dashboard. You can override the display name, icon, sort order and visibility of *cloud provider vendors*, *DNS providers*, and *machine image vendors*.
+Here’s a polished and clearer version of your sentence:
+
+You can customize the appearance of built-in vendors displayed in the dashboard. This includes overriding the display name, icon, sorting, and order of *cloud provider vendors* and *machine image vendors*. In addition, you can configure custom vendors.
+
 
 Vendor customization is configured in the arrays
 `frontendConfig.branding.infraVendors`, `frontendConfig.branding.dnsVendors`, and `frontendConfig.branding.machineImageVendors`.
@@ -38,9 +41,6 @@ Each entry must be an object with a `name` key that identifies the vendor. The f
 | `displayName` | Name displayed in the dashboard |
 | `weight`      | Sorting weight. Lower values appear first. See default weights in `frontend/src/store/config.js` |
 | `icon`        | File name of the icon located in the `public/static/assets` folder. See [Logos and Icons](#logos-and-icons) for instructions on replacing assets |
-| `hidden`        | Hide vendor in the dashboard |
-
-With this configuration, you can extend the list of built-in vendors. For cloud providers, however, only the default secret dialogs are available, and custom shoot properties are not supported.
 
 ## Colors
 Gardener Dashboard has been built with Vuetify. We use Vuetify's built-in [theming support](https://vuetifyjs.com/en/features/theme/) to centrally configure colors that are used throughout the web application.
--- a/frontend/src/store/config.js
+++ b/frontend/src/store/config.js
@@ -25,7 +25,6 @@ import isEmpty from 'lodash/isEmpty'
 import camelCase from 'lodash/camelCase'
 import find from 'lodash/find'
 import uniq from 'lodash/uniq'
-import filter from 'lodash/filter'
 import sortBy from 'lodash/sortBy'
 import head from 'lodash/head'
 
@@ -703,8 +702,7 @@ export const useConfigStore = defineStore('config', () => {
 
   const sortedDnsProviderTypeList = computed(() => {
     const dnsProviderVendors = map(dnsProviderTypesList.value, vendorDetails)
-    const visibleDnsVendors = filter(dnsProviderVendors, ({ hidden }) => !hidden)
-    const sortedVisibleDnsVendors = sortBy(visibleDnsVendors, 'weight')
+    const sortedVisibleDnsVendors = sortBy(dnsProviderVendors, 'weight')
     return map(sortedVisibleDnsVendors, 'name')
   })
 
--- a/frontend/src/composables/useCloudProfile/useMachineImages.js
+++ b/frontend/src/composables/useCloudProfile/useMachineImages.js
@@ -68,8 +68,7 @@ export function useMachineImages (cloudProfile) {
         vendor,
       }
     })
-    const visibleMachineImagesWithVendors = filter(machineImagesWithVendors, ({ vendor }) => !vendor.hidden)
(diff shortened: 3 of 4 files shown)


docs/operations/customization.md
```diff
@@ -24,6 +24,21 @@ It is possible to change the branding of the Gardener Dashboard when using the [
...
+`frontendConfig.infraVendors`, `frontendConfig.dnsVendors`, and `frontendConfig.imageVendors`.
+Each entry must be an object with the following fields:
+
+| name          | usage|
+| ------------- |------|
+| `name`        | Unique vendor identifier (required) |
+| `displayName` | Name displayed in the dashboard |
+| `weight`      | Sorting weight. Lower values appear first. See default weights in `frontend/src/store/config.js` |
+| `icon`        | ile name of the icon located in the `public/static/assets` folder. See [Logos and Icons](#logos-and-icons) for instructions on replacing assets |
+
```
klocke-io, line 44: showing an example config file would be nice

frontend/src/assets/logo.svg
```diff
@@ -1,40 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
petersutter, line 1: was this file dropped on purpose

frontend/src/store/config.js
```diff
@@ -390,6 +426,235 @@ export const useConfigStore = defineStore('config', () => {
...
+    },
+  ]
+
+  const knownVendors = [
+    ...knownInfraVendors,
+    ...knownDNSVendors,
+    ...knownImageVendors,
+  ]
+
+  const vendorDetails = function (kind) {
```
petersutter:
```suggestion
  function vendorDetails (kind) {
```

frontend/src/store/config.js
```diff
@@ -390,6 +426,235 @@ export const useConfigStore = defineStore('config', () => {
...
+
+    return {
+      name: kind,
+      weight: Number.MAX_SAFE_INTEGER,
+      ...knownVendor,
+      ...configuredVendor,
+    }
+  }
+
+  const vendorDisplayName = function (kind) {
```
petersutter:
```suggestion
  function vendorDisplayName (kind) {
```

frontend/src/store/config.js
petersutter: do you think it makes sense to have a `vendorDetailsMap` computed prop? In order to compute it once? Something like the following

```js
  const vendorDetailsMap = computed(() => {
    const map = new Map()

    // Build map from all known and configured vendors
    const allVendors = new Set([
      ...knownVendors.map(v => v.name),
      ...configVendors.value.map(v => v.name),
    ])

    for (const kind of allVendors) {
      const configuredVendor = find(configVendors.value, ['name', kind])
      const knownVendor = find(knownVendors, ['name', kind])

      map.set(kind, {
        name: kind,
        weight: Number.MAX_SAFE_INTEGER,
        ...knownVendor,
        ...configuredVendor,
      })
    }

    return map
  })

  function vendorDetails (kind) {
    return vendorDetailsMap.value.get(kind) ?? {
      name: kind,
      weight: Number.MAX_SAFE_INTEGER,
    }
  }
```

frontend/src/components/GVendorIcon.vue
```diff
@@ -52,67 +58,22 @@ const props = defineProps({
...
-    case 'gardenlinux':
-      return new URL('/src/assets/gardenlinux.svg', import.meta.url)
-    case 'flatcar':
-      return new URL('/src/assets/flatcar.svg', import.meta.url)
-    case 'hcloud':
-      return new URL('/src/assets/hcloud.svg', import.meta.url)
+  if (iconName.value) {
+    if (startsWith(iconName, 'data:image/')) {
+      return iconName
+    }
```
petersutter: is there still a reason to support data uris?

frontend/src/components/GVendorIcon.vue
```diff
@@ -52,67 +58,22 @@ const props = defineProps({
...
+    return `/static/assets/${iconName.value}`
```
petersutter: assert for safe icon names

```js
    const safeIconNameRegex = /^[a-z0-9._-]+$/i
    if (safeIconNameRegex.test(iconName.value)) {
      return `/static/assets/${iconName.value}`
    }
```

frontend/src/assets/logo.svg
grolu: yes double checked, was not used at this specific location (is also under /static/assets)

frontend/src/components/GVendorIcon.vue
grolu: removed

frontend/src/store/config.js
grolu: Yes. While working on your feedback i stumbled over that comment that there could be possible name clashes between the different vendor arrays. As they are already separated in config, I changed the code to include a type in the vendorDetails lookup.
As I did not want to adjust every component that uses the function to pass the param, I do this internally only and log a warning in case there are ambiguous  or missing vendors. This, however slightly reduces the performance effect we gained by introducing the computed as we now have to loop over the types in the compatibility function. However this will change once we pass type from the components using the function

frontend/src/store/config.js
```diff
@@ -390,6 +428,286 @@ export const useConfigStore = defineStore('config', () => {
...
+    }
+
+    return {
+      name,
+      weight: Number.MAX_SAFE_INTEGER,
+    }
+  }
+
+  function vendorDisplayName (name, type) {
+    return get(vendorDetails(name, type), ['displayName'], name)
```
petersutter: vendorDetails only accepts one (name) param

frontend/src/store/config.js
```diff
@@ -390,6 +428,286 @@ export const useConfigStore = defineStore('config', () => {
...
+    },
+    {
+      name: 'rfc2136',
+      displayName: 'Dynamic DNS (RFC2136)',
+      weight: 10500,
+      icon: 'rfc2136.svg',
+    },
+  ]
+
+  const knownImageVendors = [
```
petersutter: shouldn't we better name it to `knownOSImageVendors` or `knownMachineImageVendors`? Or do you think it is clear enough?

grolu: done

docs/operations/customization.md
grolu: Added example in helm chart values.yaml

PR #2711, opened 2025-12-23
fix(backend) replace showdown (markdown) library

### LINKED ISSUE #2707: Replace showdownjs/showdown with alternative
showdownjs/showdown is not maintained anymore and should be replaced with an alternative

**Why is this needed**:

grolu:
## Summary

This PR replaces the **unmaintained `showdown` Markdown library** with a modern, well-maintained Markdown processing pipeline based on **unified / remark / rehype**.

The goal was to:

* Preserve GitHub-flavored Markdown behavior as closely as possible
* Improve long-term maintainability and security
* Avoid breaking our existing CommonJS-based test setup
* Keep the production API changes minimal and explicit

---

## Considerations: Choosing a Markdown Library

I evaluated several Markdown libraries with the requirement that **GitHub-style Markdown** should be supported either natively or via plugins.

#### `markdown-it`

* Core library is well maintained
* Plugin ecosystem is **partially unmaintained**
* Several important plugins are **CommonJS-only**
* ESM interop is problematic in modern builds

#### `marked`

* Actively maintained and fast
* Limited plugin ecosystem
* Missing features such as emoji support
* Would require manual handling for:

  * External links
  * Task lists
  * GitHub-specific behavior

#### `micromark`

* Excellent low-level parser
* Very small and strict API
* Plugin ecosystem is comparatively small
* Requires a lot of custom glue code

#### `unified / remark / rehype` (**chosen**)

* Actively maintained
* Widely adopted in the ecosystem
* Built on top of `micromark`
* Large plugin ecosystem for GitHub-flavored Markdown
* Clear separation between parsing, transformation, and output
* **Downside:** larger dependency footprint

Given the trade-offs, **unified/remark/rehype** was chosen as the best long-term solution.

---

## Handling the CommonJS Test Base

All modern Markdown libraries (except `markdown-it`) are **ESM-only**.
Our test suite, however, is still **CommonJS-based**.

To solve this cleanly:

* The actual Markdown implementation was moved to
  **`markdown.engine.mjs`** (pure ESM)
* Rollup is configured to **copy this file to `dist` without transpiling**
* `markdown.js` acts as a wrapper:

  * Production code imports the engine directly (ESM)
  * The transpiled `markdown.cjs` dynamically loads the engine via `await import()`
* The real `markdown.engine.md` is used by the markdown test suite only
* All other indirect dependencies to the markdown engine have been replaced by a global mock

This allows:

* Production code to use modern ESM
* Tests to run against the **real Markdown engine**
* Explicit, isolated async behavior

Additionally:

* A snapshot test suite (`markdown.spec.cjs`) was introduced
* This was done before the actual code changes to outline parsing differences
* The original Showdown output was captured first
* The new engine was iteratively aligned to match behavior where reasonable

---

## Final Markdown Engine (`markdown.engine.mjs`)

The final implementation uses the following pipeline:

* `remark-parse` – Markdown parsing
* `remark-gfm` – GitHub-flavored Markdown
* `remark-breaks` – GitHub-style line breaks
* `remark-emoji` – Emoji support
* `remark-rehype` – Markdown → HTML AST
* `rehype-slug` – Heading IDs
* `rehype-external-links` – External links (`target="_blank"`)
* `rehype-stringify` – HTML output
* `sanitize-html` – Final security sanitization

Raw HTML is **not parsed into the AST**, but emitted as raw HTML and sanitized afterward.

---

### Dropped features

* **Image dimension syntax**

  ```md
  ![alt](image.png =120x80)
  ```

  This is a Showdown-specific feature with no direct replacement.
  It is not standard GitHub Markdown and was removed intentionally.

### Sanitization improvements

* Explicit allow-list of GitHub-relevant tags:

  * `details`, `summary`
  * tables (`table`, `thead`, `tbody`, `tr`, `th`, `td`)
  * headings, code blocks, task lists, etc.
* Explicitly allowed attributes (`id`, `align`, `class`, etc.)
* GitHub task lists supported via **disabled checkbox inputs**
* `transformTags` ensures:

  * All links opened in a new tab have `noopener noreferrer`
  * Only **disabled checkbox inputs** are allowed

### Why still use `sanitize-html` (and not `rehype-sanitize`)?

While `rehype-sanitize` would integrate nicely with the AST and be slightly faster:

* `sanitize-html` is the **de-facto industry standard**
* It is widely audited and well understood
* We already need to sanitize **non-Markdown HTML** elsewhere (e.g. branding)
* Using a single sanitization library everywhere reduces risk and complexity

---

### Additional Helper: `breakAfterInlineHtmlBlock`

This PR also introduces a small helper function:

```js
breakAfterInlineHtmlBlock(input)
```

#### Why this is needed

When users paste **HTML blocks directly into Markdown** (for example `<details>`, `<table>`, or headings), Markdown parsers can sometimes **“swallow” the following Markdown content** if it appears immediately after a closing HTML tag without a blank line.

Example:

```md
<details>
<summary>More</summary>
Some text
</details>
This should be a new paragraph
```

Without an explicit blank line, many Markdown parsers interpret the following content as part of the HTML block, leading to unexpected rendering.

#### What the function does

* Scans the Markdown **source text** (before parsing)
* Detects closing HTML tags that commonly start HTML blocks
* Ensures there is **at least one blank line (`\n\n`) after the closing tag**
* Only applies the fix **when non-whitespace content follows**
* Leaves all other content untouched

#### Why not solve this in the parser?

This behavior is part of the **Markdown specification’s HTML block rules**, not a bug in `remark` or `rehype`.
Applying this normalization **before parsing** avoids complex AST manipulation and keeps the pipeline predictable.

---

## Async API Changes

The new Markdown engine is **asynchronous**.

To handle this safely in production code:

* All call sites were updated accordingly
* In the config service, a **lazy cache** was introduced:

  * Sanitization runs on first request
  * Results are cached (including in-flight promises)
  * No top-level `await` is required

⚠️ Note: This cache does **not** currently handle live configuration updates.
This matches existing behavior and can be improved separately if needed.

---

## Conclusion

This PR modernizes Markdown handling with a secure, extensible, and well-maintained solution while:

* Preserving GitHub-style behavior
* Keeping test compatibility
* Minimizing production code changes
* Improving long-term maintainability and security

---

Fixes #2707

**Release note**:

```other operator
```


klocke-io (approved):
/lgtm

backend/lib/routes/config.js
```diff
@@ -15,14 +15,37 @@ import { metricsRoute } from '../middleware.js'
...
+  if (!cachedFrontendConfigPromise) {
+    cachedFrontendConfigPromise = sanitizeFrontendConfig(config.frontend)
+      .then(cfg => {
+        cachedFrontendConfig = cfg
+        return cfg
+      })
+      .catch(err => {
+        cachedFrontendConfigPromise = undefined
+        throw err
+      })
```
petersutter: use async/await

backend/lib/routes/config.js
```diff
@@ -38,12 +61,18 @@ router.route('/')
     }
   })
 
-function sanitizeFrontendConfig (frontendConfig) {
+async function sanitizeFrontendConfig (frontendConfig) {
```
petersutter, line 57: I'd manage the task queue outside of the function, even if it means that more lines have to be touched. Also in my suggestion I switched to async/await

```js
async function sanitizeFrontendConfig (frontendConfig) {
  const converter = createConverter()
  const tasks = []

  const convertAndSanitize = async (obj, key) => {
    const value = obj?.[key] // eslint-disable-line security/detect-object-injection -- key is a local fixed string
    if (value) {
      const html = await converter.makeSanitizedHtml(value)
      obj[key] = html // eslint-disable-line security/detect-object-injection -- key is a local fixed string
    }
  }

  const sanitizedFrontendConfig = _.cloneDeep(frontendConfig)
  const {
    alert = {},
    costObjects = [],
    sla = {},
    addonDefinition = {},
    accessRestriction: {
      items = [],
    } = {},
    vendorHints = [],
    resourceQuotaHelp = {},
    controlPlaneHighAvailabilityHelp = {},
  } = sanitizedFrontendConfig

  tasks.push(convertAndSanitize(alert, 'message'))

  for (const costObject of costObjects) {
    tasks.push(convertAndSanitize(costObject, 'description'))
  }

  tasks.push(convertAndSanitize(sla, 'description'))
  tasks.push(convertAndSanitize(addonDefinition, 'description'))
  tasks.push(convertAndSanitize(resourceQuotaHelp, 'text'))
  tasks.push(convertAndSanitize(controlPlaneHighAvailabilityHelp, 'text'))

  for (const item of items) {
    const {
      display = {},
      input = {},
      options = [],
    } = item
    tasks.push(convertAndSanitize(display, 'description'))
    tasks.push(convertAndSanitize(input, 'description'))
    for (const option of options) {
      const {
        display = {},
        input = {},
      } = option
      tasks.push(convertAndSanitize(display, 'description'))
      tasks.push(convertAndSanitize(input, 'description'))
    }
  }

  for (const vendorHint of vendorHints) {
    tasks.push(convertAndSanitize(vendorHint, 'message'))
  }

  await Promise.all(tasks)
  return sanitizedFrontendConfig
}
```

backend/lib/services/terminals/index.js
```diff
@@ -589,7 +589,7 @@ async function getOrCreateTerminalSession ({ user, namespace, name, target, body
       kubeApiServer: hostCluster.kubeApiServer,
       namespace: terminal.spec.host.namespace,
     },
-    imageHelpText: imageHelpText(terminal),
+    imageHelpText: await imageHelpText(terminal),
```
petersutter: assign `await imageHelpText(terminal)` to a variable

backend/lib/services/tickets.js
```diff
@@ -59,15 +59,15 @@ export function fromIssue (issue) {
         'comments',
       ])
       .assign({
-        body: converter.makeSanitizedHtml(issue.body),
+        body: await converter.makeSanitizedHtml(issue.body),
```
petersutter: also here and in line 92 I would assign it to a variable

backend/lib/routes/config.js
```diff
@@ -15,14 +15,38 @@ import { metricsRoute } from '../middleware.js'
...
+        cachedFrontendConfig = cfg
+        return cfg
+      } catch (err) {
+        cachedFrontendConfigPromise = undefined
+        throw err
+      }
+    })()
+  }
+  return cachedFrontendConfigPromise
+}
```
petersutter: Removed redundant `cachedFrontendConfig` variable and now cache solely via the promise, which memoizes the result naturally. Reduces state while keeping identical behavior and retry-on-error.

```suggestion
let cachedFrontendConfigPromise

async function getFrontendConfig() {
  if (cachedFrontendConfigPromise) {
    return cachedFrontendConfigPromise
  }

  cachedFrontendConfigPromise = (async () => {
    try {
      return await sanitizeFrontendConfig(config.frontend)
    } catch (err) {
      cachedFrontendConfigPromise = undefined
      throw err
    }
  })()

  return cachedFrontendConfigPromise
}
```

backend/lib/services/terminals/index.js
```diff
@@ -84,11 +84,11 @@ function toTerminalMetadata (terminal) {
   return metadata
 }
 
-function imageHelpText (terminal) {
+async function getImageHelpText (terminal) {
   const containerImage = _.get(terminal, ['spec', 'host', 'pod', 'container', 'image'])
   const containerImageDescriptions = getConfigValue('terminal.containerImageDescriptions', [])
   const containerImageDescription = findImageDescription(containerImage, containerImageDescriptions)
-  return converter.makeSanitizedHtml(containerImageDescription)
+  return await converter.makeSanitizedHtml(containerImageDescription)
```
petersutter: omit unnecessary await

backend/lib/services/tickets.js
```diff
@@ -105,7 +107,7 @@ export async function getOpenIssues ({ name, projectName } = {}) {
     title = `[${projectName}/${name}]`
   }
   const githubIssues = await searchIssues({ state: 'open', title })
-  return _.map(githubIssues, fromIssue)
+  return Promise.all(_.map(githubIssues || [], fromIssue))
```
petersutter, line 110: did you encounter a case where adding the fallback ` || []` was necessary?

backend/lib/markdown.engine.mjs
```diff
@@ -0,0 +1,46 @@
...
+import sanitizeHtml from 'sanitize-html'
+
+const SANITIZE = {
+  allowedTags: [...sanitizeHtml.defaults.allowedTags, 'img', 'details', 'summary'],
+}
+
+const processor = unified()
+  .use(remarkParse)
+  .use(remarkGfm)
+  .use(remarkGithub)   // ← add this
```
petersutter:
```suggestion
  .use(remarkGithub)
```

backend/lib/services/tickets.js
grolu: Yes I think if the request failed it was undefined but not 100% sure

backend/lib/routes/config.js
grolu: Nice, way better

backend/lib/markdown.engine.mjs
petersutter, line 20: add comment when fully migrated to esm

backend/lib/services/tickets.js
petersutter: then in line 134, do we also need the fallback out of the same reasons or does it behave differently and therefore not needed?
`return Promise.all(_.map(githubComments, githubComment =>`

petersutter (approved):
/approve

backend/lib/services/tickets.js
grolu: hm probably this error did never occur because if issues failed with error it will never try to load comments, but yes if this fails we should also have a fallback. I will add it

PR #2730, opened 2026-01-22
fix: disable eager loading for tooltips globally

petersutter:
Disables eager loading for VTooltip components by setting `eager: false` in Vuetify defaults configuration.

This improves performance by ensuring tooltips are only mounted when actually needed (on hover/focus), rather than being pre-rendered in the DOM. 

Fixes #

**Release note**:

```bugfix user
```


grolu (approved):
/approve

PR #2773, opened 2026-02-10
feat(frontend): add robots.txt and meta tag to prevent indexing

### LINKED ISSUE #2635: Discourage search engines from indexing Gardener dashboard

Search engines (such as Google) are currently indexing Gardener dashboard installations and show them in search results (e.g., when searching for "gardener dashboard").

While security through obfuscation is an anti-pattern, there's still not much value in having Gardener dashboard installations show up in search results (at least not by default).

**Why is this needed**:

There is no value in having Gardener dashboard URLs showing up in search results. At least, it should be the operator's choice whether indexing (by well-behaved) crawlers should happen or not.

klocke-io:
Adds robots.txt and meta tag to prevent indexing by search engines
Fixes #2635

**Release note**:

```feature user
Prevent search engines from crawling the dashboard UI. The dashboard will no longer be automatically indexed
```


frontend/public/robots.txt
```diff
@@ -0,0 +1,2 @@
+User-agent: *
+Disallow: /
```
petersutter: nit: newline
```suggestion
Disallow: /

```

petersutter (approved):
/lgtm

grolu (approved):
/lgtm

PR #2775, opened 2026-02-12
Fix incorrect warning message for forced machine image updates

### LINKED ISSUE #2772: Misleading image version expiration message
**What happened**:
The message about an expiring machine image version is misleading with regards to downgrade/upgrade options:
If a machine image has `updatePolicy: minor` set and the cluster is at the highest minor that is about to expire, but there are newer majors available, the error message states:
> Machine image gardenlinux | Version: 1443.20.0 of worker group worker-abcd: Image expires in 9 days.
Machine image version will expire soon and **there is no newer supported version available**.
However, an **older** supported version for this image vendor is available. You may want to consider **downgrading** to that version. 

The message should suggest to upgrade (or be neutral about the direction) and not state that there are no newer versions available.

https://github.com/gardener/dashboard/blob/9a0a089add857d8d1a960ab6cd28aba7d31e90e2/frontend/src/components/ShootMessages/GWorkerGroupExpirationMessage.vue#L33-L35

**Environment**:

grolu:
Improve code readability by explicitly passing regularUpdate, forcedUpdate, and noUpdate parameters instead of using severity, which had implicit semantics Align Kubernetes machine image expiration message

Clusters that had no update for the `updateStrategy` (e.g. *minor*) defined in cloud profile are incorrectly flagged as `error` even if there is a higher supported version available (e.g. *major*)
Screenshot 2026-02-12 at 13 20 13

This PR ensures correct handling for this case
Screenshot 2026-02-12 at 13 19 41

Fixes #2772

**Release note**:

```bugfix user
Fixed an incorrect warning message for forced machine image update
```

## Summary by CodeRabbit

* **New Features**
  - Clearer expiration messaging with explicit version labels and distinct regular/forced/no-update states.
  - Improved machine image auto-update eligibility that respects update strategy and informs hint severity.

* **Tests**
  - Updated test suites to validate the new expiration messages, update-availability flags, and updated utility behavior.

### DIFF
--- a/frontend/src/components/ShootMessages/GK8sExpirationMessage.vue
+++ b/frontend/src/components/ShootMessages/GK8sExpirationMessage.vue
@@ -18,12 +18,13 @@ SPDX-License-Identifier: Apache-2.0
       <span>. </span>
     </span>
     <span v-else-if="isExpired">is expired. </span>
+    <span v-else>will expire soon. </span>
     <span v-if="regularUpdate">Version will be updated in the next maintenance window</span>
     <template v-else-if="forcedUpdate">
       <span v-if="isValidTerminationDate">Version update will be enforced after that date</span>
       <span v-else>Version update will be enforced soon</span>
     </template>
-    <span v-else-if="noUpdate">Version of this cluster will expire soon and there is no supported version available. Please contact your landscape administrator</span>
+    <span v-else-if="noUpdate">There is no supported version available that you can update to. Please contact your landscape administrator</span>
   </div>
 </template>
 
--- a/frontend/src/components/ShootMessages/GK8sExpirationMessage.vue
+++ b/frontend/src/components/ShootMessages/GK8sExpirationMessage.vue
@@ -6,7 +6,9 @@ SPDX-License-Identifier: Apache-2.0
 
 <template>
   <div>
-    <span v-if="isValidTerminationDate">Kubernetes version of this cluster expires
+    Kubernetes <span class="font-weight-bold">version {{ version }}</span> of this cluster
+    <span v-if="isValidTerminationDate">
+      expires
       <g-time-string
         :date-time="expirationDate"
         mode="future"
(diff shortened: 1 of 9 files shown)


frontend/src/components/ShootMessages/GK8sExpirationMessage.vue
```diff
@@ -15,13 +17,13 @@ SPDX-License-Identifier: Apache-2.0
...
-    <span v-if="severity === 'info'">Version will be updated in the next maintenance window</span>
-    <template v-else-if="severity === 'warning'">
+    <span v-else-if="isExpired">is expired. </span>
+    <span v-if="regularUpdate">Version will be updated in the next maintenance window</span>
+    <template v-else-if="forcedUpdate">
       <span v-if="isValidTerminationDate">Version update will be enforced after that date</span>
       <span v-else>Version update will be enforced soon</span>
     </template>
-    <span v-else-if="severity === 'error'">Kubernetes version of this cluster will expire soon and there is no newer supported version available. Please contact your landscape administrator</span>
+    <span v-else-if="noUpdate">Version of this cluster will expire soon and there is no supported version available. Please contact your landscape administrator</span>
```
petersutter: "will expire soon" might not be correct (see `isExpired`).
Are the texts correct for the different combinations?

frontend/src/components/ShootMessages/GShootMessages.vue
```diff
@@ -218,8 +219,11 @@ const k8sMessage = computed(() => {
       props: {
         expirationDate,
         isValidTerminationDate,
-        severity,
         isExpired,
+        version,
+        regularUpdate: severity === 'info',
+        forcedUpdate: severity === 'warning',
+        noUpdate: severity === 'error',
```
petersutter: check if we can get rid of the severity logic here and use the proper source

frontend/src/components/ShootMessages/GWorkerGroupExpirationMessage.vue
```diff
@@ -17,13 +17,13 @@ SPDX-License-Identifier: Apache-2.0
       />
       <span>. </span>
     </span>
-    <span v-else-if="isExpired">Image is expired. </span>
-    <span v-if="severity === 'info'">Version will be updated in the next maintenance window</span>
-    <template v-else-if="severity === 'warning'">
+    <span v-else-if="isExpired"> is expired. </span>
```
petersutter, line 20: there could be case, where the sentence is not terminated, as there is no else case. The same is true for  frontend/src/components/ShootMessages/GK8sExpirationMessage.vue

grolu: This combination should be mutually impossible.
However I added
`<span v-else>will expire soon. </span>`
just to be safe...
Also did this for GK8sExpirationMessage

frontend/src/components/ShootMessages/GK8sExpirationMessage.vue
grolu: well the message is only shown if an expiration warning is triggered because some kind of expiration detection kicks in.
So it should always be a validTerminationDate (date not yet passed) OR it is already expired then the update is about to happen.
However, it is kind of repetitive and unclear if it is expired about to expire or whatever, so I adapted the text to be
`<span v-else-if="noUpdate">There is no supported version available that you can update to. Please contact your landscape administrator</span>`

Also did this for WorkerGroupExpiration component

frontend/src/components/ShootMessages/GShootMessages.vue
grolu: refactored and combined `getVersionExpirationWarning` functions to return also these variables and re-use in both composables.

petersutter (commented):
/lgtm
/approve

PR #2818, opened 2026-03-04
Removed next callback in navigation guards

grolu:

Fixes #

**Release note**:

```bugfix user

```

## Summary by CodeRabbit

* **Refactor**
  * Modernized navigation guard patterns across the app for more consistent navigation behavior
  * Moved component initialization from route-enter hooks to lifecycle mount for more reliable loading
  * Centralized login error handling into store state so errors persist and display predictably

* **Tests**
  * Adjusted tests to reflect the updated navigation and login error flows

### DIFF
--- a/frontend/src/layouts/GLogin.vue
+++ b/frontend/src/layouts/GLogin.vue
@@ -151,7 +151,7 @@ export default {
     GNotify,
   },
   inject: ['api'],
-  async beforeRouteEnter (to, from, next) {
+  async beforeRouteEnter (to) {
     let err
     if (/^#.+/.test(to.hash)) {
       const searchParams = new URLSearchParams(to.hash.substring(1))
@@ -168,31 +168,18 @@ export default {
       const redirectPath = get(to.query, ['redirectPath'], '/')
       const authnStore = useAuthnStore()
       authnStore.signinWithOidc(redirectPath)
-      return next(false)
+      return false
     }
 
-    next(vm => {
-      if (err) {
-        if (err.message !== 'NoAutoLogin') {
-          vm.error = err
-        }
-        vm.$router.replace('/login')
-      }
-    })
-  },
-  beforeRouteUpdate (to, from, next) {
-    if (this.error) {
-      const err = this.error
-      this.error = null
-      this.setError(err)
+    if (err && err.message !== 'NoAutoLogin') {
+      loginStore.loginError = err
+      return { path: '/login', replace: true }
     }
-    next()
   },
   data () {
     return {
       showToken: false,
       token: '',
-      error: null,
     }
   },
   computed: {
@@ -203,6 +190,7 @@ export default {
     ]),
     ...mapWritableState(useLoginStore, [
       'loginType',
+      'loginError',
     ]),
     breakpointName () {
       return this.$vuetify.display.name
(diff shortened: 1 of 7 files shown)


frontend/src/layouts/GLogin.vue
```diff
@@ -281,6 +269,13 @@ export default {
       immediate: true,
     },
   },
+  mounted () {
+    if (this.loginError) {
```
petersutter, line 275: should this be watched instead of handling in mounted?

frontend/src/views/GShootList.vue
```diff
@@ -692,6 +688,9 @@ export default {
       }
     },
   },
+  mounted () {
+    this.updateTableSettings()
+  },
```
petersutter: drop.

should not be necessary with https://github.com/gardener/dashboard/pull/2911 - "Replace imperative updateTableSettings() calls in lifecycle hooks and
route guards with reactive watchers on column/sortBy state."

frontend/src/layouts/GLogin.vue
```diff
@@ -168,31 +168,18 @@ export default {
...
-    })
-  },
-  beforeRouteUpdate (to, from, next) {
-    if (this.error) {
-      const err = this.error
-      this.error = null
-      this.setError(err)
+    if (err && err.message !== 'NoAutoLogin') {
+      loginStore.loginError = err
+      return { path: '/login', replace: true }
```
petersutter: if the route is e.g. /login#error=NoAutoLogin, old behavior cleaned the URL by replacing to /login; new behavior leaves the hash on the URL bar

frontend/src/views/GShootList.vue
```diff
@@ -191,14 +191,10 @@ export default {
     // Reset expanded state in case project changes
     this.resetState(this.expandedWorkerGroups, { default: false })
     this.resetState(this.expandedAccessRestrictions, { default: false })
-
-    next()
   },
-  beforeRouteLeave (to, from, next) {
+  beforeRouteLeave () {
```
petersutter, line 195: beforeRouteLeave does not return true, in the other file it returns true / false. -> inconsistent

frontend/src/layouts/GLogin.vue
grolu: I think we don't need to watch it as the only place where it is set is in `beforeRouteEnter` -> we temporarily store it and when the component mounts we read it from the login store

frontend/src/layouts/GLogin.vue
grolu: moved path replace to if(err) as before

frontend/src/views/GShootList.vue
grolu: `beforeRouteLeave` return `true` and return nothing / `undefined` has the same behavior: Continue navigation, return `false` cancels navigation. So in the other file we return true for early exit / false for cancelling the navigation. We could change the code to always return a boolean if you think this would be cleaner.

petersutter: it's a minor inconsistency, no blocker. we can still merge

petersutter (commented):
/lgtm
/approve

PR #2827, opened 2026-03-11
Do not set `spec.addons` in case all addons are disabled

grolu:

Fixes #

**Release note**:

```bugfix user

```

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed addon handling: disabled addon sets are removed from manifests; enabling an addon recreates addon entries cleanly without retaining stale entries.

* **Tests**
  * Added tests for addon behavior: omission when none are enabled, correct population when an addon is enabled, and recreation when enabling after a manifest without addons.

### DIFF
--- a/frontend/src/composables/useShootContext.js
+++ b/frontend/src/composables/useShootContext.js
@@ -96,15 +96,6 @@ export function createShootContextComposable (options = {}) {
     return some(value, ['enabled', true])
   }
 
-  function ensureAddonsObject () {
-    let value = get(manifest.value, ['spec', 'addons'])
-    if (!value) {
-      value = {}
-      set(manifest.value, ['spec', 'addons'], value)
-    }
-    return value
-  }
-
   function normalizeManifest (value) {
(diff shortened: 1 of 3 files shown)


frontend/src/composables/useShootContext.js
```diff
@@ -91,11 +92,18 @@ export function createShootContextComposable (options = {}) {
...
+  function hasEnabledAddons (value) {
+    return some(value, ['enabled', true])
+  }
+
   function normalizeManifest (value) {
     const object = Object.assign({
       apiVersion: 'core.gardener.cloud/v1beta1',
       kind: 'Shoot',
     }, value)
+    if (!hasEnabledAddons(get(object, ['spec', 'addons']))) {
```
petersutter, line 104: I'd pass the object directly to hasEnabledAddons util function, but that is just my preference and we also soon will get rid of this code, so no need to change

petersutter (commented):
/lgtm

PR #2838, opened 2026-03-17
feat(frontend) WorkloadIdentity Support for DNS Configuration

grolu:

Fixes #

**Release note**:

```feature user
Added support for using Workload Identities when managing DNS providers in the dashboard
```

```other user
Refactored DNS secret handling to align with the new specification and remove deprecated fields
```

## Summary by CodeRabbit

* **New Features**
  * Components now accept structured credential objects and can hide the credential icon.

* **Improvements**
  * Unified credential model (kind + name) across UI and selection flows.
  * Automatic, backward-compatible migration from legacy secretName to structured credentials.
  * Consistent enrichment of resource metadata and new helpers for credential/resource naming.

* **Tests**
  * Expanded test coverage for credential types, migration paths, and list normalization.


frontend/src/composables/credential/helper.js
```diff
@@ -18,31 +19,76 @@ export function isWorkloadIdentity (credential) {
...
 
+// DNS Provider references
+export function dnsProviderCredentialsRef (provider) {
+  if (!provider) {
+    return undefined
+  }
+  if (provider.credentialsRef) {
+    return provider.credentialsRef
+  }
+  // Support legacy field secretName for backward compatibility, but prefer credentialsRef if both are set
```
petersutter, line 88:
```suggestion
  // Support legacy field secretName for backward compatibility, but prefer credentialsRef if both are set
  // TODO(grolu): drop support for legacy field secretName
```

frontend/src/components/ShootDns/GDnsProviderRow.vue
```diff
@@ -239,15 +243,27 @@ export default {
...
+      const { metadata: { name }, kind } = credential
+      if (!name || !kind) {
+        return false
+      }
+      const resourceName = dnsExtensionProviderResourceName(this.dnsProvider)
+      const currentResourceRef = this.getResourceRef(resourceName)
+      if (name === currentResourceRef?.name && kind === currentResourceRef?.kind) {
+        return true
+      }
+      return !this.usedResourceRefs.some(ref => ref?.name === name && ref?.kind === kind)
```
petersutter: - align it with the style already used elsewhere in this file and nearby code by avoiding nested destructuring and using explicit variables
- avoid same name/kind comparison by using a tiny local `matchesCredential` helper which also makes the intent a bit clearer

```suggestion
      const credentialName = credential?.metadata?.name
      const credentialKind = credential?.kind
      if (!credentialName || !credentialKind) {
        return false
      }
      const currentResourceRef = this.getResourceRef(dnsExtensionProviderResourceName(this.dnsProvider))
      const matchesCredential = ref => ref?.name === credentialName && ref?.kind === credentialKind

      if (matchesCredential(currentResourceRef)) {
        return true
      }
      return !this.usedResourceRefs.some(matchesCredential)
```

frontend/src/components/ShootDns/GDnsProviderRow.vue
```diff
@@ -239,15 +243,27 @@ export default {
         set(this.dnsProvider, ['zones', 'include'], value)
       },
     },
-    usedSecretNames () {
-      return this.dnsServiceExtensionProviders.map(({ secretName }) => this.getResourceRefName(secretName))
+    usedResourceRefs () {
```
petersutter: rename to `dnsExtensionProviderResourceRefs`

frontend/src/components/ShootDns/GManageDns.vue
```diff
@@ -269,11 +270,18 @@ export default {
     primaryDnsProviderCredential: {
       get () {
         return find(this.dnsPrimaryProviderCredentials, credential => {
```
petersutter: I was confused with `dnsPrimaryProviderCredentials` naming, wouldn't  `eligibleDnsPrimaryProviderCredentials` be a better choice? No strong opinion though

frontend/src/components/ShootDns/GManageDns.vue
```diff
@@ -269,11 +270,18 @@ export default {
...
-          return credential?.metadata?.name === this.dnsPrimaryProviderSecretName
+          return credential?.metadata?.name === this.dnsPrimaryProviderCredentialsRef?.name &&
+            credential?.kind === this.dnsPrimaryProviderCredentialsRef?.kind
         })
```
petersutter: consider to extract predicate to variable for better readability
```js
primaryDnsProviderCredential: {
      get () {
        const matchesPrimaryDnsProviderCredential = credential => {
          return credential?.metadata?.name === this.dnsPrimaryProviderCredentialsRef?.name &&
            credential?.kind === this.dnsPrimaryProviderCredentialsRef?.kind
        }

        return find(this.dnsPrimaryProviderCredentials, matchesPrimaryDnsProviderCredential)
      },
// ...
```

frontend/src/components/ShootDns/GManageDns.vue
```diff
@@ -269,11 +270,18 @@ export default {
...
       },
       set (credential) {
-        this.dnsPrimaryProviderSecretName = credential?.metadata?.name
+        this.dnsPrimaryProviderCredentialsRef = credential
+          ? {
+              apiVersion: credential.apiVersion,
+              kind: credential.kind,
+              name: credential.metadata?.name,
+            }
+          : undefined
```
petersutter: not sure if I like that multiline ternary. Might as well be something we want to avoid using eslint rule :D 

the alternative would be the following, which is slightly more explicit/readable. No strong opinion, feel free to keep as is

```js
if (!credential) {
  this.dnsPrimaryProviderCredentialsRef = undefined
  return
}

this.dnsPrimaryProviderCredentialsRef = {
  apiVersion: credential.apiVersion,
  kind: credential.kind,
  name: credential.metadata?.name,
}
```

frontend/src/components/ShootDns/GManageDns.vue
```diff
@@ -301,9 +309,19 @@ export default {
...
   },
   methods: {
     getErrorMessages,
+    dnsProviderRowKey (provider) {
+      return dnsExtensionProviderResourceName(provider) ?? JSON.stringify({
+        type: provider?.type,
+        domains: provider?.domains,
+        zones: provider?.zones,
+      })
+    },
```
petersutter: let's talk about that

frontend/src/composables/credential/helper.js
```diff
@@ -18,31 +18,88 @@ export function isWorkloadIdentity (credential) {
...
 export function credentialProviderType (credential) {
-  // TODO check for provider.extensions.gardener.cloud once wlids are supported
   const labels = credential?.metadata?.labels
-  if (!labels) {
-    return undefined
-  }
-
   const DASHBOARD = 'dashboard.gardener.cloud/dnsProviderType'
-  const PREFIX = 'provider.shoot.gardener.cloud/'
+  const SHOOT_PREFIX = 'provider.shoot.gardener.cloud/'
```
petersutter: how about:

```js
export function credentialProviderType (credential) {
  const labels = credential?.metadata?.labels
  if (!labels) {
    return undefined
  }

  // For DNS credentials: prefer the dashboard-specific label
  const dnsProviderType = labels['dashboard.gardener.cloud/dnsProviderType']
  if (dnsProviderType) {
    return dnsProviderType
  }

  const providerPrefixes = [
    'provider.shoot.gardener.cloud/',     // Infrastructure credentials referenced on the shoot (e.g. Shoot.spec.credentialsBindingName)
    'provider.extensions.gardener.cloud/', // WorkloadIdentities referenced via resource ref in extensions
  ]

  for (const prefix of providerPrefixes) {
    const key = Object.keys(labels).find(k => k.startsWith(prefix) && isTruthyValue(labels[k]))
    if (key) {
      return key.slice(prefix.length)
    }
  }

  return undefined
}
```

frontend/src/composables/credential/useCloudProviderCredential.js
```diff
@@ -45,10 +45,15 @@ export const useCloudProviderCredential = (credential, options = {}) => {
   })
 
   const credentialUsageCount = computed(() => {
-    // TODO: Add count for wlid once supported
     const name = credentialName.value
-    const kind = credentialKind.value || 'Secret'
-    const byProvider = providers => some(providers, ['secretName', name])
+    const kind = credentialKind.value
+    const byProvider = providers => some(providers, provider => {
```
petersutter: how about
```js
  const credentialUsageCount = computed(() => {
    const name = credentialName.value
    const kind = credentialKind.value
    const isCredentialReferencedByDnsProvider = shoot => some(shoot.spec?.dns?.providers, provider => {
      if (provider?.credentialsRef) {
        return provider.credentialsRef.name === name && provider.credentialsRef.kind === kind
      }
      // secretName is supported for backward compatibility, but credentialsRef is preferred if both are set
      return kind === 'Secret' && provider?.secretName === name
    })
    const isCredentialReferencedByResource = shoot => some(shoot.spec?.resources, { resourceRef: { kind, name } })

    let count = 0
    for (const shoot of shootList.value) {
      if (isCredentialReferencedByDnsProvider(shoot) || isCredentialReferencedByResource(shoot)) {
        count++
      }
    }
    return count
  })
```

frontend/src/composables/useShootDns.js
```diff
@@ -74,12 +80,33 @@ export const useShootDns = (manifest, options) => {
...
+  }
+
   /* dns */
   const dnsDomain = computed({
     get () {
       return get(manifest.value, ['spec', 'dns', 'domain'])
     },
     set (value) {
+      normalizeDnsServiceExtensionProviders()
+      normalizeDnsPrimaryProviderCredentialsRefs()
```
petersutter: should we add a TODO to drop these normalize functions once the legacy fields are dropped from the API?

frontend/src/composables/useShootDns.js
```diff
@@ -125,12 +172,16 @@ export const useShootDns = (manifest, options) => {
...
-  const dnsPrimaryProviderSecretName = computed({
+  const dnsPrimaryProviderCredentialsRef = computed({
     get () {
-      return dnsPrimaryProvider.value?.secretName
+      return getDnsPrimaryProviderCredentialsRef(dnsPrimaryProvider.value)
     },
-    set (secretName) {
-      patchDnsPrimaryProvider({ secretName })
+    set (credentialsRef) {
+      const provider = omit(dnsPrimaryProvider.value, ['secretName'])
```
petersutter: - The setter manually strips `secretName` and spreads the full provider, but `patchDnsPrimaryProvider` already does the same normalization internally
- The setter could be simplified to just `patchDnsPrimaryProvider({ credentialsRef })`

frontend/src/composables/useShootDns.js
```diff
@@ -139,22 +190,32 @@ export const useShootDns = (manifest, options) => {
...
+      const resourceName = dnsExtensionProviderResourceName(provider)
+      const providerCredentialRef = getResourceRef(resourceName)
       return provider.type === dnsPrimaryProvider.value.type &&
-        providerSecretName === dnsPrimaryProvider.value.secretName
+        providerCredentialRef?.name === dnsPrimaryProviderCredentialsRef.value?.name &&
+        providerCredentialRef?.kind === dnsPrimaryProviderCredentialsRef.value?.kind
     })
   })
 
+  const primaryDnsServiceExtensionProviderResourceRef = computed(() => {
```
petersutter: is this acutally used?

frontend/src/composables/useShootDns.js
```diff
@@ -170,54 +231,89 @@ export const useShootDns = (manifest, options) => {
     }
   }
 
-  function getDefaultSecretName (type) {
+  function getDefaultCredential (type) {
     const credentials = useCloudProviderEntityList(toRef(type), { credentialStore, gardenerExtensionStore, cloudProfileStore })
```
petersutter: `useCloudProviderEntityList`:
- extract a pure helper, e.g. `getCloudProviderEntityList(providerType, { credentialStore, gardenerExtensionStore, cloudProfileStore })`, returning a plain array
- keep `useCloudProviderEntityList(ref, options)` as the reactive wrapper: `computed(() => getCloudProviderEntityList(providerType.value, options))`
- update `useShootDns` to use the pure helper in imperative code paths (`getDefaultCredential`, `addDnsServiceExtensionProvider`), while components like `GDnsProviderRow` and `GManageDns` continue using the composable with a real ref

also rename `credentials` to something like `credentialsForProviderType`

and how about the following (keep in mind this is pre-refactoring, so the above mentioned suggestion is not included below, but makes it more readable)
```js
function getDefaultCredential(providerType) {
  const credentials = useCloudProviderEntityList(toRef(providerType), {
    credentialStore,
    gardenerExtensionStore,
    cloudProfileStore,
  })

  const usedResourceRefs = map(resources.value, 'resourceRef')

  const isCredentialUsed = credential => {
    const kind = credential?.kind
    const name = credential?.metadata?.name

    return usedResourceRefs.some(resourceRef =>
      resourceRef?.kind === kind &&
      resourceRef?.name === name
    )
  }

  return credentials.value.find(credential => !isCredentialUsed(credential))
}
```

frontend/src/composables/useShootDns.js
```diff
@@ -170,54 +231,89 @@ export const useShootDns = (manifest, options) => {
...
+        return resourceRef?.kind === credentialKind &&
+          resourceRef?.name === credentialName
+      })
     })
 
-    return credential ? credential?.metadata?.name : undefined
+    return credential
   }
 
   function addDnsServiceExtensionProvider (options = {}) {
```
petersutter: should we rename the `options` param to `provider`? 
also let's talk about this function, this looks overly complex.

frontend/src/composables/useShootDns.js
```diff
@@ -170,54 +231,89 @@ export const useShootDns = (manifest, options) => {
...
-    const {
-      type = head(gardenerExtensionStore.dnsProviderTypes),
-      secretName = getDefaultSecretName(type),
-    } = options
+    normalizeDnsServiceExtensionProviders()
+    normalizeDnsPrimaryProviderCredentialsRefs()
+
+    const type = options.type ?? head(gardenerExtensionStore.dnsProviderTypes)
```
petersutter: also here, rename to `providerType`

frontend/src/composables/useShootDns.js
```diff
@@ -170,54 +231,89 @@ export const useShootDns = (manifest, options) => {
...
+          kind: 'Secret',
+          name: options.secretName,
+        }
+      : defaultCredential
+        ? {
+            apiVersion: defaultCredential.apiVersion,
+            kind: defaultCredential.kind,
+            name: defaultCredential.metadata?.name,
+          }
+        : undefined)
```
petersutter: please make this more readable

frontend/src/composables/useShootDns.js
```diff
@@ -170,54 +231,89 @@ export const useShootDns = (manifest, options) => {
...
+
+    const credentials = useCloudProviderEntityList(toRef(type), { credentialStore, gardenerExtensionStore, cloudProfileStore })
```
petersutter: rename credentials to `credentialsForProviderType` (as above)

frontend/src/composables/useShootDns.js
```diff
@@ -170,54 +231,89 @@ export const useShootDns = (manifest, options) => {
...
+        resourceRef,
+      })
+    }
 
     return provider
   }
 
   function deleteDnsServiceExtensionProvider (index) {
+    normalizeDnsServiceExtensionProviders()
+    normalizeDnsPrimaryProviderCredentialsRefs()
```
petersutter: let's also talk about that

frontend/src/composables/credential/helper.js
```diff
@@ -18,31 +18,88 @@ export function isWorkloadIdentity (credential) {
...
+      apiVersion: 'v1',
+      kind: 'Secret',
+      name: provider.secretName,
+    }
+  }
+  return undefined
+}
+
+export function dnsExtensionProviderResourceName (provider) {
+  // secretName is supported for backward compatibility, but credentialsRef is preferred if both are set
```
petersutter, line 101: also add TODO here to drop legacy field support

packages/kube-client/lib/mixins.js
```diff
@@ -533,6 +534,11 @@ function normalizeSearchParams (method, searchParams, options) {
   return normalizedSearchParams
 }
 
+function normalizeListResponse (response, Resource) {
+  return Promise.resolve(response)
+    .then(body => normalizeResourceListItems(body, Resource))
+}
```
petersutter:
```suggestion
async function normalizeListResponse (response, Resource) {
  const body = await response
  return normalizeResourceListItems(body, Resource)
}
```

packages/kube-client/lib/mixins.js
```diff
@@ -72,7 +73,7 @@ ClusterScoped.Readable = superclass => class extends superclass {
     const method = 'get'
     const url = clusterScopedUrl(this.constructor.names)
     searchParams = normalizeSearchParams(method, searchParams, options)
-    return this[http.request](url, { method, searchParams })
+    return normalizeListResponse(this[http.request](url, { method, searchParams, signal }), this.constructor)
```
petersutter: split into two lines for better readability, like

```js
const response = this[http.request](url, { method, searchParams, signal })
return normalizeListResponse(response, this.constructor)
```

frontend/src/components/ShootDns/GManageDns.vue
grolu: renamed to `availableCredentialsForPrimaryDnsProvider`

frontend/src/components/ShootDns/GManageDns.vue
grolu: DONE

frontend/src/components/ShootDns/GManageDns.vue
grolu: Yeah this was vibe coded. Left it in while refactoring but also agree this is not the best way to format the code. I will change it

frontend/src/composables/credential/helper.js
grolu: Looks way nicer

frontend/src/composables/useShootDns.js
grolu: hmm from my experience this will not be anytime in the near future. For example the sync mechanism is still in place after 2 years and they did not implement any kind of migration logic. So old shoots still configure dns the old way. if we want to add the TODO here we should also add it at the other places like the `syncProvidersFromShootSpecDNS` etc.

frontend/src/components/ShootDns/GManageDns.vue
grolu: the problem here is that provider values may be empty. Another way to handle this would be to generate a temporary UID. I think we do this on the worker configuration dialog. However in this case the data is not a local copy but centralized in the store. Lets discuss how we can solve this

frontend/src/composables/useShootDns.js
grolu: No I think this is a leftover from previous refactoring optimization

frontend/src/composables/credential/helper.js
grolu: ok lets discuss but we should add it everywhere

frontend/src/components/ShootDns/GManageDns.vue
grolu: also identified other places using this formatting style

frontend/src/composables/useShootDns.js
grolu: done

frontend/src/components/ShootDns/GManageDns.vue
grolu: Store row key in WeakMap in composable

frontend/src/composables/credential/helper.js
grolu: added to all relevant code paths

frontend/src/composables/useShootDns.js
grolu: refactored to make the normalization in useShootContext normalization function and useShootDns computed setters

frontend/src/components/ShootDns/GDnsProviderRow.vue
```diff
@@ -239,15 +243,28 @@ export default {
         set(this.dnsProvider, ['zones', 'include'], value)
       },
     },
-    usedSecretNames () {
-      return this.dnsServiceExtensionProviders.map(({ secretName }) => this.getResourceRefName(secretName))
+    dnsExtensionProviderResourceRefs () {
+      return this.dnsServiceExtensionProviders.map(provider => this.getResourceRef(dnsExtensionProviderResourceName(provider)))
```
petersutter: Too nested? Extract parts to variable

frontend/src/components/ShootDns/GDnsProviderRow.vue
```diff
@@ -170,39 +170,43 @@ export default {
...
           name: resourceName,
           resourceRef: {
-            apiVersion: 'v1',
-            kind: 'Secret',
-            name: secretName,
+            apiVersion: credential.apiVersion,
+            kind: credentialKind,
+            name: credentialName,
           },
         })
```
petersutter, line 211: nit
```suggestion
      set (credential) {
        const currentResourceName = dnsExtensionProviderResourceName(this.dnsProvider)

        this.deleteResource(currentResourceName)

        if (!credential) {
          this.dnsProvider.credentials = undefined
          this.dnsProvider.secretName = undefined
          return
        }

        const credentialKind = credential.kind
        const credentialName = credential.metadata?.name
        const newResourceName = this.getDnsServiceExtensionResourceName({
          kind: credentialKind,
          name: credentialName,
        })

        this.dnsProvider.credentials = newResourceName
        delete this.dnsProvider.secretName

        this.setResource({
          name: newResourceName,
          resourceRef: {
            apiVersion: credential.apiVersion,
            kind: credentialKind,
            name: credentialName,
          },
        })
      }
```

frontend/src/components/ShootDns/GDnsProviderRow.vue
```diff
@@ -239,15 +243,28 @@ export default {
...
   },
   methods: {
     getErrorMessages,
+    credentialFilter (credential) {
+      const credentialName = credential?.metadata?.name
+      const credentialKind = credential?.kind
+      if (!credentialName || !credentialKind) {
+        return false
+      }
+      const currentResourceRef = this.getResourceRef(dnsExtensionProviderResourceName(this.dnsProvider))
```
petersutter: extract to variable
```suggestion
      const resourceName = dnsExtensionProviderResourceName(this.dnsProvider)
      const currentResourceRef = this.getResourceRef(resourceName)
```

frontend/src/composables/useShootContext.js
```diff
@@ -149,7 +163,7 @@ export function createShootContextComposable (options = {}) {
 
   function setShootManifest (value) {
     initialManifest.value = value
-    manifest.value = cloneDeep(initialManifest.value)
+    manifest.value = cloneDeep(normalizeManifest(initialManifest.value))
```
petersutter: `cloneDeep(normalizeManifest(initialManifest.value))`
why don't we use normalizedInitialManifest here?

frontend/src/composables/useShootDns.js
```diff
@@ -70,10 +78,24 @@ export const useShootDns = (manifest, options) => {
...
+  function getDnsServiceExtensionProviderUid (provider) {
+    if (!provider || typeof provider !== 'object') {
+      return undefined
+    }
+
+    let uid = dnsServiceExtensionProviderUidMap.get(provider)
+    if (!uid) {
+      uid = uuidv4()
+      dnsServiceExtensionProviderUidMap.set(provider, uid)
+    }
```
petersutter: This is probably fine here since the worst case is an unnecessary re-render, but we should avoid copying this pattern because object-identity-based keys can cause avoidable remounts and performance issues elsewhere. A better alternative is to use a deterministic business key or attach a persistent local UI id to each provider when it is created (non-enumerable).

frontend/src/composables/useShootContext.js
grolu: I think this is wrong. why do we normalize here... we don't do this in other `setManifest` functions. I will try to revert this change

petersutter (commented):
/lgtm
/approve

PR #2880, opened 2026-04-15
fix(frontend) DNS spec leftovers when configuring custom cluster domain

grolu:

Fixes #

Introduced with WorkloadIdentity Support for DNS (mostly) 

**Release note**:

```bugfix user
NONE
```

## Bug Fixes
* Refined custom domain activation logic to trigger based on partial DNS configuration
* Added automatic cleanup of DNS fields when custom domain is disabled
* Improved automatic initialization of DNS provider fields when custom domain is enabled
* Enhanced user experience with automatic focus on domain input field

### DIFF
--- a/frontend/src/components/ShootDns/GManageDns.vue
+++ b/frontend/src/components/ShootDns/GManageDns.vue
@@ -198,7 +198,7 @@ export default {
     const gardenerExtensionStore = useGardenerExtensionStore()
     const cloudProfileStore = useCloudProfileStore()
 
-    const customDomain = ref(!!dnsDomain.value && !!dnsPrimaryProviderType.value)
+    const customDomain = ref(!!dnsDomain.value || !!dnsPrimaryProviderType.value)
     const availableCredentialsForPrimaryDnsProvider = useCloudProviderEntityList(dnsPrimaryProviderType, { credentialStore, gardenerExtensionStore, cloudProfileStore })
 
     return {
@@ -315,8 +315,10 @@ export default {
         this.v$.dnsDomain.$reset()
       }
     },
-    dnsPrimaryProviderType () {
-      this.primaryDnsProviderCredential = head(this.availableCredentialsForPrimaryDnsProvider)
+    dnsPrimaryProviderType (value) {
+      if (value) {
+        this.primaryDnsProviderCredential = head(this.availableCredentialsForPrimaryDnsProvider)
+      }
     },
   },
   methods: {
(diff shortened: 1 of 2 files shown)


frontend/src/components/ShootDns/GManageDns.vue
```diff
@@ -198,7 +198,7 @@ export default {
     const gardenerExtensionStore = useGardenerExtensionStore()
     const cloudProfileStore = useCloudProfileStore()
 
-    const customDomain = ref(!!dnsDomain.value && !!dnsPrimaryProviderType.value)
+    const customDomain = ref(!!dnsDomain.value || !!dnsPrimaryProviderType.value)
```
petersutter: we should drop the customDomainEnabled getter/setter with reset logic and instead do something like
```js
const customDomainEnabled = ref(
  !!dnsDomain.value || !!dnsPrimaryProviderType.value || !!dnsPrimaryProviderCredentialsRef.value
)

// ....

watch(customDomainEnabled, enabled => {
  if (!enabled) {
    dnsDomain.value = undefined
    resetDnsPrimaryProvider()
    v$.value.dnsDomain.$reset()
    return
  }

  if (!dnsPrimaryProviderType.value) {
    dnsPrimaryProviderType.value = head(dnsProviderTypesWithPrimarySupport.value)
  }

  if (!dnsPrimaryProviderCredentialsRef.value) {
    // choose first matching credential
  }

  nextTick(() => dnsDomainRef.value?.focus())
})
```

frontend/src/composables/useShootContext.js
```diff
@@ -107,7 +107,12 @@ export function createShootContextComposable (options = {}) {
     }, value)
     const dnsProviders = get(object, ['spec', 'dns', 'providers'])
     if (dnsProviders) {
-      set(object, ['spec', 'dns', 'providers'], normalizeDnsPrimaryProviderCredentialsRefs(dnsProviders))
+      const normalizedProviders = filter(normalizeDnsPrimaryProviderCredentialsRefs(dnsProviders), 'type')
+      if (normalizedProviders.length > 0) {
+        set(object, ['spec', 'dns', 'providers'], normalizedProviders)
+      } else {
+        unset(object, ['spec', 'dns', 'providers'])
+      }
```
petersutter: if there is no type, the user should rather run into an error instead of trying to fix this, which might be unexpected for the user

grolu: okay, removed

frontend/src/components/ShootDns/GManageDns.vue
grolu: done

petersutter (commented):
/lgtm
/approve

PR #2883, opened 2026-04-15
Fix eager terminal config fetch on shoot details page

petersutter:
- `GCreateTerminalSessionDialog` fetched `terminalConfig` eagerly on the shoot details page, triggered by a `defaultTarget` watcher chain, even when the dialog was not open
- Guard `updateSettings()` in the `targetTab.selectedTarget` watcher so it only runs when the creation dialog is actually open (`newTerminalPrompt`) 
- Gate `GCreateTerminalSessionDialog` with `v-if="hasShootTerminalAccess"` in `GTerminalSplitpanes` so it is not mounted for users without terminal access

**Root Cause**:
When shoot data loads, `defaultTarget` changes from `undefined` to `'shoot'`, which sets `targetTab.selectedTarget`, triggering `updateSettings()` -> `terminalConfig` API call. For viewers this results in a 403 on `shoots/adminkubeconfig`. For all users it is a wasted call since the terminal creation dialog is not open.
  
Fixes #

**Release note**:

```bugfix user

```

### DIFF
--- a/frontend/src/components/GTerminalSplitpanes.vue
+++ b/frontend/src/components/GTerminalSplitpanes.vue
@@ -31,11 +31,15 @@ SPDX-License-Identifier: Apache-2.0
         />
       </template>
     </g-splitpane>
-    <g-create-terminal-session-dialog />
+    <g-create-terminal-session-dialog v-if="hasShootTerminalAccess" />
   </div>
 </template>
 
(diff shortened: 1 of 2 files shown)


grolu (approved):
/lgtm
/approve

PR #2908, opened 2026-04-20
feat(frontend) Use new `rotate-etcd-encryption-key` annotation to perform ETCD Encryption Key

grolu:

Fixes #

Required because of this change https://github.com/gardener/gardener/pull/12605

**Release note**:

```bugfix user
Use the new `rotate-etcd-encryption-key` annotation to rotate the ETCD encryption key. This operation is no longer performed in two steps. This updated approach is required for clusters running Kubernetes version 1.34.0 and later
```

## Release Notes

* **Bug Fixes**
  * Simplified etcd encryption key rotation from a two-step process to a single-step operation.
  * Improved operation detection for credential rotation workflows.

* **Improvements**
  * Updated credential rotation messaging for better clarity and consistency.

### DIFF
--- a/frontend/__tests__/components/GShootCredentialRotation.spec.js
+++ b/frontend/__tests__/components/GShootCredentialRotation.spec.js
@@ -94,12 +94,11 @@ describe('components', () => {
                 lastCompletionTime: '2022-07-05T10:01:42Z',
               },
               etcdEncryptionKey: {
-                phase: 'Completing',
                 lastInitiationTime: '2022-07-05T09:22:33Z',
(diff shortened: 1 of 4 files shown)


grolu:
@coderabbitai

>Guard completion-mode operation selection when no completion operation exists.

This should not be an issue as this code is only executed for two phase operations (that still exists btw) and not for one phase operations. Those operations will not have Phase 'Prepared' or 'Completing'

petersutter (commented):
/lgtm
/approve

PR #2943, opened 2026-04-30
ci: restrict workflow permissions to least privilege

petersutter:
/area security
/kind enhancement

Adds top-level `permissions: {}` to all workflow files so no implicit grants leak to jobs. Each job must explicitly declare what it needs (least-privilege principle). Adds missing per-job `permissions` blocks where previously inherited from workflow-level defaults.

Fixes #

Jobs using OIDC federation (`gardener/cc-utils/.github/actions/github-auth`) already declare `id-token: write` per job. The `prepare-hotfix` job also needs `contents: read` for `actions/checkout` (the checkout token comes from the app, but the action itself requires read access to clone).

**Release note**:
```other operator
NONE
```

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -11,6 +11,8 @@ on:
           the mode to use. either `snapshot` or `release`. Will affect effective version, as well
           as target-oci-registry.
 
+permissions: {}
+
 jobs:
   prepare:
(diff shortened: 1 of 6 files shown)


8R0WNI3 (approved):
/lgtm

PR #2962, opened 2026-05-12
fix(socket): guard against missing operator during store patch

grolu:
/area quality
/kind bug

Guards the socket event handler against a missing operator when the
store has not been initialized yet. Events arriving before store
initialization are deferred and requeued. If the operator cannot be
created for an already-initialized store (permanent misconfiguration),
an error is logged instead of silently swallowing events.

Fixes #

The default operator returns undefined when `state.list` is not an
array. Stores initialize `list` as `null` and set it to an array after
fetching. The guard distinguishes these two cases to avoid hiding real
failures behind a debug log.

**Release note**:
```bugfix user
Fixed socket event handler crashing when events arrive before store initialization.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Socket events are deferred and re-queued when the local store isn’t initialized, preventing lost/incorrect updates; queued events are flushed only when the store is ready or visibility changes.
  * Synchronization failures are logged and roll back without causing repeated retries; certain not-found statuses still remove stale update references.

* **New Features**
  * Store exposes an initialization state to gate event processing.

* **Tests**
  * Added tests for deferred handling, re-queueing, visibility gating, initialization behavior, and operator error handling.

### DIFF
--- a/frontend/src/composables/useSocketEventHandler.js
+++ b/frontend/src/composables/useSocketEventHandler.js
@@ -91,8 +91,13 @@ export function useSocketEventHandler (useStore, options = {}) {
 
         uidMap.set(uid, false)
       }
+      let patchResult = 'applied'
       store.$patch(state => {
         const operator = createOperator(state)
+        if (!operator) {
+          patchResult = state.list === null ? 'uninitialized' : 'invalid'
+          return
+        }
         // Delete items first
         for (const [uid, item] of uidMap) {
           if (typeof item === 'boolean' && !item) {
@@ -106,6 +111,17 @@ export function useSocketEventHandler (useStore, options = {}) {
           }
         }
       })
+      if (patchResult === 'uninitialized') {
+        logger.debug('Skipped synchronization of %s: store not yet initialized', pluralName)
+        for (const event of events) {
+          const { uid } = event
+          if (!eventMap.has(uid)) {
+            eventMap.set(uid, event)
+          }
+        }
+      } else if (patchResult === 'invalid') {
+        logger.error('Failed to synchronize %s: operator could not be created for current store state', pluralName)
+      }
     } catch (err) {
       if (isTooManyRequestsError(err)) {
         logger.info('Skipped synchronization of modified %s: %s', pluralName, err.message)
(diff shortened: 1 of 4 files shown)


frontend/src/composables/useSocketEventHandler.js
```diff
@@ -42,51 +57,100 @@ export function createListOperator (list) {
...
+      } else {
+        logger.error('Failed to synchronize modified %s: %s', pluralName, err.message)
+      }
+      // Synchronization failed -> Rollback events
+      restoreEvents(events)
+      return
+    }
+    if (!isInitialized(store)) {
+      restoreEvents(events)
+      return
```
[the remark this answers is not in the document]
petersutter: Isn't it the case that if the store resets during an active synchronize(), the connection was lost.. which triggers a disconnect/reconnect cycle that initiates a fresh synchronize with the new store state anyway? The stale patch would never survive because the entire sync restarts.

Wouldn't adding an epoch counter introduce complexity for a race condition that already self-heals through the existing reconnect mechanism?

frontend/src/composables/useSocketEventHandler.js
```diff
@@ -42,51 +57,100 @@ export function createListOperator (list) {
   }
 }
 
+function isStoreInitialized (store) {
+  return store.isInitial !== true
```
grolu, line 61: The useShootStore doesn't define isInitial, so isStoreInitialized returns true by coincidence (undefined !== true). This works today but is fragile — if someone later adds isInitial to the shoot store, the behavior silently changes.

frontend/src/composables/useSocketEventHandler.js
```diff
@@ -42,55 +57,103 @@ export function createListOperator (list) {
...
+    }
+    if (visibility.value !== 'visible') {
+      return
+    }
+    throttledHandleEvents()
+  }
+
   async function handleEvents (store) {
+    if (!isInitialized(store)) {
+      return
```
petersutter, line 102: log when handleEvents deferred?

petersutter (commented):
/lgtm
/approve

PR #3000, opened 2026-06-01
refactor(machineImages): replace distro name guessing with exact vendor lookup

grolu:
/area quality
/kind cleanup

Removes `getDistroFromImageName` which used substring matching to map
machine image names to vendor identifiers. This caused bugs where
images like `memoryone-gardenlinux` were silently mapped to the
`gardenlinux` vendor due to greedy `includes()` checks.

Machine image names are now looked up directly by exact name in the
vendor config. Adds missing vendor entries for `gardenlinux-fips` and
`memoryone-gardenlinux`. Unknown images fall back to displaying their
raw name instead of being silently misclassified.

Fixes #

The old `getDistroFromImageName` also matched `vsmp` — this only
appears as a machineType name, never as a machineImage name, so it
was dead code.

**Release note**:
```bugfix user
Fix machine image vendor matching to use exact names instead of substring matching, preventing misclassification of images like memorone-gardenlinux or gardenlinux-fips
```

```breaking operator
Machine image vendor matching no longer uses substring/wildcard patterns. Image names must now exactly match an entry in `knownMachineImageVendors` (or the branding config `vendors.machineImage` array). Images without an exact match will display their raw name without vendor icon or grouping
```

## Release Notes

* **New Features**
  * Added support for two new machine image vendors: Garden Linux (FIPS) and MemoryOne Garden Linux.

* **Refactor**
  * Simplified internal machine image vendor detection logic for improved maintainability.

### DIFF
--- a/frontend/src/composables/helper.js
+++ b/frontend/src/composables/helper.js
@@ -16,7 +16,6 @@ import get from 'lodash/get'
 import map from 'lodash/map'
 import isEqual from 'lodash/isEqual'
 import includes from 'lodash/includes'
-import lowerCase from 'lodash/lowerCase'
 
 export function cleanup (obj) {
   const cleanupObject = obj => {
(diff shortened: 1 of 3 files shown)


petersutter (commented):
/lgtm
/approve

PR #3014, opened 2026-06-11
feat(backend): support TLS termination in dashboard server

grolu:
/area security
/kind enhancement

Adds optional TLS termination support to the Gardener dashboard backend server. When `tls.certFile` and `tls.privateKeyFile` are configured, the server starts as HTTPS instead of plain HTTP.

- Validates that both `certFile` and `privateKeyFile` are provided together
- Creates an `https.Server` with the configured cert/key when TLS is enabled
- Falls back to plain `http.Server` when TLS is not configured
- Logs protocol (HTTP/HTTPS) on startup for observability

Fixes #

Certificate and key files are read synchronously at server startup via `fs.readFileSync`. The file paths come from validated config, not user input.

**Release note**:
```feature operator
The Gardener dashboard backend now supports optional TLS termination. Configure `tls.certFile` and `tls.privateKeyFile` in the dashboard configuration to enable HTTPS.
```

## Summary by CodeRabbit

* **New Features**
  * Optional HTTPS/TLS support: the server can start in HTTPS mode when both a certificate file and private key file are configured.

* **Bug Fixes**
  * TLS configuration is validated: if either the certificate or private key is missing, startup fails with a clear error instead of silently misconfiguring TLS.
  * Startup logging now accurately reports whether the service is running as HTTP or HTTPS.

* **Tests**
  * Added test coverage for TLS validation, HTTP fallback behavior, and HTTPS startup (including correct server options and protocol logging).

### DIFF
--- a/backend/lib/server.js
+++ b/backend/lib/server.js
@@ -5,6 +5,8 @@
 //
 
 import http from 'http'
+import https from 'https'
+import fs from 'fs'
 import { pTimeout } from './utils/p-timeout.js'
 import terminus from '@godaddy/terminus'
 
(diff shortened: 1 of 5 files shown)


backend/lib/server.js
```diff
@@ -19,13 +21,24 @@ function toMilliseconds (seconds) {
...
   const hooks = app.get('hooks')
+  const isHttps = Boolean(tls?.certFile && tls?.privateKeyFile)
 
   // create server
-  const server = http.createServer(app)
+  let server
+  if (isHttps) {
+    const tlsOptions = {
+      cert: fs.readFileSync(tls.certFile), // eslint-disable-line security/detect-non-literal-fs-filename -- path from validated config
+      key: fs.readFileSync(tls.privateKeyFile), // eslint-disable-line security/detect-non-literal-fs-filename -- path from validated config
```
petersutter: Suggestion: read the TLS cert/key files during config loading and set `config.tls = { cert, key }`. Then `server.js` can simply consume the prepared TLS options. This also leaves room for future validation of the certificate/private key contents.

grolu: done

petersutter (commented):
/lgtm
/approve

PR #3015, opened 2026-06-11
feat(dashboard): enhance landscape viewer UX with role badge and operator UI

petersutter:

/area usability
/kind enhancement

Broadens the visibility of operations-oriented UI to users with the `landscape-viewer` role. UI surfaces aimed at the operations team — control plane status chip, seed readiness tag, seed link on shoots, operator-only filters and columns on the shoot list, gardener version info — were previously gated by `isAdmin` and are now shown to any user satisfying the new `canViewLandscape` selector. `isAdmin` continues to gate mutating/admin-only capabilities such as terminal access.

The previously admin-only avatar role badge and chip in the main toolbar now also identify landscape viewers, with a dedicated title and icon. The internal flag `showAdminOnly` is renamed to `showLandscapeViewerOnly` to reflect the broader audience.

Fixes #

Prefer the new `canViewLandscape` authz selector over direct `isAdmin` checks for any UI that is informational/operational rather than mutating.

**Release note**:

```feature operator
Landscape viewers now see operations-oriented UI elements such as the control plane (CP) chip, seed readiness, seed link on shoots, and a role badge in the main toolbar
```
```breaking operator
The condition visibility configuration field `showAdminOnly` has been replaced by `showLandscapeViewerOnly`. Landscape operators that define additional conditions or override condition configuration must update their configuration accordingly. Conditions marked with `showLandscapeViewerOnly` are now visible to users with landscape-viewer access, instead of being restricted to dashboard admins only
```

## Summary by CodeRabbit

* **Refactor**
  * Updated authorization model to support landscape viewer roles in addition to admin roles.
  * Permission-gated UI elements now controlled by landscape viewer access instead of admin status.
  * User role display updated to show "Operator" or "Landscape Viewer" designation.
  * Seed and cluster monitoring visibility tied to landscape viewer permissions.

### DIFF
--- a/frontend/src/components/GStatusTag.vue
+++ b/frontend/src/components/GStatusTag.vue
@@ -77,7 +77,7 @@ SPDX-License-Identifier: Apache-2.0
 <script>
 import { mapState } from 'pinia'
 
-import { useAuthnStore } from '@/store/authn'
+import { useAuthzStore } from '@/store/authz'
 
 import GShootMessageDetails from '@/components/GShootMessageDetails.vue'
 
@@ -119,8 +119,8 @@ export default {
     },
   },
   computed: {
-    ...mapState(useAuthnStore, [
-      'isAdmin',
+    ...mapState(useAuthzStore, [
+      'canViewLandscape',
     ]),
     popoverKey () {
       return `g-status-tag[${this.condition.type}]:${this.shootMetadata.uid}`
@@ -173,7 +173,7 @@ export default {
       if (this.isUnknown) {
         return 'mdi-help-circle-outline'
       }
-      if (this.isProgressing && this.isAdmin) {
+      if (this.isProgressing && this.canViewLandscape) {
         return 'mdi-progress-alert'
       }
 
@@ -224,7 +224,7 @@ export default {
       if (this.isError) {
         return 'error'
       }
-      if (this.isProgressing && this.isAdmin) {
+      if (this.isProgressing && this.canViewLandscape) {
         return 'info'
       }
       return 'primary'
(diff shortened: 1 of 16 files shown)


frontend/src/views/GMembers.vue
```diff
@@ -351,13 +351,13 @@ const {
 } = storeToRefs(projectStore)
 const {
   namespace,
+  canViewLandscape,
   canManageMembers,
   canManageServiceAccountMembers,
   canCreateServiceAccounts,
 } = storeToRefs(authzStore)
```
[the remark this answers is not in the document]
petersutter: > After a landscape viewer removes their own project role

the landscape viewer is not a dedicated project role and thus can't be removed here -> not an issue

frontend/src/components/GStatusTag.vue
```diff
@@ -236,8 +236,8 @@ export default {
       return this.color
     },
     visible () {
-      if (!this.isAdmin) {
-        return !get(this.condition, ['showAdminOnly'], false)
```
grolu, line 240: `showAdminOnly` can be overridden by configuration / also additional conditions could make use of this flag. I suggest to keep support for `isAdmin` and make `showLandscapeViewerOnly` an additional visibility selector

petersutter: I tend to mark this as breaking change in a release note

grolu: Are we sure the isAdmin check never makes sense here? Probably yes. If the user can technically see it, the only reason to hide it would be to prevent confusion. Landscape viewers should have enough context to interpret the information correctly.
So I guess yes a breaking release note is ok in this case.

PR #3044, opened 2026-06-26
feat(filter): support field-qualified search terms

petersutter:
/area usability
/kind enhancement

Extends the shoot and seed list filters to support field-qualified search terms, e.g. `seed:aws-ha` or `-region:eu`.

Also fixes the search hint tooltip, which was sticking open while the field had focus — now shows on hover only (500 ms delay).

Fixes #

**Release note**:

```feature user
Support field-qualified search terms in shoot and seed list filters (e.g. `seed:aws-ha`, `-region:eu`).
```

## Summary by CodeRabbit

* **New Features**
  * Search now supports field-qualified queries (`field:value` via supported fields), quoted exact phrases, and negation.
  * Empty-value handling is more precise for field-specific and unqualified terms.

* **Bug Fixes**
  * Improved query parsing and matching for whitespace, tabs/newlines, and escaped quotes (including unterminated quotes).
  * Better results when certain fields are missing, null, non-strings, or boolean/numeric values.
  * Table search tooltip is now less intrusive (delayed and not opened on focus).

### DIFF
--- a/frontend/src/components/GTableSearch.vue
+++ b/frontend/src/components/GTableSearch.vue
@@ -5,7 +5,11 @@ SPDX-License-Identifier: Apache-2.0
  -->
 
 <template>
-  <v-tooltip location="top">
+  <v-tooltip
+    location="top"
+    :open-on-focus="false"
+    :open-delay="500"
+  >
     <template #activator="{ props: tooltipProps }">
       <v-text-field
         v-bind="tooltipProps"
--- a/frontend/src/composables/useTableFilter/helper.js
+++ b/frontend/src/composables/useTableFilter/helper.js
@@ -96,7 +96,8 @@ export class SearchQuery {
         if (value == null || value === '') {
           return false
         }
-        return term.exact ? value === term.value : value.includes(term.value)
+        const stringValue = typeof value === 'string' ? value : String(value)
+        return term.exact ? stringValue === term.value : stringValue.includes(term.value)
       }
       const found = values.some(matchesTermValue)
       if ((!found && !term.exclude) || (found && term.exclude)) {
--- a/frontend/src/composables/useTableFilter/helper.js
+++ b/frontend/src/composables/useTableFilter/helper.js
@@ -4,16 +4,80 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
-import find from 'lodash/find'
-import includes from 'lodash/includes'
+// Identifier-shaped `field:` prefix at the start of a token (after an optional
+// leading `-`). Used to split a token into qualifier and value.
+const fieldPrefixPattern = /^([A-Za-z_][A-Za-z0-9_]*):/
 
-const tokenizePattern = /(-?"([^"]|"")*"|\S+)/g
+const CHAR_QUOTE = 34 // "
+const CHAR_SPACE = 32
+const CHAR_TAB = 9
+const CHAR_LF = 10
+const CHAR_CR = 13
 
+function isTokenSeparator (code) {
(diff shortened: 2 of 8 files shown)


frontend/src/components/GTableSearch.vue
```diff
@@ -5,7 +5,11 @@ SPDX-License-Identifier: Apache-2.0
  -->
 
 <template>
-  <v-tooltip location="top">
+  <v-tooltip
+    location="top"
+    :open-on-focus="false"
+    :open-delay="500"
+  >
```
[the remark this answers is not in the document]
petersutter: won't-fix for now

frontend/src/store/shoot/helper.js
```diff
@@ -299,38 +316,29 @@ export function searchItemsFn (state, context) {
...
+        region: getRawVal(context, item, 'region'),
+        seed: getRawVal(context, item, 'seed'),
+        project: getRawVal(context, item, 'project'),
+        createdBy: getRawVal(context, item, 'createdBy'),
+        purpose: getRawVal(context, item, 'purpose'),
+        k8sVersion: getRawVal(context, item, 'k8sVersion'),
+        ticketLabels: getRawVal(context, item, 'ticketLabels'),
+        errorCodes: getRawVal(context, item, 'errorCodes'),
+        controlPlaneHighAvailability: getRawVal(context, item, 'controlPlaneHighAvailability'),
       }
```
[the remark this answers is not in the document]
petersutter: we rather need an "is" search directive, so we can do something like is:workerless. Not part of this PR though

frontend/src/composables/useTableFilter/helper.js
```diff
@@ -4,26 +4,101 @@
...
-      const found = !!find(values, value => term.exact ? value === term.value : includes(value, term.value))
+      const values = getTermValues(fields, allValues, term.field)
+      const matchesTermValue = value => {
+        if (term.value === '') {
+          return value == null || value === ''
+        }
+        if (value == null || value === '') {
+          return false
+        }
+        return term.exact ? value === term.value : value.includes(term.value)
```
grolu: if `value` is not a string, e.g.g a number, the search crashes. Searchable custom fields can return numbers/booleans unchanged

frontend/src/store/shoot/helper.js
```diff
@@ -299,38 +316,29 @@ export function searchItemsFn (state, context) {
...
-        return true
+  return search => {
+    const searchQuery = parseSearch(search, QUALIFIED_SEARCH_FIELDS)
+
+    return item => {
+      const fields = {
+        name: getRawVal(context, item, 'name'),
+        provider: getRawVal(context, item, 'provider'),
+        region: getRawVal(context, item, 'region'),
+        seed: getRawVal(context, item, 'seed'),
```
grolu, line 326:
```suggestion
        seed: getRawVal(context, item, 'seed'),
        infrastructure: getRawVal(context, item, 'infrastructure'),
```

grolu (changes requested):

frontend/src/store/shoot/helper.js
petersutter: I have dropped `infrastructure` from the `QUALIFIED_SEARCH_FIELDS` instead. Dedicated `provider` or `region` fields should be used instead

frontend/src/composables/useTableFilter/helper.js
petersutter: fixed with https://github.com/gardener/dashboard/pull/3044/commits/a54dfe1fb173e024f24385b3307ecfc6df21499b

PR #3059, opened 2026-07-03
feat(backend): enable weak ETags for non-API routes

grolu:
/area performance
/kind enhancement

Enables weak ETag generation on the main Express app so that static and non-API responses can benefit from conditional request caching (304 Not Modified). API routes are isolated into a dedicated sub-app with ETags explicitly disabled to preserve existing behavior for API consumers.

Fixes #

A new `apiApp` Express instance is created solely to carry the `etag: false` setting and the API router. The main app switches from `etag: false` to `etag: 'weak'`, enabling conditional responses for everything outside `/api`.

**Release note**:
```bugfix user
The Gardener dashboard now sends ETag headers for SPA fallback responses such as /login. This improves browser cache validation and fixes an issue where rolling back to a previous dashboard version could make browsers reuse stale HTML that referenced assets no longer present on the server, causing 404 errors and failed page loads
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated HTTP caching behavior so SPA fallback responses include an `etag`, while API, auth, and webhook responses omit `etag` to avoid unintended caching.

* **Tests**
  * Added end-to-end acceptance coverage verifying `etag` header behavior across SPA, API, auth, and webhook flows.

### DIFF
--- a/backend/__tests__/acceptance/caching.spec.js
+++ b/backend/__tests__/acceptance/caching.spec.js
@@ -1,0 +1,126 @@
+//
+// SPDX-FileCopyrightText: 2026 SAP SE or an SAP affiliate company and Gardener contributors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import {
+  describe,
+  it,
+  expect,
+  beforeAll,
+  afterAll,
+  beforeEach,
+} from 'vitest'
+import {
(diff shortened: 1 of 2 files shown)


petersutter (commented):
/lgtm
/approve

PR #3078, opened 2026-07-14
refactor(dashboard): redesign and unify detail tooltips

petersutter:
/area usability
/kind cleanup

Introduces two shared tooltip components (`GDetailTooltip`, `GStatusTagTooltip`) to eliminate duplicated tooltip markup and establish a consistent visual design across all affected tooltips.

Fixes #

**Release note**:
```other user
Redesigned tooltips: project details, infrastructure details, shoot health, and status tags now use a unified layout.
```
```other operator
Redesigned tooltips: shoot health donut, seed capacity indicators, and seed condition status now use a unified layout.
```

## Summary by CodeRabbit

* **New Features**
  * Added a reusable detail-tooltip component with consistent title/body/footer support and configurable sizing/behavior.
  * Migrated project, vendor, seed capacity, shoot health, and status tag tooltips to the new component with improved layouts and contextual messaging.
  * Enhanced tooltip/activator accessibility via improved activator handling, ARIA labels, and keyboard focus support.
  * Updated DNS credential provider column to display the “DNS” tooltip title.

* **Tests**
  * Added comprehensive unit tests covering slot rendering, conditional sections, activator wiring/prop forwarding, and tooltip validation/accessibility behaviors.

### DIFF
--- a/frontend/src/components/GShootHealthDonut.vue
+++ b/frontend/src/components/GShootHealthDonut.vue
@@ -251,8 +251,8 @@ const ariaLabel = computed(() => {
   }
   const parts = [
     'Shoot health distribution',
-    `${shootCount.value} shoots`,
-    `${matchingUnhealthy.value} unhealthy shoots`,
+    formatShootCount(shootCount.value),
+    formatShootCount(matchingUnhealthy.value, 'unhealthy'),
   ]
   if (activeFilterLabels.value.length > 0) {
     const desc = filterDescription.value
--- a/frontend/src/components/GShootHealthDonut.vue
+++ b/frontend/src/components/GShootHealthDonut.vue
@@ -6,7 +6,7 @@ SPDX-License-Identifier: Apache-2.0
 
 <template>
   <div
-    class="d-inline-flex align-center justify-center"
+    class="activator d-inline-flex align-center justify-center"
     :style="{ minWidth: `${donut.size}px`, minHeight: `${donut.size}px` }"
     tabindex="0"
     role="img"
@@ -14,7 +14,7 @@ SPDX-License-Identifier: Apache-2.0
   >
     <span
       v-if="shootCount === 0"
-      class="empty text-medium-emphasis"
+      class="text-medium-emphasis"
       aria-hidden="true"
     >-</span>
     <svg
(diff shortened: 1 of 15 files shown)


frontend/src/components/Seeds/GSeedCapacityIndicator.vue
```diff
@@ -114,6 +112,17 @@ const capacityText = computed(() => {
   return `${props.shootCount} / ${props.allocatableShoots}`
 })
 
+const capacityUsageLabel = computed(() => {
```
grolu, line 119: This label uses the already capped progress percentage, so an overcommitted seed such as 12 assigned out of 10 allocatable reports only “100% used”. Could we cap only the progress bar and calculate the visible and ARIA percentages from the raw ratio?

frontend/src/components/GStatusTagTooltip.vue
```diff
@@ -0,0 +1,89 @@
...
+
+<template>
+  <g-detail-tooltip
+    :activator="activator"
+    :disabled="disabled"
+    :open-delay="750"
+    :title="title"
+    :width="320"
+  >
+    <template v-if="userErrors.length">
```
grolu: The previous tooltip explicitly showed the current status, but the redesigned tooltip only renders errors or a description. If no description is configured, I think this could lead to an empty tooltip?

frontend/src/components/GVendor.vue
```diff
@@ -6,58 +6,60 @@ SPDX-License-Identifier: Apache-2.0
...
+    />
+    <span
+      v-if="description"
+      class="ml-2"
+    >
+      {{ description }}
+    </span>
+    <g-detail-tooltip
+      activator="parent"
+      title="Infrastructure"
```
grolu: `GVendor` is also used for DNS credential providers, where the hard-coded “Infrastructure” tooltip and accessible name are misleading. Pass the proper name or vendor type from the parent component

frontend/src/components/GStatusTag.vue
```diff
@@ -142,6 +122,9 @@ export default {
     chipText () {
       return this.condition.shortName || ''
     },
+    chipAriaLabel () {
```
grolu, line 125: The new accessible name ignores `staleShoot`, so a stale value is announced as current, e.g. “API Server: Healthy”, while the popover calls it “Last Status”. Could we include a “Last status” qualifier in both components when the value is stale?

frontend/src/components/GShootHealthDonut.vue
grolu: use the new singular-aware helper
Also, the test passes as 

`300 expect(label).toContain('1 unhealthy shoot')` passes also for `1 unhealthy shoots`

grolu (changes requested):

frontend/src/components/GVendor.vue
petersutter: for the dns credential providers, "DNS" is now shown as tooltip title

frontend/src/components/GStatusTagTooltip.vue
petersutter: the status omission was intentional. The tooltip still exposes the unabbreviated condition name, however the body container is now omitted when there is no description or user error.

frontend/src/components/GShootHealthDonut.vue
petersutter: ✅

frontend/src/components/GStatusTag.vue
petersutter: ✅

PR #3090, opened 2026-07-20
enh(frontend): use light background for infra icons

julestree:

/area usability
/kind enhancement

Adds a light background color to the infra icons so that they are easily visible in dark mode. This includes both the infra tiles on the tables and the infra tiles on the create cluster screen.

Fixes #

**Release note**:

```bugfix user
NONE
```
#### Before:
Screenshot 2026-07-20 at 16 10 40

Screenshot 2026-07-24 at 17 07 30

#### After:
Screenshot 2026-07-24 at 17 06 13

Screenshot 2026-07-24 at 17 06 45

## Summary by CodeRabbit

* **Style**
  * Improved dark-mode readability of vendor icons by using a fixed light background color.
  * Updated vendor icon rendering by removing the prior tile/rounded styling behavior.
  * Adjusted the infrastructure card to display the vendor icon with default background behavior (no longer overriding it).

### DIFF
--- a/frontend/src/components/GVendorIcon.vue
+++ b/frontend/src/components/GVendorIcon.vue
@@ -109,10 +109,7 @@ const iconStyle = computed(() => {
 </script>
 
 <style lang="scss" scoped>
-  @use 'vuetify/settings' as vuetify;
-  @use 'sass:map';
-
   .v-theme--dark .icon-background {
-    background-color: map.get(vuetify.$grey, 'darken-2');
+    background-color: #e0e0e0;
   }
 </style>
--- a/frontend/src/components/GVendorIcon.vue
+++ b/frontend/src/components/GVendorIcon.vue
@@ -7,7 +7,6 @@ SPDX-License-Identifier: Apache-2.0
 <template>
   <v-avatar
     :size="size"
-    rounded="lg"
     tile
     :class="{ 'icon-background': !noBackground }"
   >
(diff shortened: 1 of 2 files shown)


grolu:
Let's discuss if we can improve icons to be able to have transparent background also in dark mode.

petersutter:
> Let's discuss if we can improve icons to be able to have transparent background also in dark mode.

Could we keep the background transparent and add a subtle white(?) halo from the top in dark mode, e.g. via `filter: drop-shadow(...)`?

julestree:
@petersutter As we discussed, i've changed the icons to have a rounded border (check out the updated screenshots in the description). With this it looks a bit less harsh.

frontend/src/components/GVendorIcon.vue
```diff
@@ -7,7 +7,6 @@ SPDX-License-Identifier: Apache-2.0
 <template>
   <v-avatar
     :size="size"
-    rounded="lg"
     tile
```
grolu:
```suggestion
      rounded="lg"
```

frontend/src/components/GVendorIcon.vue
```diff
@@ -109,10 +107,8 @@ const iconStyle = computed(() => {
 </script>
 
 <style lang="scss" scoped>
-  @use 'vuetify/settings' as vuetify;
-  @use 'sass:map';
-
   .v-theme--dark .icon-background {
-    background-color: map.get(vuetify.$grey, 'darken-2');
+    background-color: #e0e0e0;
+    border-radius: 7px !important;
```
grolu:
```suggestion
```

grolu (changes requested):
See my suggestions. I think it is sufficient to remove `tile` which makes the avatar rectangular by removing any `border-radius` from the avatar component.

PR #3092, opened 2026-07-21
fix: improve color contrast of the UI

julestree:

/area usability
/area compliance
/kind enhancement
This is a first attempt to make the UI of the dashboard compliant with WCAG contrast guidelines. It changes the colors for both light and dark modes, so that they have at least 4.5:1 contrast ratio (AA rating), without changing the color theme too much.

#### Notable changes
- Removed custom styling of the menu highlighting - in dark mode it didn't have enough contrast. The default from vuetify seems to be good enough here.
- Added the culori lib for calculating contrast where vuetify does not provide adequate contrast, and replaced the get-contrast lib usages with culori
- Changed the default light and dark color schemes to provide more contrast out-of-the-box
- Explicitly set the login teaser to use dark theme, so the text colors will be picked correctly
- Introduce derived color tokens (`tonalPrimary`, `flatPrimary`, `on-flatPrimary`, etc.) which are calculated to ensure tonal and flat chips/alerts always meet contrast requirements
- Updated all tonal and flat chip and alert color props across the codebase to use derived tokens instead of raw semantic names
- Operators can override individual derived tokens (e.g. `tonalWarning`) via `frontend.themes` config. This overrides the calculated colors

#### Customizing the tonal colors
The tonal chip colors are adjusted automatically for contrast. To override a specific tonal color without changing the base color, set the tonal token directly by adjusting the Helm values file. For example for the tonalWarning color

```yaml
global:
    dashboard:
      frontendConfig:
        themes:
          light:
            tonal-warning: "#BF360C"
          dark:
            tonal-warning: "#FFAB40"
```

### Restoring the previous colors
To restore the previous colors, set the following in your Helm values:

```yaml
global:
  dashboard:
    frontendConfig:
      themes:
        light:
          primary: "#0b8062"
          tonal-primary: "#0b8062"
          anchor: "#0b8062"
          toolbar-background: "#0b8062"
          error: "#ff5252"
          tonal-error: "#ff5252"
          info: "#2196f3"
          tonal-info: "#2196f3"
          success: "#4caf50"
          tonal-success: "#4caf50"
          warning: "#fb8c00"
          tonal-warning: "#fb8c00"
        dark:
          primary: "#0b8062"
          tonal-primary: "#0b8062"
          anchor: "#0b8062"
          toolbar-background: "#0b8062"
          error: "#b71c1c"
          tonal-error: "#b71c1c"
          info: "#2196f3"
          tonal-info: "#2196f3"
          success: "#4caf50"
          tonal-success: "#4caf50"
          warning: "#e65100"
          tonal-warning: "#e65100"
```

Fixes #

**Release note**:

```other user
Update the default light and dark color schemes to meet accessibility standards for WCAG 2.1 AA. See [#3092](https://github.com/gardener/dashboard/pull/3092) for details on customizing or restoring previous tonal colors.
```

## Summary by CodeRabbit

* **Style**
  * Refined light and dark theme colors across navigation, buttons, chips, alerts, status indicators, and dialogs.
  * Improved contrast and readability for text, tonal colors, and custom color schemes.
  * Updated logout styling and dark-theme teaser presentation.
  * Standardized semantic styling for informational, warning, error, and primary messages.
  * Navigation now uses consistent default active-item highlighting.
* **Bug Fixes**
  * Improved color handling for labels and status indicators to maintain accessible contrast across themes.
* **Tests**
  * Expanded coverage for custom colors and accessibility behavior.


grolu:
If we want to go this direction we need to adapt the documentation to include the new theme colors. Also, we need to provide a breaking change release note.
But let's first discuss if this is what we want.

petersutter:
@coderabbitai review

grolu:
I had another idea that might be cleaner—WDYT?
We could define tonalError, tonalWarning, tonalSuccess, etc. globally when applying custom theme colors in useCustomColors. The resolution order would be: user override → calculated accessible color → original color fallback, so these colors are always available through the Vuetify theme.
A small chip wrapper could then use tonalError instead of error whenever variant="tonal", while keeping the original color for other variants. This would avoid chip-specific CSS overrides and make the tonal colors reusable for alerts, buttons, and similar components.

petersutter:
now with the latest changes incorporated, will start a new coderabbit review

@coderabbitai review

frontend/src/composables/useCustomColors.js
```diff
@@ -11,35 +11,118 @@ import {
...
+  const surfaceColor = get(themeColors, ['surface'])
+
+  for (const colorName of SEMANTIC_COLOR_NAMES) {
+    const tonalColorName = getTonalColorName(colorName)
+    const baseColor = get(themeColors, [colorName])
+    const customTonalColor = get(customDerivedColors, [tonalColorName])
+
+    if (customTonalColor) {
+      set(themeColors, [tonalColorName], customTonalColor)
+      continue
```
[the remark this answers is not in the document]
petersutter: I don't agree with that. This feature is there to restore previous visuals, even though the contrast might be suboptimal. This is in responsibility of the admin that configures / overrides the theme

frontend/src/utils/themeColors.js
```diff
@@ -0,0 +1,55 @@
...
+  'error',
+  'info',
+  'success',
+  'warning',
+  'unknown',
+])
+
+const semanticColorNames = new Set(SEMANTIC_COLOR_NAMES)
+
+function capitalize (value) {
```
petersutter: Do we want to use kebab-case consistently for these derived theme colors? The current approach produces mixed names such as `flatPrimary` and `on-flatPrimary`

`flat-primary`, `tonal-primary`, and `on-flat-primary` would match Vuetify’s multiword theme-color convention and our existing tokens.

frontend/src/components/GLoginTeaser.vue
```diff
@@ -12,7 +12,7 @@ SPDX-License-Identifier: Apache-2.0
   <!-- eslint-disable vue/no-v-html -->
   <div
     v-else
-    class="d-flex flex-column align-center justify-center bg-main-background-darken-1 pa-3"
+    class="v-theme--dark d-flex flex-column align-center justify-center bg-main-background-darken-1 pa-3"
```
petersutter: why was this change needed? I don't see a change on the login screen for both themes 🤔

frontend/src/composables/useCustomColors.js
```diff
@@ -11,35 +11,118 @@ import {
...
+}
+
+function patchThemes (themes, customThemes) {
+  for (const colorMode of ['light', 'dark']) {
+    const themeColors = get(themes, [colorMode, 'colors'], {})
+    const customThemeColors = get(customThemes, [colorMode], {})
+    const customDerivedColors = patchThemeColors(themeColors, customThemeColors)
+    setTonalThemeColors(themeColors, customDerivedColors)
+    setFlatThemeColors(themeColors, customDerivedColors)
   }
```
petersutter: I know you did not touch `useCustomColors`, but I think we should improve it with the following code while we’re here (we can then also drop the `nextTick` import):

```js
export const useCustomColors = async (customThemes, theme = useTheme()) => {
  let value = toValue(customThemes)

  // A nullish value means configuration is still loading; `{}` means loaded without overrides.
  if (value == null) {
    value = await new Promise((resolve, reject) => {
      const timeoutId = setTimeout(() => {
        unwatch()
        reject(new Error('Setting custom colors timed out'))
      }, 3000)
      const unwatch = watch(customThemes, value => {
        if (value == null) {
          return
        }
        clearTimeout(timeoutId)
        unwatch()
        resolve(value)
      })
    })
  }

  const themes = toValue(theme.themes) ?? {}
  patchThemes(themes, value)
}
```

The current version relies on an immediate watcher and `nextTick` just to safely unwatch. The suggested version is simpler: use the value if it’s already there, otherwise wait for it, then apply the colors.

petersutter (commented):
/lgtm
/approve

PR #3098, opened 2026-07-22
fix(frontend): harden shoot creation defaults

grolu:
Follow-up to #2476 containing the fixes intentionally separated from the shoot-default configuration work.

## What changed

- Preserve concrete values selected through the catch-all wildcard.
- Safely handle explicitly empty maintenance-hour configurations and invalid maintenance windows.
- Correct workerless initialization, worker restoration, field visibility, and validation behavior while preserving explicit state for new Shoots.
- Deduplicate the sorted networking-type choices.
- Correct the OpenStack load balancer provider validation field name.
- Touch Firewall Size validation when the Firewall Size field loses focus instead of touching Firewall Image.
- Add focused regression coverage for these repairs.

## Why

These repairs were intentionally separated from the configuration and default-application logic in #2476 so regressions can be isolated cleanly.

## Validation

- `make verify` passed.
- Full frontend suite: 73 files, 707 tests passed.

## Summary by CodeRabbit

- **New Features**
  - Workerless mode now hides infrastructure-specific options and makes their required validation conditional.
  - Loaded shoot manifests correctly infer workerless mode, and re-enabling workers restores default worker state when needed.
- **Bug Fixes**
  - Custom wildcard behavior now preserves the correct prefix/suffix handling.
  - Maintenance-hour inputs are validated; invalid settings fall back to safe defaults, including random maintenance begin generation.
  - Duplicate networking types are removed from selection lists.
  - Maintenance scheduling avoids invalid empty time ranges; firewall validation correctly updates after blur.
- **Tests**
  - Added/updated unit tests covering workerless mode, wildcard behavior, maintenance-hour validation, and component validation rules.

### DIFF
--- a/frontend/__tests__/components/GNewShootInfrastructureDetails.spec.js
+++ b/frontend/__tests__/components/GNewShootInfrastructureDetails.spec.js
@@ -1,0 +1,215 @@
+//
+// SPDX-FileCopyrightText: 2026 SAP SE or an SAP affiliate company and Gardener contributors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import {
+  nextTick,
+  ref,
+} from 'vue'
+import { shallowMount } from '@vue/test-utils'
+
+import GNewShootInfrastructureDetails from '@/components/NewShoot/GNewShootInfrastructureDetails.vue'
+
+function refList () {
+  return ref([])
+}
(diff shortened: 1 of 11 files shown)


petersutter (commented):
/lgtm
/approve

PR #3110, opened 2026-07-28
enh: port improvements from gardener demo terminal to dashboard

julestree:

/area usability
/kind enhancement

This PR ports improvements that were previously made in the terminal now-deprecated gardener-demo back to the dashboard. This includes:

- copy and paste shortcuts now work on windows
- adds the xterm WebGL renderer for faster/sharper drawing
- adds the Unicode 11 addon so wide/emoji characters are displayed correctly
- coloring/style adjustments to ensure contrast in both light and dark modes
- ~~removes the behavior that dimmed the terminal when it went out of focus~~ we will keep this behavior
- ~~stops sending the keep-alive heartbeat after the terminal has been hidden for 5 mins. When the terminal comes back into view it gets sent again~~
- revamp the burger button menu for splitting the panes - now it uses a list with items instead of cards and buttons, this makes it look more like the other menus
- split terminals: if a terminal window was drag-and-dropped to an invalid location, the drop position is now ignored so the terminal window does not get destroyed. Previously such a terminal would have been removed.

Fixes #

**Release note**:

```feature user
Add theming support and support for copy/paste shortcuts in terminal
```

## Summary by CodeRabbit

* **New Features**
  * Added Unicode 11 support to the terminal, with optional WebGL acceleration when available.
  * Terminal background now dims when the terminal is not focused.

* **Bug Fixes**
  * Terminal heartbeat now pauses when the page is hidden and resumes when visible.
  * Improved reliability of terminal attachment, including resizing and ping behavior.

* **Style**
  * Updated splitters and terminal system bar to use theme-aware visuals, including dimmed states.
  * Refreshed terminal menus and simplified the terminal page layout.


frontend/src/components/GTerminal.vue
```diff
@@ -131,82 +131,74 @@ SPDX-License-Identifier: Apache-2.0
...
                 v-bind="props"
                 icon="mdi-menu"
                 size="small"
-                color="grey-lighten-1"
+                color="on-surface"
                 variant="text"
                 class="text-none g-system-bar-button mx-1 g-ignore-drag"
               />
             </template>
-            <v-card rounded="0">
```
grolu, line 146: double check the menu. The settings button seems a bit misaligned (check vertical spacing)

petersutter: the cog color also does not match the split pane icon colors
maybe we could also add some padding to the keyboard shortcut

Image

old menu as comparison

Image

frontend/src/lib/terminal.js
```diff
@@ -166,6 +166,32 @@ export class TerminalSession {
     this.vm.term.loadAddon(attachAddon)
     let reconnectTimeoutId
     let heartbeatIntervalId
+    let inactiveTimeoutId
+
+    const startHeartbeat = () => {
```
grolu: should we handle the current visibility state here as well? If the user switches tabs before the WebSocket opens—or it reconnects in the background—the page may already be hidden when this listener is registered.

frontend/src/components/GTerminal.vue
julestree: Okay, i've implemented the changes, here are some screenshots:

Image
Image

petersutter: much better, very nice. thank you

frontend/src/lib/terminal-themes.js
```diff
@@ -0,0 +1,86 @@
...
+const dimmedBackground = {
+  dark: '#333333',
+  light: '#e0e0e0',
+}
+
+export function getTerminalTheme (isDark, hasFocus = true) {
+  const theme = { ...(isDark ? dark : light) }
+  if (!hasFocus) {
+    theme.background = isDark ? dimmedBackground.dark : dimmedBackground.light
+  }
```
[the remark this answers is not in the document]
julestree: i updated the description

petersutter (commented):
/lgtm

PR #3117, opened 2026-07-29
Proposal: Provide tonal colors in vuetify theme

grolu:

/area TODO
/kind TODO

Fixes #

**Release note**:

```bugfix user

```


frontend/src/components/GSeedStatusTag.vue
```diff
@@ -203,6 +205,11 @@ export default {
       }
       return 'primary'
     },
+    chipColor () {
+      return this.isError
+        ? this.color
```
julestree, line 210: I think we still need to fix the error chip color - the error color is used not only for flat chips with white text, but also for some icons on the dark background

frontend/src/components/GStatusTag.vue
```diff
@@ -214,6 +215,11 @@ export default {
       }
       return 'primary'
     },
+    chipColor () {
+      return this.isError
+        ? this.color
```
julestree, line 220: same here

julestree:
closing as the changes were added to #3092

PR #3122, opened 2026-07-30
[operations-view 1/7] refactor(frontend): support typed table search fields

petersutter:
/area quality
/kind enhancement

Makes table-search field definitions lazy and extensible. Previously every field was eagerly evaluated; now each field exposes a `match` function called only when a search term is active. This removes unnecessary computation and opens the door for domain-specific matching strategies (e.g. typed Shoot-search terms) without touching call sites.

Existing Seed and Shoot search behaviour is unchanged.

Fixes #

This is a prerequisite refactor — no user-visible behaviour change is intended. The API surface of `useTableFilter/helper` changes (match functions replace pre-computed strings), but all consumers are updated in this commit.

**Release note**:
```other developer
NONE
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved table and seed search matching for numeric, boolean-like, and other non-string values.
  * Corrected exact and partial matching, including null and empty values.
  * Ensured custom matching, qualified fields, and free-text exclusions work consistently.
  * Improved search efficiency by evaluating values only when needed.
* **Tests**
  * Expanded coverage for custom matching, qualified searches, empty values, and stringified value matching.

### DIFF
--- a/frontend/src/composables/useTableFilter/helper.js
+++ b/frontend/src/composables/useTableFilter/helper.js
@@ -18,14 +18,14 @@ function isTokenSeparator (code) {
   return code === CHAR_SPACE || code === CHAR_TAB || code === CHAR_LF || code === CHAR_CR
 }
 
-function getTermValues (fields, allValues, field) {
-  if (field === null) {
-    return allValues
+export function defaultFieldMatch (term, value) {
+  if (term.value === '') {
+    return value == null || value === ''
   }
-  if (!Object.hasOwn(fields, field)) {
-    return [undefined]
+  if (value == null || value === '') {
+    return false
   }
-  return [fields[field]] // eslint-disable-line security/detect-object-injection -- key validated by hasOwn above; fields is an app-constructed object literal, field is constrained to allowedFields
+  return term.exact ? value === term.value : value.includes(term.value)
 }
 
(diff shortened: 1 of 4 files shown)


frontend/src/composables/useTableFilter/helper.js
```diff
@@ -86,20 +87,17 @@ export class SearchQuery {
...
-          return value == null || value === ''
-        }
-        if (value == null || value === '') {
-          return false
-        }
-        const stringValue = typeof value === 'string' ? value : String(value)
-        return term.exact ? stringValue === term.value : stringValue.includes(term.value)
+      let found
+      if (term.field !== null) {
+        const spec = fields[term.field]
```
grolu: If allowed field names may eventually come from dynamic or user-controlled configuration, consider retaining an own-property check here. Otherwise, names such as constructor can resolve inherited properties and cause spec.get() to throw. The current hard-coded allowlists make this unreachable today, so this is mainly future-proofing

petersutter: https://github.com/gardener/dashboard/pull/3122/commits/6da06bc68600d604f7fadb66765ace30ec23461b

grolu (approved):

PR #3126, opened 2026-07-30
Trim leading whitespace from project filter input

jguipi:
/area usability
/kind bug

Project filter search ignored leading whitespace, causing no results when users accidentally typed a space before the project name. Now trims leading whitespace before normalizing the filter string.

Fixes #

One-line change in `GMainProjectSelection.vue` — `trimStart()` applied before `toLower()` on the filter value.

**Release note**:

```bugfix user
Project name filter now ignores leading and trailing whitespace, preventing empty search results caused by accidental spaces.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Project searches now ignore leading spaces, providing more accurate filtering results.

### DIFF
--- a/frontend/src/components/GMainProjectSelection.vue
+++ b/frontend/src/components/GMainProjectSelection.vue
@@ -288,7 +288,7 @@ const sortedAndFilteredProjectItems = computed(() => {
     }
   })
 
-  const normalizedFilter = toLower(projectFilter.value)
+  const normalizedFilter = toLower(projectFilter.value.trimStart())
 
   const predicate = item => {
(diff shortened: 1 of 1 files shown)


petersutter:
Hi @jguipi, thanks for the PR! Why not use `trim()` instead, so trailing whitespace is removed as well?

petersutter:
> Hi @jguipi, thanks for the PR! Why not use `trim()` instead, so trailing whitespace is removed as well?

we can do a follow up PR if we want to improve..

/lgtm
/approve

PR #3129, opened 2026-07-31
[merged to feature branch only][operations-view 3/7] refactor(frontend): sync list search with route query

petersutter:
/area usability
/kind enhancement

Adds router-scoped shallow synchronization for list searches. Search edits remain bookmarkable and survive browser history without running a full Vue Router navigation on every keystroke. This PR adopts the primitive only where it does not affect server subscription coverage.

Base branch is `enh/shoot-search-terms`.

**Release note**:
```other developer
NONE
```

### DIFF
--- a/frontend/src/composables/useRouteSearchQuery.js
+++ b/frontend/src/composables/useRouteSearchQuery.js
@@ -1,0 +1,140 @@
+//
+// SPDX-FileCopyrightText: 2026 SAP SE or an SAP affiliate company and Gardener contributors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import {
+  computed,
+  effectScope,
+  shallowRef,
+  watch,
+} from 'vue'
+import { useRouter } from 'vue-router'
+
+// Shares shallow search state between all consumers of the same router.
+const searchStateByRouter = new WeakMap()
+
+function normalizeSearch (value) {
+  return (Array.isArray(value) ? value[0] : value) ?? ''
+}
+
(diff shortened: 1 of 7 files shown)


frontend/src/composables/useRouteSearchQuery.js
```diff
@@ -0,0 +1,140 @@
...
+const searchStateByRouter = new WeakMap()
+
+function normalizeSearch (value) {
+  return (Array.isArray(value) ? value[0] : value) ?? ''
+}
+
+/**
+ * Creates the shallow search state shared by all consumers of a router.
+ */
+function createSearchState (router) {
```
grolu, line 25: AI suggested something like this

```
function createSearchState (router) {
  const routerHistory = router.options.history
  let popNavigationPending = false

  routerHistory.listen(() => {
    popNavigationPending = true
  })

  const clearPendingPop = () => {
    popNavigationPending = false
  }

  router.afterEach(clearPendingPop)
  router.onError(clearPendingPop)

  function replace (value) {
    if (popNavigationPending) {
      return
    }

    // Existing replacement logic...
  }
}
```

to prevent corrupt search state when editing search during navigation. However, this is more a theoretical issue as the time window is narrow and the UI blocks editing by the loading overlay anyway. So I leave it up to you if you want to add the guards...

petersutter: agree, the overlay should anyhow prevent this. won't fix

PR #3140, opened 2026-08-05
[operations-view 6/7] feat(frontend): add operations view controls

petersutter:

/area ops-productivity
/kind enhancement

Makes the Operations View visible and controllable from the Shoot list toolbar. Users can identify the current operations view state and intentionally restore it or show all clusters without changing saved configuration.

**Release note**:

```feature user
The all projects Shoot list toolbar now shows the active Operations View state and provides controls to restore it or show all clusters.
```

## Summary by CodeRabbit

- **New Features**
  - Added a dedicated Kubernetes Clusters toolbar with search, focus mode, Operations View controls, cluster creation, and column selection.
  - Added responsive layouts for desktop, tablet, and mobile screens.
  - Search fields can expand to full width when needed.
  - Column-selection controls support customizable button styling.

- **Improvements**
  - Simplified cluster-list filtering and updated active filter labels.
  - Improved preservation and handling of search, route, project-scope, and Operations View settings.
  - Focus mode now exits automatically when switching to a different cluster-list subscription.

### DIFF
--- a/frontend/src/components/GShootListToolbar.vue
+++ b/frontend/src/components/GShootListToolbar.vue
@@ -1,0 +1,595 @@
+<!--
+SPDX-FileCopyrightText: 2026 SAP SE or an SAP affiliate company and Gardener contributors
+
+SPDX-License-Identifier: Apache-2.0
+-->
+
+<template>
+  <v-sheet
+    class="toolbar text-toolbar-title"
+    color="toolbar-background"
+  >
+    <div class="heading">
+      <v-icon
+        class="heading-icon"
+        icon="mdi-hexagon-multiple"
+      />
+      <div class="heading-title text-title-large">
+        Kubernetes Clusters
+      </div>
+    </div>
+    <div class="controls">
+      <div class="search-context">
+        <div
+          class="search"
+          :class="{ 'search--expanded': modelValue }"
+        >
+          <g-table-search
+            :model-value="modelValue"
+            :examples="['my-shoot', 'John Doe']"
+            :exclude-examples="['myproject', 'Jane Doe']"
+            fluid
+            @update:model-value="$emit('update:modelValue', $event)"
+          />
+        </div>
+        <v-menu
+          v-if="!projectScope"
+          v-model="operationsViewMenuOpen"
+          :activator-props="{ 'aria-haspopup': 'dialog' }"
+          :content-props="{
+            role: 'dialog',
+            'aria-labelledby': 'operations-view-menu-title',
+            'aria-describedby': 'operations-view-menu-description operations-view-menu-status',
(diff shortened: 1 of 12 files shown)


frontend/src/components/GShootListToolbar.vue
```diff
@@ -0,0 +1,595 @@
...
+      activatorIcon: 'mdi-filter-off-outline',
+      statusLabel: 'All clusters',
+      statusIcon: 'mdi-filter-off-outline',
+      tooltip: 'Operations View — showing all clusters',
+      applyActionLabel: 'Apply Operations View',
+      applyActionDescription: 'Uses the Operations View criteria to show clusters that may need attention.',
+    }
+  }
+
+  const shootSearchTerms = sortBy(tokenizeSearch(props.modelValue))
```
grolu: Could we classify this using the same normalized query representation as the Shoot store instead of comparing raw token strings?
Supported queries such as -health:healthy -progressing:true -operatorAction:false -allTicketsIgnored:true normalize to the configured Operations View and use identical filtering/subscription behavior, but this code reports Custom search and offers to reapply the view. A whitespace-only query similarly has zero effective terms but is not classified as All clusters.
Please canonicalize both queries before comparing them and add cases for negated-equivalent and zero-term input.

frontend/src/components/GShootListToolbar.vue
```diff
@@ -0,0 +1,595 @@
...
+const focusModeInternal = computed({
+  get () {
+    return props.focusMode
+  },
+  set (value) {
+    emit('update:focusMode', value)
+  },
+})
+
+function showAllClusters () {
```
grolu, line 406: Could these actions be disabled while Focus Mode is active, or explicitly exit Focus Mode before changing the search?
When the health predicate changes, activateShootList replaces the subscription. The unsubscribe path calls clearAll(), which clears the frozen IDs and sets focusMode to false. Consequently, “Show all clusters” or “Apply Operations View” can silently discard the frozen snapshot even though Focus Mode says new rows are held until the user disables it. The previous filter controls were disabled while focused.

frontend/src/components/GShootListToolbar.vue
petersutter: https://github.com/gardener/dashboard/pull/3140/commits/ba97dfa4dc4dc61036d11fc258a6d02c634a5f91

frontend/src/components/GShootListToolbar.vue
petersutter: https://github.com/gardener/dashboard/pull/3140/commits/28eb3f043628f607fdff71bbbcdf2411be7ab198

PR #3101, opened 2026-07-23
Introduce generic credential input fields

grolu:
**How to categorize this PR?**

/area usability
/area quality
/kind cleanup

**What this PR does / why we need it**:

This is the fourth preparatory refactoring extracted from #2824. It follows #3075, #3082, and #3085.

- Introduces generic input-field components with configurable field types, validators, defaults, structured YAML/JSON parsing, Secret visibility, and file drop support.
- Uses a `yaml-secret` generic input field only in the already existing generic Secret dialog.
- Explicitly declares and forwards the generic dialog's `credential` prop.
- Generalizes the text-field drop helper with file-extension fallback, rejection feedback, and listener cleanup.

This gives the generic field abstraction a real, reviewable consumer without migrating provider-specific dialogs at the same time. Provider-specific dialogs, routing, vendor field configuration, and `useSecretContext` serialization remain unchanged.

**Which issue(s) this PR fixes**:

Part of #2824.

**Special notes for your reviewer**:

For valid YAML mappings, Secret handling is unchanged: the parsed mapping is assigned through the existing `secretStringData` encoder and produces the same top-level, base64-encoded Secret `.data`. Integration tests cover both create and edit flows, including dotted keys such as `serviceaccount.json`, and verify that existing values are populated when editing.

Empty, malformed, scalar, and array-root YAML is rejected before submission. Array-root YAML was accidentally accepted by the previous implementation and converted into numeric Secret keys.

The generic input fields component already contains logic for the follow-up PRs that will remove the provider specific dialogs. Eventually all / most providers will be handled by the GenericSecretDialog. Please review the logic even if not all of it is currently being used.

Validation: `make verify` (including 76 frontend test files and 735 frontend tests).

**Release note**:

```other developer
NONE
```

## Summary by CodeRabbit

* **New Features**
  * Added reusable input fields supporting text, passwords, selections, and structured JSON/YAML data.
  * Added secret visibility toggling, validation, and structured secret editing while preserving existing metadata.
  * Added JSON/YAML file drop support with extension detection and clear rejection messages.

* **Bug Fixes**
  * Improved handling of defaults, existing values, empty inputs, invalid structured data, and secret updates.
  * Improved cleanup of file-drop event handling.

* **Tests**
  * Added comprehensive coverage for generic inputs, secret dialogs, structured text handling, file drops, and input type classification.

PR #3075, opened 2026-07-14
Extract built-in vendor registry

grolu:
**What this PR does / why we need it**:

This is the first preparatory refactor extracted from #2824.

It moves the built-in infrastructure, DNS, and machine-image vendor registries out of the configuration store and into dedicated modules under `frontend/src/data/vendors`.

Each infrastructure and DNS provider now has its own module, providing the foundation for moving additional provider-specific configuration out of application code in subsequent PRs.

This PR intentionally does not change:

- vendor lookup or merge behavior
- vendor names, display names, icons, weights, or ordering
- shoot defaults
- credential dialogs or Secret serialization
- generated Gardener resources

The extracted registry objects are identical to the previous inline definitions.

**Related PR**:

- Original combined refactor: #2824

**Which issue(s) this PR fixes**:

None.

**Special notes for your reviewer**:

Please review this as the first preparatory refactor in the sequential split of #2824. Follow-up PRs will build on these modules one concern at a time.

Validation:

- `make verify`
- Registry values compared with the previous inline definitions

**Release note**:

```other developer
NONE
```

## Summary by CodeRabbit

* **New Features**
  * Added vendor listings for additional infrastructure providers, DNS services, and machine image distributions.
  * Added recognizable names, icons, and display ordering for each supported vendor.
  * Improved vendor selection consistency across infrastructure, DNS, and machine image options.

PR #3148, opened 2026-08-06
Select credentials by vendor type

grolu:

/area usability
/area quality
/kind cleanup

This is the fifth preparatory refactoring extracted from #2824. It follows #3075, #3082, #3085, and #3101.

Credential selection previously inferred the vendor domain from the provider name by checking the configured infrastructure and DNS provider lists. This was ambiguous when the same provider name existed in both domains: infrastructure was checked first, so a DNS selector could receive infrastructure bindings. It also coupled credential selection to provider-registry stores even though each caller already knows whether it is handling infrastructure or DNS credentials.

This change:

- requires credential selectors to provide an explicit vendor type (`infra` or `dns`)
- propagates that type through credential-list selection and the Secret dialog
- selects infrastructure bindings or DNS credentials directly from the caller's known context
- fixes access to the computed `isSecretBinding` value while selecting the binding context
- adds regression coverage proving that infrastructure and DNS credentials with the same provider name remain independent

For existing providers without overlapping names, behavior remains unchanged. Configurations that reuse a provider name across vendor domains now resolve the intended credential type deterministically.

Part of #2824.

This applies the same principle introduced by #3082 for `vendorDisplayName({ type, name })` and other vendor lookups: a provider name does not identify its vendor domain, so callers pass the domain they already know. Here, that explicit domain controls credential storage and dialog behavior rather than branding lookup.

Validation:

- `make verify` (including 76 frontend test files / 747 tests and 45 backend test files / 423 tests)
- focused `useShootDns` tests (16 tests)
- pre-commit checks

**Release note**:

```other developer
NONE
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved credential selection when infrastructure and DNS providers share the same name.
  * Infrastructure and DNS credential lists now consistently show credentials matching the selected credential type.
  * Updated credential dialogs and selectors to correctly preserve the selected binding context.

* **Tests**
  * Added coverage for provider names shared across infrastructure and DNS credential types.

### DIFF
--- a/frontend/__tests__/composables/useShootDns.spec.js
+++ b/frontend/__tests__/composables/useShootDns.spec.js
@@ -38,17 +38,40 @@ describe('composables', () => {
 
     it('should return dns credentials as a plain array for a dns provider type', () => {
       const credentialStore = useCredentialStore()
-      const gardenerExtensionStore = useGardenerExtensionStore()
-      const cloudProfileStore = useCloudProfileStore()
 
       const credentialsForProviderType = getCloudProviderEntityList('aws-route53', {
         credentialStore,
-        gardenerExtensionStore,
-        cloudProfileStore,
+        vendorType: 'dns',
       })
 
       expect(Array.isArray(credentialsForProviderType)).toBe(true)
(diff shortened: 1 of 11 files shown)


petersutter (commented):
/lgtm
/approve

PR #3154, opened 2026-08-07
Use generic dialog for AWS credentials

grolu:
## What changed

- Migrate AWS infrastructure and Route53 credential handling to `GSecretDialogGeneric`.
- Define the AWS access key, secret access key, and optional Route53 region as generic input fields.
- Add a named help slot to the generic dialog while preserving its configured HTML and default help fallbacks.
- Render the AWS-specific IAM guidance and policy code blocks through a dedicated, code-owned help component.
- Remove the legacy AWS credential dialog.

## Why

This is the next focused slice extracted from https://github.com/gardener/dashboard/pull/2824. The AWS credential fields fit the generic input-field model, but their help content contains application components such as external links and a JSON code block. Keeping this content in a dedicated Vue component preserves those capabilities without making component rendering configurable through vendor data or reintroducing generic `helpJSONTemplate` handling.

The Route53 region uses `omitWhenEmpty`, so clearing the optional field removes it without changing the global empty-value behavior of legacy dialogs.

## Validation

- `make verify`
- Focused generic-dialog and wrapper tests cover AWS/Route53 routing, field serialization, custom help-slot behavior, and optional-region removal.

## Summary by CodeRabbit

* **New Features**
  * Added AWS access key ID and secret access key fields with validation.
  * Added optional AWS Region configuration for Route 53 credentials.
  * Added provider-specific AWS help with IAM guidance and policies.
  * AWS and Route 53 credentials now use a shared dialog with contextual help.
  * Added support for customized help content in secret dialogs.

* **Bug Fixes**
  * Empty optional Route 53 regions are excluded from saved secret data.
  * Improved validation and encoding of AWS credential values.


frontend/src/components/Credentials/GSecretDialogHelpAws.vue
```diff
@@ -0,0 +1,144 @@
...
+    In this example, the placeholder for the hosted zone is Z2XXXXXXXXXXXX
+  </p>
+  <g-code-block
+    max-height="100%"
+    lang="json"
+    :content="policy"
+  />
+</template>
+
+<script>
```
petersutter: use script setup for new components

frontend/src/components/Credentials/GSecretDialogGeneric.vue
```diff
@@ -26,20 +26,22 @@ SPDX-License-Identifier: Apache-2.0
...
-      <div v-else>
-        <p>
-          This is a generic secret dialog.
-        </p>
-        <p>
-          Please enter data required for <code>{{ vendorName }}</code>.
-        </p>
-      </div>
+      <slot name="help">
+        <!-- eslint-disable vue/no-v-html -->
```
petersutter: we should add a reason why we disable the elsint rule here

frontend/src/data/vendors/dns/aws-route53.js
```diff
@@ -12,5 +14,15 @@ export default {
         },
       },
     ],
+    fields: [
+      ...aws.secret.fields,
```
petersutter: hm, not sure if I like that :) 
How about ex/importing field constants instead of "blindly" spreading the fields here? this would also make it more explicit

frontend/src/components/Credentials/GSecretDialogGeneric.vue
petersutter: `<!-- eslint-disable vue/no-v-html -- HTML comes exclusively from static vendor definitions and cannot be controlled by users. -->`

if it works as one line, I'd prefer that

grolu: done

petersutter (commented):
/approve
/lgtm

PR #3207, opened 2026-08-26
add eslint rule for tonal color tokens

julestree:
/area quality
/area dev-productivity
/kind enhancement

Adds an ESLint rule requiring semantic tonal color tokens such as `tonal-primary` when a vuetify component uses the tonal variant. Works with kebab-case and PascalCase components.

This is related to PR #3092 which adds the color tokens for tonal variants

NONE

`v-chip` uses the tonal variant by default. The rule therefore also checks `v-chip`s without an explicit variant.
Dynamic bindings, conditional expressions, helper calls, hexadecimal colors, and non-semantic colors are intentionally ignored.

**Release note**:
```other developer
NONE
```

## Summary by CodeRabbit

* **New Features**
  * Added linting guidance for tonal `v-alert` and `v-chip` colors.
  * Flags non-tonal semantic colors and suggests the appropriate tonal color token.
  * Supports explicit tonal variants and common component usage patterns while avoiding unrelated components and bindings.

* **Tests**
  * Added coverage for valid tonal usage, dynamic values, non-tonal components, and invalid color tokens.
  * Verifies suggested replacements and diagnostic messages for unsupported semantic colors.

### DIFF
--- a/frontend/eslint-rules/require-tonal-color-token.cjs
+++ b/frontend/eslint-rules/require-tonal-color-token.cjs
@@ -64,11 +64,14 @@ module.exports = {
           return
         }
 
+        const hasUnqualifiedBind = node.startTag.attributes.some(a =>
+          a.directive && a.key.name.name === 'bind' && !a.key.argument,
+        )
         const variantAttribute = findAttribute(node, 'variant')
         const variant = getStaticAttributeValue(node, 'variant')
         const isTonal =
           variant === 'tonal' ||
-          (component === 'v-chip' && !variantAttribute)
+          (component === 'v-chip' && !variantAttribute && !hasUnqualifiedBind)
 
         if (!isTonal) {
           return
--- a/frontend/eslint-rules/require-tonal-color-token.cjs
+++ b/frontend/eslint-rules/require-tonal-color-token.cjs
@@ -1,0 +1,96 @@
+//
+// SPDX-FileCopyrightText: 2026 SAP SE or an SAP affiliate company and Gardener contributors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+const tonalColors = new Map([
+  ['primary', 'tonal-primary'],
+  ['warning', 'tonal-warning'],
+  ['error', 'tonal-error'],
+  ['info', 'tonal-info'],
+  ['success', 'tonal-success'],
+])
+
+function findAttribute (node, name) {
+  return node.startTag.attributes.find(attribute => {
+    if (!attribute.directive) {
+      return attribute.key.name === name
+    }
+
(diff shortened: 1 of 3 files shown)


frontend/eslint-rules/require-tonal-color-token.cjs
```diff
@@ -0,0 +1,99 @@
+//
+// SPDX-FileCopyrightText: 2026 SAP SE or an SAP affiliate company and Gardener contributors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+const tonalColors = new Map([
```
grolu: Could we derive this mapping from the shared SEMANTIC_COLOR_NAMES definition instead of maintaining a second list? This currently omits secondary, accent, and unknown, although the theme creates corresponding tonal-* tokens for them. Sharing the definition would cover every semantic color and prevent the rule from drifting when tokens are added later.

frontend/eslint-rules/require-tonal-color-token.cjs
```diff
@@ -0,0 +1,99 @@
...
+    schema: [],
+    messages: {
+      useTonalColor: 'Use "{{expected}}" instead of "{{actual}}" for tonal {{component}}.',
+    },
+  },
+
+  create (context) {
+    return context.sourceCode.parserServices.defineTemplateBodyVisitor({
+      VElement (node) {
+        const component = node.rawName
```
grolu: Could we normalize the component name or explicitly support the PascalCase forms as well? node.rawName preserves the template spelling, so <VChip> and <VAlert> currently bypass the rule even though both are valid in Vue SFC templates.

frontend/eslint-rules/require-tonal-color-token.cjs
```diff
@@ -0,0 +1,99 @@
...
+
+        if (component !== 'v-alert' && component !== 'v-chip') {
+          return
+        }
+
+        const hasUnqualifiedBind = node.startTag.attributes.some(a =>
```
grolu: Could we make the handling of unqualified v-bind order-aware? In Vue 3, regular props are merged in declaration order and the last value wins. The current check handles the default chip variant, but <v-chip variant="tonal" color="primary" v-bind="chipProps" /> is still reported even though chipProps can override either prop. We would not need to evaluate the object itself—a conservative left-to-right scan could mark both values unknown at each object binding and let later explicit attributes make them known again.
I think this one is optional - we currently do not have such a case in our code. nevertheless this would make the rule more robust.

julestree: since we don't have a case like this in our code, I don't think we should add this

Issue #2612, opened 2025-09-02
Enhance Input Validation


grolu:
This issue tracks all open **frontend input tasks** that currently do not have a dedicated GitHub issue.  

## Task List

- [x] Fix ANSI conversion  
  ➡️ Normalize ANSI output to ensure correct rendering
  - first find out if the shoot statuses still contain ANSI output. If they don't, then we can remove the ANSI parsing altogether as it wouldn't be used anywhere -> PR https://github.com/gardener/dashboard/pull/3251

- ~~[ ] Switch to `ansi-html-community`~~ - **Won't do** - this is not needed. The both ansi-html and ansi-html-community have no dependencies and have not been updated since 2021

- [ ] Double-check direct usage of `href`, `src`, etc.  
  ➡️ Ensure these attributes only contain safe user input -> PR: https://github.com/gardener/dashboard/pull/3252

- [ ] Add sanitization for `mailto:` links  
  ➡️ Prevent injection of malicious query parameters (e.g., subject, body) -> related PR: https://github.com/gardener/dashboard/pull/3247

- [ ] Sanitize GitHub links  
  ➡️ Ensure all GitHub-provided links are properly validated before rendering
       See also #883

- [ ] Usage of Shoot Metadata
  ➡️ Implement safe getters in central composables to ensure input is validated (useObjectMetadata likely the best place) (gardenctl [ref](https://github.com/gardener/gardenctl-v2/blob/12024a86ea05f0cd3fdff365475d9c257e167cef/pkg/target/target.go#L115-L131))

- [x] Validate CSS color values from GitHub labels
  ➡️ Sanitize label.color from GitHub API responses to match hex color format (`/^[0-9a-fA-F]{6}$/`) before using in `:style` bindings in GTicketLabel.vue -> PR https://github.com/gardener/dashboard/pull/3250

PR #3247, opened 2026-09-08
fix: strip query params from mailto links

julestree:
/area security
/kind bug

`mailto:` links with query params (`?subject=`, `?body=`) can pre-fill email content when clicked, enabling phishing. Two fixes:
- strip query params from `mailto:` hrefs for any external URL passed through the sanitizer, as well as for markdown parsed in the ticket bodies. This means subject and body will not be parsed into the link.

related to #2612 

I couldn't reproduce that this was actually an issue for the standard mailto links generated by the dashboard. If you put something like `email@example.com?subject=foo&body=bar` as a user email, the query params are not rendered as a link on the dashboard. The only place the issue really seems to exist is in mailto links in the github issue comments (which do not get sanitized when viewing directly on github, by the way)

So I am unsure whether we should really be this defensive in this change.

**Release note**:
```bugfix user
Strip query params from mailto: links to prevent phishing via pre-filled subject/body injection.
```

## Bug Fixes

* Email links now retain only the email address, with query parameters removed during sanitization.
* Potentially unsafe content in email link subjects and bodies is no longer preserved.
* Email links are handled consistently across content processing.
* Email addresses remain intact while query parameters are removed despite capitalization, whitespace, or embedded control characters in the URL scheme.

### DIFF
--- a/backend/lib/markdown.js
+++ b/backend/lib/markdown.js
@@ -17,6 +17,15 @@ import sanitizeHtml from 'sanitize-html'
 
 const SANITIZE = {
   allowedTags: [...sanitizeHtml.defaults.allowedTags, 'img', 'details', 'summary'],
+  transformTags: {
+    a (tagName, attribs) {
+      if (attribs.href?.startsWith('mailto:')) {
+        const url = new URL(attribs.href)
+        attribs.href = `mailto:${url.pathname}`
+      }
+      return { tagName, attribs }
+    },
+  },
 }
 
 const processor = unified()
--- a/backend/lib/markdown.js
+++ b/backend/lib/markdown.js
@@ -19,7 +19,7 @@ const SANITIZE = {
   allowedTags: [...sanitizeHtml.defaults.allowedTags, 'img', 'details', 'summary'],
   transformTags: {
     a (tagName, attribs) {
-      if (attribs.href?.startsWith('mailto:')) {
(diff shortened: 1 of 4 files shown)


grolu:
The review showed that the actual check for the member email happens in `isEmail` function in `utils/index.js`.
It seems that the original issue is already handled here.
Nevertheless it maybe still makes sense to avoid query params in other mailto links. Moreover, we could discuss if we want to stop rendering emails as links.
Let's discuss with @petersutter 
/hold

petersutter:
> we could discuss if we want to stop rendering emails as links

yes, as discussed we could at least show the copy button next to the mail

backend/lib/markdown.js
```diff
@@ -17,6 +17,17 @@ import sanitizeHtml from 'sanitize-html'
 
 const SANITIZE = {
   allowedTags: [...sanitizeHtml.defaults.allowedTags, 'img', 'details', 'summary'],
+  transformTags: {
+    a (tagName, attribs) {
+      // eslint-disable-next-line no-control-regex -- intentional: strip control chars and whitespace before scheme check
+      const href = attribs.href?.replace(/[\u0000-\u0020\u007f]/g, '')
+      if (href?.toLowerCase().startsWith('mailto:')) {
+        const url = new URL(href)
```
chrisstiefel: How about adding also a positive email regex?

PR #3262, opened 2026-09-14
Hide infrastructure cards on new shoot page if only one provider is available

matthias-horne:

/area usability
/kind enhancement

This PR hides the infrastructure provider card on the new shoot page, if only a single provider is available. In this case the user cannot chose a different provider anyways. This is most useful for air-gapped environments like GDC, where there is also no ambiguity about the infrastructure where the shoot will be created.

This PR does not overlap with #2478 because it hides this card automatically and does not require explicit configuration by an operator.

Fixes #

Split from #3222 

**Release note**:

```other user
The infrastructure provider card will be hidden on the new shoot page if only a single provider is available.
```

## Summary by CodeRabbit

- **User Interface**
  - The new shoot form now shows infrastructure selection only when multiple provider types are available.
  - Form cards use consistent spacing for a cleaner, more uniform layout.
  - Available infrastructure providers remain presented in sorted order.

- **Bug Fixes**
  - When only one infrastructure provider is available, it is selected automatically instead of requiring manual selection.


grolu:
We discussed this internally and think this makes sense. If someone explicitly want to show the infrastructure selection even if there is only one infrastructure available, we can add a configuration like `alwaysShowInfrastructureSelection` to the config. But we will do this only if someone asks for it.
However, I have an improvement suggestion: If we hide the infra selection from the shoot creation page, we should also skip the infra / dns credential dropdown when clicking on the `+` button on the credentials page. In this case we should also directly open the dialogs. WDYT?

matthias-horne:
> We discussed this internally and think this makes sense. If someone explicitly want to show the infrastructure selection even if there is only one infrastructure available, we can add a configuration like `alwaysShowInfrastructureSelection` to the config. But we will do this only if someone asks for it. However, I have an improvement suggestion: If we hide the infra selection from the shoot creation page, we should also skip the infra / dns credential dropdown when clicking on the `+` button on the credentials page. In this case we should also directly open the dialogs. WDYT?

Thanks for the suggestion. Implemented in the latest commit.

frontend/src/views/GNewShoot.vue
```diff
@@ -292,4 +287,12 @@ export default {
   height: 48px;
   padding-right: 10px;
 }
+
+.card-container > .v-card {
+  margin-top: 16px;
+
+  &:first-child {
+    margin-top: 0px;
+  }
```
petersutter, line 296: We should probably use `gap` for the card spacing in the future. Nothing to address in this PR, though. cc @grolu

petersutter (approved):
/lgtm
/approve

PR #3265, opened 2026-09-14
Support secret key aliases

grolu:
Add support for documented alternative Secret keys in credential dialogs. The Dashboard now reads the primary key first and then configured field aliases, while field updates write the preferred key and preserve unexposed Secret data.

The OpenStack dialog now uses the common field access logic instead of secretStringDataRefs. This enables Designate OS_* aliases and prevents unexposed settings such as region or certificate data from being dropped. OpenStack authentication behavior remains unchanged.

Conflict detection is intentionally omitted. When a primary key and one or more aliases coexist, the Dashboard reads the primary value. Editing the field writes the primary key and removes its aliases. We consider a separate warning unnecessary because the entries describe the same exposed credential field; all other Secret data remains untouched.

**Release note**:

```feature user
Credential dialogs now support documented alternative Secret keys. Updating credentials writes the provider’s preferred key while preserving unexposed data, including additional OpenStack Designate settings.
```

## Summary by CodeRabbit

- **New Features**
  - Added support for common environment-variable aliases across DNS provider credentials.
  - Credential dialogs now load existing secrets more reliably and preserve unrelated secret data when fields are edited.
  - Improved OpenStack credential editing and hydration for existing secrets.
  - Standardized credential field handling across Azure, AWS, Netlify, Cloudflare, PowerDNS, and OpenStack providers.

- **Bug Fixes**
  - Unsupported legacy credential key formats are no longer mistakenly recognized.
  - Removed the HMAC-MD5 option from RFC2136 TSIG algorithm selection.
  - Corrected Azure and RFC2136 credential detail resolution.


frontend/src/components/Credentials/GSecretDialogOpenstack.vue
```diff
@@ -110,6 +111,22 @@ import GGenericInputField from '@/components/GGenericInputField'
 
 import { useProvideSecretContext } from '@/composables/credential/useSecretContext'
 
+import get from 'lodash/get'
+
+function secretFieldValue (key) {
```
petersutter: rename to useSecretFieldValue?

frontend/src/composables/credential/helper.js
```diff
@@ -328,13 +329,29 @@ export function isInfrastructureBinding ({ binding, infraProviderTypes }) {
   return infraProviderTypes.includes(bindingProviderType(binding))
 }
 
+// Resolve aliases before field decoding or detail extraction, without changing the Secret.
+export function resolveSecretDataAliases (data = {}, fields = []) {
```
petersutter: rename to "copySecretAliasesToKeys"?

frontend/src/composables/credential/helper.js
```diff
@@ -328,13 +329,29 @@ export function isInfrastructureBinding ({ binding, infraProviderTypes }) {
   return infraProviderTypes.includes(bindingProviderType(binding))
 }
 
+// Resolve aliases before field decoding or detail extraction, without changing the Secret.
```
petersutter: this comment does not belong here, it does not describe the fn itself

MartinWeindel:
lgtm, thanks! 👍

Good to see the fix for RFC2136 algorithm. `hmac-md5` was never supported.

Regarding the `apiToken` field for Netlify, I think it is better for consistency with other providers to introduce it.
Please keep it in your code.
See [github.com/gardener/external-dns-management/pull/1093](https://github.com/gardener/external-dns-management/pull/1093)

frontend/src/components/Credentials/GSecretDialogOpenstack.vue
grolu: Renamed it to createSecretFieldValue. I avoided the use prefix because it creates an Options API computed descriptor rather than a composable.

frontend/src/composables/credential/helper.js
grolu: Good suggestion, renamed accordingly. This better describes that aliases are copied to their primary keys for reading.

frontend/src/composables/credential/helper.js
grolu: Agreed, removed it. The updated function name describes the behavior sufficiently.

frontend/src/components/Credentials/GSecretDialogOpenstack.vue
```diff
@@ -176,6 +173,22 @@ export default {
...
   },
   computed: {
+    secretFieldValues: {
+      get () {
+        return this.getSecretFieldValues(this.providerFields)
+      },
+      set (value) {
+        this.setSecretFieldValues(this.providerFields, value)
+      },
+    },
```
petersutter: Can we drop this forwarding writable-computed indirection and call `setSecretFieldValues()` directly from the field setter? 

```diff
diff --git i/frontend/src/components/Credentials/GSecretDialogOpenstack.vue w/frontend/src/components/Credentials/GSecretDialogOpenstack.vue
index f0bb83095..7ca6d64d7 100644
--- i/frontend/src/components/Credentials/GSecretDialogOpenstack.vue
+++ w/frontend/src/components/Credentials/GSecretDialogOpenstack.vue
@@ -119,10 +119,10 @@ function createSecretFieldValue (key) {
       return get(this.secretFieldValues, [key], '')
     },
     set (value) {
-      this.secretFieldValues = {
+      this.setSecretFieldValues(this.providerFields, {
         ...this.secretFieldValues,
         [key]: value,
-      }
+      })
     },
   }
 }
@@ -173,13 +173,8 @@ export default {
     }
   },
   computed: {
-    secretFieldValues: {
-      get () {
-        return this.getSecretFieldValues(this.providerFields)
-      },
-      set (value) {
-        this.setSecretFieldValues(this.providerFields, value)
-      },
+    secretFieldValues () {
+      return this.getSecretFieldValues(this.providerFields)
     },
     authURL: createSecretFieldValue('authURL'),
     domainName: createSecretFieldValue('domainName'),
```

PR #3266, opened 2026-09-15
chore(eslint): migrate to eslint-plugin-import-x

julestree:
/area dev-productivity
/kind technical-debt

`eslint-plugin-import` uses APIs removed in ESLint 10 and has no ESLint 10-compatible release. This PR replaces it with `eslint-plugin-import-x` across all packages.

**Changes**:
- Migrate resolver config from import/resolver (v2 key map) to import-x/resolver-next (v3 object array)
- Remove eslint-import-resolver-local.cjs and replace with inline resolver objects in backend/charts configs and ignore patterns for virtual modules in frontend
- Remove stale .yarnrc.yml entries
- Suppress import-x/named in test files (vi.mock false positives)
- Fix pre-existing lint violations in files touched by rebase (trailing spaces, keyword spacing, operator linebreak)

```bugfix user
NONE
```


backend/lib/app.js
```diff
@@ -3,6 +3,7 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 //
+/* eslint-disable import-x/no-named-as-default-member -- helmet exposes sub-methods on its default export; this is the documented usage pattern */
```
petersutter: shouldn't we disable the warning only for the (next) line instead of disabling it for the whole file?

julestree: it's a good point, but there are 7 lines where this is a problem in this file. Do you prefer that we disable them one by one, in a block, or for the whole file?

petersutter: In this case, I think I'd prefer to disable the block

petersutter (commented):
/lgtm
/approve

